Issue #23, opened 2015-09-30
Instanced 3D Model and Composite Tile Formats


pjcozzi:
First, see the original notes for [Instanced 3D Model](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/Instanced3DModel/README.md) and [Composite](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/Composite/README.md) tile formats. To help design the Instanced 3D Model tile format, see the [Batched 3D Model](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/b3dm/README.md) tile format.

We now have significant work towards instancing in Cesium.  See https://github.com/AnalyticalGraphicsInc/cesium/pull/3049.

Here's some initial thoughts:

## Instanced 3D Model

```
Header
------
magic            // uchar[4], "i3dm"
version          // uint32,   1
batchTableLength // uint32,   can be zero, length of the batch table in bytes
glTFLength       // uint32,   length of glTF section (after the Batch Table) in bytes
glTFFormat       // uint32,   0 = url, 1 = embedded binary glTF
instancesLength  // uint32,   number of instances, NOT number of bytes, can be zero I guess

Batch Table
-----------
// Same as Batched 3D Model.  Optional.

glTF
----
// UTF-8 string or binary blob depending on Header.glTFFormat
// Length in bytes is Header.glTFLength

Instances
---------
There are Header.instancesLength of these:
* x, y, z    // doubles in WGS84 coordinates
* x, y, z, w // doubles in WGS84 coordinates (quaternion)
* batchId    // uint32, index into batchTable, zero when Header.batchTableLength === 0
```

* [x] ~~We can significantly optimize the size of each `instance` field, e.g., positions can be quantized relative to the bounding volume (see [EXT_quantized_attributes](https://github.com/mlimper/glTF/tree/spec-1.0/extensions/MultiVendor/EXT_quantized_attributes)).  Rotation could also be optional (part of the batchTable) and default to Y up (glTF default) in a local frame for the instance's position.  Rotation could use fewer fields.  RTC.   Delta encoding... SOA vs. AOS...  Look into point cloud compression...~~ #33
* [x] ~~Should `instance.batchId` be omitted when `header.batchTableLength === 0`.  This will complicate the client, but save memory; however, the case it saves memory for is probably rare in practice.~~ #33
[...]

PR #31, opened 2015-11-05
Composite tile format spec

pjcozzi:
@slchow can you please copyedit?  The bulk of the new writing is in `TileFormats/Composite/README.md`.

@lilleyse can you please do a technical review?  Just one question for you:
* [x] Inner tiles are stored tightly packed, right?  There's not need, for example, to have them on four-byte boundaries?

Part of #23.


TileFormats/Composite/README.md
```diff
@@ -1,17 +1,59 @@
...
+![](figures/layout.png)
+
+### Header
+
+The 16-byte header contains:
+
+* `magic` - 4-byte ANSI string `cmpt`.  This can be used to identify the arraybuffer as a Composite.
+* `version` - `uint32`, which contains the version of the Composite format. It is currently `1`.
+* `byteLength` - `uint32`, the length of the entire Composite tile, including the header and each inner tile, in bytes.
+* `tilesLength` - `uint32`, the number of tiles in the Composite.  Must be greater than or equal to zero.
```
lilleyse: Do we want to allow Composite tiles that have zero inner tiles?

lilleyse:
I don't see a problem with inner tiles being tightly packed.

TileFormats/Composite/README.md
pjcozzi: I thought about requiring this to be greater than zero.  It could be fine, but it is not consistent with other `Length` properties in 3D Tiles, e.g., `batchTableLength`.  Perhaps the consistency is less important than being as precise as possible for the context.

I'm OK with either.  Are you still thinking `> 0`?

pjcozzi:
> I don't see a problem with inner tiles being tightly packed.

Agreed.

TileFormats/Composite/README.md
lilleyse: Actually I think its okay as `>= 0`. Forcing it to be greater than zero seems too strict.

pjcozzi: OK, we'll keep it.  We can revisit before 1.0 if experience proves otherwise.

slchow:
@pjcozzi  Copyedited. Just minor changes to TileFormats/Composite/README.md; other files all look good.

pjcozzi:
Wow, that was fast, thanks @slchow.

pjcozzi:
@lilleyse please merge when you think this is ready.

pjcozzi:
@lilleyse this is ready.

lilleyse:
The composite tile page looks good to me. I have a suggestions that may not be relevant to this pull request:

* The **TileFormats** section only has Batched and Points right now.
* The Batched 3d Model spec should have a `batchSize` or `batchLength` property, and the header size should be `20` since we removed this from `tiles.json`.

pjcozzi:
Good points.  Updated.  This is ready.

lilleyse:
Looks good. Merged.

Issue #33, opened 2015-11-06
Instanced 3D Model updates


pjcozzi:
* [x] We can significantly optimize the size of each `instance` field, e.g., positions can be quantized relative to the bounding volume (see [EXT_quantized_attributes](https://github.com/mlimper/glTF/tree/spec-1.0/extensions/MultiVendor/EXT_quantized_attributes)).  Rotation could also be optional (part of the batchTable) and default to Y up (glTF default) in a local frame for the instance's position.  Rotation could use fewer fields.  RTC.   Delta encoding... SOA vs. AOS...  Look into point cloud compression...
* [x] Should `instance.batchId` be omitted when `header.batchTableLength === 0`.  This will complicate the client, but save memory; however, the case it saves memory for is probably rare in practice.
* [x] Semantics for batch table properties such as scale.

lilleyse:
Omitting `instance.batchId` seems fine to me and should be easy to handle on the client side.

pjcozzi:
Sounds good, updated the spec in 17743a072826ddfbc7a151c127c89152ab44ac36.

Please update https://github.com/AnalyticalGraphicsInc/cesium/pull/3158 if you haven't already.

lilleyse:
All updated there.

pjcozzi:
Thanks :+1:

lilleyse:
We have a lot of fields that say "This must be greater than or equal to zero" that are of type `uint32`. Is that redundant?

pjcozzi:
Good catch, #40.

arneschilling:
Hi all,

we are now stating to implement I3DM for our vegetation datasets.
After reading the spec I have a couple of questions and suggestions for improvements.

1. the instances field should be extended to include position, scale, rotation and the model's center point. This could be done by adding more numbers to each instance, e.g.
        Tx, Ty, Tz, Rx, Ry, Rz, Ra, Sx, Sy, Sz, Cx, Cy, Cz
        ( T: translation, R: rotation as rotation axis vector and angle in rad, S: scale, C: center point within the model used for positioning). See [http://www.web3d.org/documents/specifications/19775-1/V3.2/Part01/components/group.html#Transform] for X3D Transform spec.
      or simply by using a transformation matrix ,e,g
[...]

PR #34, opened 2015-11-06
Instanced 3D Model spec

### LINKED ISSUE #23: Instanced 3D Model and Composite Tile Formats

pjcozzi:
Fixes #23

@slchow please copyedit.  The new writing is in `TileFormats/Instanced3DModel/README.md`.  Note that most of this PR is the same changes as #31.

@lilleyse please do a technical review.  Does the Cesium client handle the corner cases for when, for example, `instancesLength` or `batchTableLength` is zero?

@lilleyse merge #31 first.

Before we declare 1.0, there's a few things to address in #33


lilleyse:
* Right now I'm using `uint16` for `batchId`, but we could go with `uint32` as well.
* Should `gltfLength` and `batchTableLength` be rename with `byteLength`?
* I tested the corner cases, and they work correctly.

TileFormats/Instanced3DModel/README.md
```diff
@@ -1,13 +1,113 @@
...
+
+* a binary blob containing binary glTF.
+
+In either case, `header.glTFLength` contains the length of the glTF field in bytes.
+
+## Instances
+
+The `instances` field immediately follows the `glTF` field (which may be omitted when `header.glTFLength` is `0`).
+
+The `instances` field contains `header.instancesLength` tightly packed instances.  Each instance has three fields:
```
slchow: Is there a word missing here? I couldn't make sense of "contains `header.instancesLength` tightly packed instances"

pjcozzi: A word is not missing.  That should read as "the instances field contains x tightly packed instances" where the value of "x" is defined by `header.instancesLength`.

If you think we could make it more clear, please have at it.

pjcozzi:
> Right now I'm using uint16 for batchId, but we could go with uint32 as well.

`uint16` is OK.  That allows 64K per tile.  That ought to be enough for anybody. :)

If experience proves otherwise, we can make this `unit16` or `uint32` depending on the value of `header.batchTableLength` or just `uint32` if compression make the size difference not matter.

> Should gltfLength and batchTableLength be rename with byteLength?

Yes, I will make the rename for b3dm and i3dm in this branch.

pjcozzi:
> If experience proves otherwise, we can make this unit16 or uint32 depending on the value of header.batchTableLength or just uint32 if compression make the size difference not matter.

Sorry, I mean the length of the arrays in the batch table, not the size of the JSON string in bytes.

pjcozzi:
> Should gltfLength and batchTableLength be rename with byteLength?

Also, please make these same renames throughout the Cesium implementation.

`glTFLength` -> `gltfByteLength`

`glTFFormat` -> `gltfFormat`

`batchTableLength` -> `batchTableByteLength`

pjcozzi:
@lilleyse this is updated.  I'll do the b3dm `batchTableByteLength` rename in the `composite` branch, which has the other b3dm changes, now.

@slchow are you still copyediting?

slchow:
@pjcozzi I'm done.

pjcozzi:
Thanks @slchow!

@lilleyse merge when ready.

lilleyse:
The layout image still has `batchId` as `uint32`, Other than that, it looks good to me.

pjcozzi:
Good catch, @lilleyse.  Updated.

lilleyse:
Merged.

PR #44, opened 2015-12-03
Tighten up spatial coherence wording

pjcozzi:
Fixes #43.  See the new "Bounding volume spatial coherence" section.

@lilleyse can you review and merge?  This is a small change; most of it is new TOC links.

Based on discussion with @kring - https://groups.google.com/forum/#!profile/cesium-dev/APn2wQdDf8J2kNnhMkqr9zuD_pygdVUnsj_Ry_fyJdzzNnQ1nVxZA0EUyytOfBuqM3VR6JSBsHGH/cesium-dev/tCCooBxpZFU/y3nKJT9eBgAJ

PR #45, opened 2015-12-03
External tileset

### LINKED ISSUE #42: Tighten up spec when content.url is another tileset
Lessons learned from https://github.com/AnalyticalGraphicsInc/cesium/pull/3237

A tile with `content.url` pointing to an external tileset (another tile.json):

* Cannot have any `children`.
* Is semantically the same as the external tileset's root tile so `root.geometricError === tile.geometricError` and `root.box === tile.content.box` (or equals `tile.box` if `tile.content.box` is `undefined`).
* Cannot be used to create cycles, for example, by pointing to the same tiles.json containing the tile or by pointing to another tiles.json that then points back to the tiles.json containing the tile.

@e-andersson any other corner cases we should add to the spec?

pjcozzi:
Fixes #42.

@e-andersson and @lilleyse please review.  See the new "External Tilesets" section.

@slchow can you please create a figure similar to https://github.com/AnalyticalGraphicsInc/cesium/pull/3237#issuecomment-161545525 in the same style as your other figures for the remaining TODO?  The legend in the top left and the caption at the bottom does not need to be part of the figure.

Merge #44 first.  This change is only one commit: 6f0637a5cd0384d1a301759cc5334046e37f2328


lilleyse:
The new section looks good to me.

pjcozzi:
Thanks @lilleyse.

@e-andersson any comments based on your implementation experience?

slchow:
@pjcozzi I can make the figure but it may not happen tomorrow. When do you need it?

e-andersson:
Looks good. It might be helpful to add a note somewhere on how to use empty tiles to create the kind of tree needed for the global tileset of tilesets described, but perhaps that'll come with the coming figure.

pjcozzi:
@slchow next week is OK, no rush.

@e-andersson I agree the figure will help.  @lilleyse and I also plan to write a series of tutorials showing how the flexibility of 3D Tiles can be used for diverse use cases like this, see https://github.com/AnalyticalGraphicsInc/cesium/issues/3241.  You are also welcome to write a tutorial if you would like.

pjcozzi:
@slchow added the figured.

PR #47, opened 2015-12-09
Figure

slchow:
Fixes #45  Is this OK? I moved the labels inside the boxes. Happy to make any changes you'd like.


pjcozzi:
Thanks @slchow.  I should have caught this before, but I don't think we need the dashed lined at the top.

Can we also make the spacing consistent on the bottom row, e.g., on the right, the two tiles.json boxes are right next to each other?

Also, can you include the original .eps if there is one?

slchow:
@pjcozzi Here's the revised figures. Let me know if there's more.

pjcozzi:
Perfect, thanks @slchow!

PR #52, opened 2016-01-15
Change refine to inherit from parent

### LINKED ISSUE #7: Should the refine property inherit from its parent tile?
Currently `refine` is optional and defaults to `"add"` (as opposed to `"replace"`).

Instead, if it is omitted, should it inherit its parent's value recursively up the tree?

pjcozzi:
Fixes #7.

Cesium code change to follow shortly.


pjcozzi:
Cesium code change: https://github.com/AnalyticalGraphicsInc/cesium/pull/3434

lilleyse:
Looks good to me.

PR #53, opened 2016-01-15
Rename tiles.json to tileset.json

pjcozzi:
Merge #52 first.

Fixes #49.

@TomPed can you make the corresponding code change in Cesium?  Make sure to rename the `tiles.json` files used in the tests.

PR #56, opened 2016-01-15
Add asset object for versioning

### LINKED ISSUE #50: tileset.json should specify tileset version or id
Since most 3D Tiles tilesets are generated once and then served as static files, it's good practice to provide an aggressive caching policy in order to avoid unnecessary server hits.  This is similar to how we serve Cesium's world terrain, for which we set a cache policy of one year.  The problem with this approach is that if I want to update an existing tileset, browsers that already have older versions of the tiles won't request new ones (for a year at least).

This is a common problem in the industry and is solved via query parameters.  We can add a property to the root tiles.json file whose value gets appended as a query string to every tile request.  tiles.json would then be served with a less aggressive caching policy (1 hour) and the browsers would automatically detect when it changes causing new copies of the previously cached tiles to be fetched as needed.

As an example, in STK Terrain Server's [layer.json](http://assets.agi.com/stk-terrain/world/layer.json), there is a version property that indicates the version of the tileset (not the version of the spec).  Requests are then made by Cesium with the `?version=<value>` query parameter as I described above.  We could use a `version` property in tileset.json as well, since it has semantics that allow developers to tell what version of a dataset a client is using, or we could simply add an `id` property that require it to be a GUID string so that it's always unique uniquely identify the tileset.

I'm not partial to either approach, as long as the spec enforces one of them. (or optionally determine a complete different way to solve the caching problem).

pjcozzi:
Fixes #50.

@mramato if this works for you, I will update Cesium to expose this in the 3D Tiles API.

Merge #52 and #53 first.  This is only f8b3c67170a7862f25c4794bb22f6fd468ca70de.


pjcozzi:
> This is only f8b3c67.

It is also 1234f7330a999d8ba3fcc0208363ae42c8b24394.  I forgot to save!

lilleyse:
Looks good, merged.

lilleyse:
Ah... I forgot to wait for @mramato feedback.

mramato:
No worries, I took a quick glance and `tilesetVersion` sounds exactly like what I wanted.

pjcozzi:
Cool, will update Cesium today or Monday.

PR #58, opened 2016-01-21
Tighten up spec for tile.content

pjcozzi:
Made `tile.content` required for leaf nodes to support this:
```
// When a tile with content is loaded, its parent can safely refine to it without any gaps in rendering
// Since an empty tile doesn't have content of its own, its descendants with content need to be loaded
// before the parent is able to refine to it.
```


lilleyse:
Looks good to me, merged.

Issue #26, opened 2015-10-03
Spec Writing Guide


pjcozzi:
As we add more tile formats, 3D Tiles needs to stay consistent.

## Terminology

* [x] Tiles are composed of `sections` such as `header` and `body`.  Sections are composed of `fields` such as `magic` and `version`.
* [x] "Feature" - indicates one model in a batch of models (b3dm), one instance in a collection of instances (i3dm), one point in a point cloud (pnts), one polygon in a vector tile (vctr), etc.

## Fields

* [x] field names are in camelCase.
* [x] `Length` - a `Length` suffix on a field name indicates the number of elements in an array.
* [x] `ByteLength` - a `ByteLength` suffix indicates the number of bytes, not to be confused with just `Length`.

## Header

* [x] Each tile format starts with a header that starts with the following fields:
```
magic            // uchar[4], indicates the tile format
version          // uint32,   1
byteLength       // uint32,   length, in bytes, of the entire tile.
```

## Binary

* [x] All binary data, e.g., tile formats, are in little endian.

_This is a work-in-progress.  There are more guidelines to come._

PR #59, opened 2016-01-26
Rename model/instance to feature

pjcozzi:
Part of #26.

This introduces the generic term _feature_ to describe individual models, instances, etc. in a tile.

This is just spec language and doesn't change the tileset.json schema or any tile formats.

Note that I didn't heavily change the b3dm and i3dm specs since, in context, I think it is better to use model/instance, and then use feature when speaking generally as in the core 3D Tiles spec.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ To support tight fitting volumes for a variety of datasets from regularly divide
 
 _TODO: include a screenshot of each bounding volume type._
 
-A tile usually references a _model_ or set of _models_, e.g., 3D buildings.  These models may be batched together into essentially a single model to reduce client-side load time and WebGL draw call overhead.
+A tile references a _feature_ or set of _features_, such as 3D models representing buildings or trees, points in a point cloud, or polygons, polylines, and points in a vector dataset.  These features may be batched together into essentially a single feature to reduce client-side load time and WebGL draw call overhead.
 
 ## Tile metadata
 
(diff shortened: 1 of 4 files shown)


pjcozzi:
Also, I'll make the rename in Cesium today.

README.md
```diff
@@ -70,7 +70,7 @@ To support tight fitting volumes for a variety of datasets from regularly divide
 
 _TODO: include a screenshot of each bounding volume type._
 
-A tile usually references a _model_ or set of _models_, e.g., 3D buildings.  These models may be batched together into essentially a single model to reduce client-side load time and WebGL draw call overhead.
+A tile references a _feature_ or set of _features_, such as 3D models representing buildings or trees, points in a point cloud, or polygons, polylines, and points in a vector dataset.  These features may be batched together into essentially a single feature to reduce client-side load time and WebGL draw call overhead.
```
lilleyse: Does it make sense to call a group of batched features a feature? I think this sentence made more sense when the word model was still being used.

pjcozzi: It makes as much sense as it does to call it a model.  Geospatial folks think of a polygon as a feature, if we combine two polygons into one polygon, that batched polygon would also be a feature.  It is not as native to us as graphics developers, but I think it is fine.

lilleyse: Ok that's good with me then.

lilleyse:
The rest is good, merged.

README.md
mramato: Just an FYI, in geospatial terms, multiple features (heterogeneous or not) are almost always referred to as a `feature collection`.

PR #62, opened 2016-01-26
Feature rename tweaks

pjcozzi:


lilleyse:
I made a small edit, but this is fine.

PR #65, opened 2016-01-28
add JSONObject Attribute to batchTable

jbo023:
We are working with datasets which often have inhomogeneous or even nested object attribute datasets. 

Because of this we would prefer if the specification doesn't set strict presets on the allowed datatype for batchTable attributes. With inhomogeneous data this leads to a Batchtable where many fields are "null". 

In the pull request I just added JSONObject to the allowed datatypes. This will allow us to add complex data as an batchTable attribute. We have done some testing with this and it actually already works with cesium, as long as the batchtable is a valid JSON.

### DIFF
--- a/TileFormats/Batched3DModel/README.md
+++ b/TileFormats/Batched3DModel/README.md
@@ -45,14 +45,15 @@ The batch table maps each `batchId` to per-model properties.  If present, the ba
 
 The batch table is a `UTF-8` string containing JSON.  It immediately follows the header.  It can be extracted from the arraybuffer using the `TextDecoder` JavaScript API and transformed to a JavaScript object with `JSON.parse`.
 
-Each property in the object is an array with its length equal to `header.batchLength`.  Each array is a homogeneous collection of `String`, `Number`, or `Boolean` elements.  Elements may be `null`.
+Each property in the object is an array with its length equal to `header.batchLength`.  Each array is a homogeneous collection of `String`, `Number`, `Boolean` or `JSONObject` elements.  Elements may be `null`.
 
 A vertex's `batchId` is used to access elements in each array and extract the corresponding properties.  For example, the following batch table has properties for a batch of two models.
 ```json
 {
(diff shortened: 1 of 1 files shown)


pjcozzi:
Thanks @jbo023!  I am actually just now starting to flush out the 3D Tiles properties access in the Cesium API so this is good timing.  Give me a few days to get to this.

pjcozzi:
@jbo023 I need to look more carefully at this Cesium implementation, but this is OK with me if it doesn't have any big implication.  From memory, I think it is fine.

For the declarative styling (#2) we would just implement the member operator (`.`) to allow access to sub-properties.

pjcozzi:
@jbo023 sorry for the slow response here, but I reviewed the spec and Cesium code and talked this over with @tfili who did some of the implementation work, and it should be fine to store `Object` or even `Array` here.

@lilleyse could you update the unit tests in Cesium with an example where properties are objects and arrays before we update the spec here?

mramato:
I'm sure there's a good answer to this, but what is the type information used for outside of parsing?  From a parsing implementation standpoint, arrays of strings, numbers, and booleans are also valid JSON so technically it's all the same type (JSON).  Is there a need for the client to know the type after parsing is done, thus requiring it to be specifically specified?  Is it for validation purposes?

pjcozzi:
The original intention might have been to avoid some `clone`s on the client, but we already using it.

pjcozzi:
Cesium tests are in https://github.com/AnalyticalGraphicsInc/cesium/pull/3616.

Thanks again, @jbo023.  I'm going to merge this and open another pull request to update this a bit more and also update the i3dm tile format.

arneschilling:
@pjcozzi : thanks a lot
@mramato : the in built JSON parser of the browser shouldnt complain about it. We just just want to avoid the allowed datatypes being whitelisted in Cesium. We need the client implementation to know the type in order to display it correctly, but that's not an issue.

PR #70, opened 2016-02-18
Color Comments

ggetz:
I added my comments below the TODO's.

The `color('white')` constructor I implemented is lowercase, similar to `rgb()`, `hsl()`, etc. but I can see why it could be argued to be uppercase. I changed all occurrences here to lowercase, but if I should change to implementation to uppercase, let me know.

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -43,29 +43,40 @@ All of the types except `Color` are derived from JavaScript.  `Color` is derived
 * `undefined`
 * `1.0`
 * `'Cesium'`
-* `Color('#00FFFF')`
+* `color('#00FFFF')`
 
 ### Color
 
 Colors are created with the following constructor functions:
-* `Color(keyword : String)`
-* `Color(6-digit-hex : String)`
-* `Color(3-digit-hex : String)`
+* `color(keyword : String)`
+* `color(6-digit-hex : String)`
+* `color(3-digit-hex : String)`
 * `rgb(red : Number, green : Number, blue : number)`
 * `rgba(red : Number, green : Number, blue : number, alpha : Number)`
 * `hsl(hue : Number, saturation : Number/Percentage, lightness : Number/Percentage)`
 * `hsla(hue : Number, saturation : Number/Percentage, lightness : Number/Percentage, alpha : Number)`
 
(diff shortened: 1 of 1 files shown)


pjcozzi:
OK, the convention for JavaScript constructor functions is Pascal case, as is the convention for JavaScript types, e.g., `Number(1)`.

ggetz:
So in that case should we also make `rgb()`, etc. Pascal case?

pjcozzi:
I meant to just open PRs into Cesium for things, not have the discussion in the .md file.  Moving the comments here:

> I changed all occurrences here to lowercase, but if I should change to implementation to uppercase, let me know

Make it uppercase throughout.

> If we want to go with Cesium behavior, an empty color() produces white. 

We should follow the CSS3 Colors spec.  See what it says.

> "color" : "rgb(255, 0, 0, (${Visible} ? 255 : 0))"
>
> we should not have the %. Also % do not parse wil jsep, which would make the % difficult to implement.

We would just pre-process them, but I agree I don't like them either and they don't fit into a strict grammar.

> **TODO: should `color` have an optional second argument for alpha, e.g., `Color('red', 0.5)`?  I think so.**
>
> // while that is not normal css bevhavior, I would say that is easy to implement and useful, so yes. 

CSS doesn't have a `Color` type in the same sense we do so this is still "CSS behavior" in that our Color type takes a CSS string and separate alpha.

> **TODO: is keyword case sensitive?**
>
> // According to W3, they are not case sensative

OK, just add  case-insensitive before the first type our spec says "keyword"

> **TODO: `rgb`, `hsl`, `rgba`, and `hsla` should require all their arguments (assuming that is true in CSS).**
>
> // Yes, currently I have a todo in the code to check for this.

Please do this soon so we can keep the spec and implementation in sync.

> So in that case should we also make rgb(), etc. Pascal case?

No, think of them as regular functions that return a `Color` object.

ggetz:
> I meant to just open PRs into Cesium for things, not have the discussion in the .md file.

Sorry, my confusion.

ggetz:
> rgb, hsl, rgba, and hsla should require all their arguments
> Please do this soon so we can keep the spec and implementation in sync.

OK

ggetz:
> We should follow the CSS3 Colors spec. See what it says.

The spec does not specifically mention empty values.

pjcozzi:
OK, white is fine.  Please update our spec in this branch.

ggetz:
Updated.

pjcozzi:
Nicely done!

PR #75, opened 2016-02-24
Added Section for Regular Expressions

ggetz:


pjcozzi:
Looks good!

Just curious, why the limitation?

> All arguments must be literals, not expressions.

In general, moving forward, we'll want to be able to pass expressions to functions, e.g., 
```
color.darken(${amount})
```

ggetz:
I mentioned it briefly in my PR. Basically if it's not a literal, it will be have to be evaluated at runtime, and I thought in pretty much all cases the regex will not be stored in a property or constructed with an expression, so we can save the extra computation time. 

If desired, I can add passing expressions to the regExp constructor.

pjcozzi:
> If desired, I can add passing expressions to the regExp constructo

Let's do it; otherwise, it is awkward that, for example, we can pass expressions to colors but not regexps.  Think about this as if it were JavaScript.

We can, of course, optimize the case if a literal is passed to regexp.  This is probably a good one to do right away.

ggetz:
OK, I'll make that change.

ggetz:
I opened a PR with the implementation changes, and made the correction to the spec. I also fixed the capitalization of the `regExp` function in the spec.

pjcozzi:
:+1:

PR #76, opened 2016-02-24
Changed Color constructor to color function

ggetz:
I updated all instances of the color function to lowercase, leaving references to the Color type uppercase.

I also removed/rephrased where we call the color function a constructor.


pjcozzi:
:+1:

PR #77, opened 2016-02-26
Updated Regex and Variable Sections

ggetz:
I updated the spec to clarify more on RegExp and Variables. The implementation for the regex `toString` function is to follow.

I also made a few spelling corrections.


pjcozzi:
Thanks!

PR #80, opened 2016-03-03
Style

pjcozzi:
Here's the first round of the [style spec](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/style/Styling/README.md).  Part of #2.

The scope will expand as we get more experience with this; there's unlimited potential here.  Any scope-expanding feedback in this PR will be added to #2 for consideration along with the rest of the roadmap.

The expression language is pretty well defined for this first cut, but for a formal spec, it will need a context-free grammar explicitly defined, full reference doc for a standard library, and whatever else a typical language spec needs.  We'll get there when we need to.

**TODO** before merging:

* [x] @ggetz - test all the examples in the spec.  Perhaps even add unit tests for them to Cesium.
* [x] @ggetz - slowly and carefully read the spec **and the schema** (I suspect there are corner cases with conditions that are not tested yet), and verify that Cesium and the spec are in-sync (the only exception for now can be "Variables can also be substituted inside strings defined with back-ticks") and that everything is reasonably specified so that someone else could implement the same behavior.
* [x] @pjcozzi - Add screenshots to the examples section.
* [x] @slchow - copyedit (will let you know when it is ready).

We'll generate the reference doc from the schema for the **Schema Reference** section once the schema is more battle tested.

Expect the related `3d-tiles-style -> 3d-tiles` pull request next week.


ggetz:
I've been adding the spec example tests in `Cesium3DTileStyleSpec`. Should they be in their own spec file?

pjcozzi:
:+1: Their own file is OK given the somewhat unique case here, instead of spreading them across several files for their specific classes.

pjcozzi:
@slchow I am going to add a screenshot or two, but this is ready for copyedit in the meantime.

Styling/README.md
```diff
@@ -0,0 +1,433 @@
...
+```
+
+```
+${enabled} === true
+${description} === null
+${order} === 1
+${name} === 'Feature name'
+```
+
+Variables can be used to constructor colors, for example:
```
slchow: "to constructor colors": is there a word missing here? Should it be "as constructor colors"?

Styling/schema/style.colorExpression.schema.json
```diff
@@ -0,0 +1,6 @@
+{
+    "$schema" : "http://json-schema.org/draft-03/schema",
+    "title" : "color expression",
+    "type" : "string",
+    "description" : "3D Tiles style expression that evaluates to a Color."
```
slchow: Does "Color" need to be capitalized here?

Styling/README.md
pjcozzi: Ah, should be "to construct colors"

Styling/schema/style.colorExpression.schema.json
pjcozzi: Yes, think of it like `Boolean` in this context.

pjcozzi:
@mramato if you have time, take a quick pass over [this](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/style/Styling/README.md).  Just keep in mind that we are not increasing the scope of this PR as described above, https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/80#issue-138219696.

Styling/README.md
pjcozzi: Fixed.

pjcozzi:
Merging this so the spec is all in one place for tomorrow.  This is still an early spec so feedback is still welcome, of course, in #2.

PR #84, opened 2016-03-11
RegExp Operators

ggetz:
Added section for `=~` and `!~` operators.

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -167,7 +167,7 @@ The following operators are supported with the same semantics and precedence as 
 
 * Unary: `+`, `-`, `!`
    * Not supported: `~`
-* Binary: `||`, `&&`, `===`, `!==`, `<`, `>`, `<=`, `>=`, `+`, `-`, `*`, `/`, `%`
+* Binary: `||`, `&&`, `===`, `!==`, `<`, `>`, `<=`, `>=`, `+`, `-`, `*`, `/`, `%`, `=~`, `!~` 
    * Not supported: `|`, `^`, `&`, `==`, `!=`, `<<`, `>>`, and `>>>`
 * Ternary: `? :`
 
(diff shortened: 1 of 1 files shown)


Styling/README.md
```diff
@@ -307,6 +307,19 @@ Regular expressions are treated as `NaN` when performing operations with any ope
 
 Regular expressions do not expose any other functions or a `prototype` object.
 
+The operators `=~` and `!=` are available to use with regular expressions. The `=~` operator matches the behavior of the `test` function, and tests the specified string for a match. It returns `true` if one is found, and `false` if not found. The `!=` operator is the inverse of the `=~` operator. It returns `true` if no matches are found, and `false` if a match is found. Both operators are communitive.
```
tfili: Don't you mean `!~` instead of `!=`?

ggetz: Yes, thanks!

pjcozzi:
Tweaked in a7635ec6bdf68b194c9f4ce072104319520839d4.

Please keep an eye out for old parts of the spec that need to be updated when we add something.  In this case, this needed to be updated:
> Regular expressions are treated as `NaN` when performing operations with any operator.

pjcozzi:
What do these do:
```
regExp('a') =~ regExp('a')
```
```
'abc' =~ 'abc'
```
And does the behavior match that of `test`?

ggetz:
`regExp('a') =~ regExp('a')` behaves like `regExp('a').test(regExp('a'))`. It will use the left regular expression as the object, and the right as the parameter. I'll add a line about this.

`'abc' =~ 'abc'` will return false since neither side is a regular expression. This is covered under: 

> If no `RegExp` is supplied as and operand, both operators will return `false`.

pjcozzi:
:+1:

PR #85, opened 2016-03-11
Added meta properties

ggetz:


Styling/README.md
```diff
@@ -53,11 +53,12 @@ Styles are defined with JSON and expressions written in a small subset of JavaSc
 
 ## Examples
 
-The following style assigns the default show and color properties to each feature:
+The following style assigns the default show, color, and meta properties to each feature:
```
pjcozzi: I would leave `meta` out here.

Styling/README.md
```diff
@@ -141,6 +142,27 @@ Since `expression` is optional and conditions are evaluated in order, the above
...
+    "meta" : {
+        "description" : "'Hello, ${featureName}.'"
+    }
+}
+```
+
+A meta property expression can evaluate to any type. For example:
+```json
+{
+    "meta" : {
```
pjcozzi: No need to change anything yet, but I almost feel like these are basically declaring variables and it would be cool to be able to use them in expressions like we do in `conditions`.  Let's see how it plays out.

pjcozzi:
Update the schema.

For lots of examples of JSON Schema (still version 3), see https://github.com/KhronosGroup/glTF/tree/master/specification/schema

ggetz:
I updated the schema.

Styling/schema/style.meta.schema.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema" : "http://json-schema.org/draft-03/schema",
+    "title" : "meta",
+    "type" : "object",
+    "description" : "A series of property names and the expression to evaluate for the value of each property.",
+    "properties" : {
+    },
+    "additionalProperties" : {
+        "$ref" : "style.metaProperty.schema.json"
```
pjcozzi: I _think_ you can delete `style.metaProperty.schema.json` and `style.expression.schema.json`, and replace them with a file like https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/Styling/schema/style.condition.condition.schema.json (named `style.meta.property.schema.json`)

ggetz: I think we would still need the `style.expression.schema.json` because it can evaluate to anything, which is different from `booleanExpression` and `colorExpression`.

pjcozzi: Ah, I see.  Can you make the description more precise since we have only defined boolean and color expressions and this could also be a constant string, expression returning a number, etc.

ggetz: I said "A valid 3D Tiles style expression", should I list every possible type as a possibility? Because it could also be a color object or a regex.

pjcozzi: OK, perhaps that is sufficient for this schema doc, but I still think we need to think through the interop between the styling language types (returned by the expression) and JavaScript/Cesium types.

What do you think about adding a small section to the spec about this?  Basically:
* `Boolean`, `Null`, `Undefined`, `Number`, and `String` map directly to JavaScript types.
* `Color` maps to `Cesium.Color`
* `RegExp` maps to what?  Converted to a string with its `toString` function?  Converted to a JavaScript `RegExp`?  Is in valid?

Issue #98, opened 2016-07-06
Per-tileset or per-tile transform


pjcozzi:
WGS84 is not appropriate for all tilesets.  For example, the inside of a building or a point cloud may be relative to local coordinate system.

This could be solved by adding a 4x4 `modelMatrix` (or `matrix` like glTF, or maybe `transform`) to `tileset.json`, which transforms the tileset's local coordinates to WGS84.  This would default to being the identity matrix, meaning that the local coordinate system is WGS84.

Alternatively, each tile in `tileset.json`could have a 4x4 `modelMatrix` that transforms to the parent's coordinate system (this would include the bounding volume), which provides more flexibility and is standard in the computer graphics community, but has a bit more complexity.  This is most likely the right approach to elegantly support heterogeneous tilesets of tilesets.

GatorScott:
A matrix won't suffice if the transformation is complex as in the U.S. State Plane coordinate system which can be one of two projected coordinate systems, Transverse Mercator or Lambert Conformal Conic.

pjcozzi:
The spirit of 3D Tiles is similar to the spirit of glTF - keep the client fast and simple (well, simple-ish) and put heavy work that can be done offline to the server so I would not encourage us adding more complex transforms that would, for example, require clients to include proj.4 in order to have a complete implementation.

pjcozzi:
@lilleyse for the spec, what do you think of an optional per-tile column-major 4x4 matrix named `transform` that is concatenated top-down at runtime just like glTF, including tilesets of tilesets?  It defaults to identity and the root tileset world space is still WGS84?

Example:
```javascript
{
  // ..
  "root": {
    "transform": [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 7000000, 0, 0, 1]
    "boundingVolume": {
      "box" : {
        // ...
      }
    },
    "content": {
      "boundingVolume": {
         // ...
      },
      "url": "0/0/0.b3dm"
    },
    "geometricError": 268.37878244706053,
    "refine": "add",
    "children": [
      {
        "boundingVolume": {
          "region": [
            // in WGS84 despite any transform properties
          ]
        },
        "content": {
          "boundingVolume": {
            // ...
          },
          "url": "1/0/0.b3dm"
        },
```

This would apply to the bounding volume and all geometry in tiles.  **However, it does not apply to region**, which is cartographic and fixed in WGS84.

[...]

PR #100, opened 2016-07-08
I3dm updates

lasalvavida:
@pjcozzi these are the i3dm spec changes. Can you take a look over this and tell me what you think?


lasalvavida:
#33

pjcozzi:
@lasalvavida did you draw all the figures?  If not, what is their terms of use?

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,25 +20,39 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
 
 A tile is composed of a header section immediately followed by a body section.
 
+
 **Figure 1**: Instanced 3D Model layout (dashes indicate optional fields).
 
-![](figures/layout.png)
+![header layout](figures/header-layout.png)
```
pjcozzi: In the Cesium implementation `span` is called `scale`.  What is the standard terminology for quantization?  Let's use it in all cases.

pjcozzi: Same comment with `origin` and `translation`.

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,25 +20,39 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
...
+
+![body layout](figures/body-layout.png)
+
+![instance layout](figures/instance-layout.png)
```
pjcozzi: What is the canonical name for `v1` and `v2` in the graphics literature?  Perhaps `i` and `j` are better.  I dunno.

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,25 +20,39 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
...
+| Field name | Data type | Description |
+| --- | --- | --- |
 | `magic` | 4-byte ANSI string | `"i3dm"`.  This can be used to identify the arraybuffer as an Instanced 3D Model tile. |
 | `version` | `uint32` | The version of the Instanced 3D Model format. It is currently `1`. |
 | `byteLength` | `uint32` | The length of the entire tile, including the header, in bytes. |
 | `batchTableByteLength` | `uint32` | The length of the batch table in bytes. Zero indicates there is not a batch table. |
 | `gltfByteLength` | `uint32` | The length of the glTF field in bytes. |
 | `gltfFormat` | `uint32` | Indicates the format of the glTF field of the body.  `0` indicates it is a url, `1` indicates it is embedded binary glTF.  See the glTF section below. |
 | `instancesLength` | `uint32` | The number of instances, also called features, in the tile. |
+| `origin.x` | `double` | The `x` coordinate of the origin for the instance region. |
```
pjcozzi: Here and below, what is the "instance region?"  Is it different than a tile?  Perhaps just call this tile.  The spec needs to be very precise so we need to be careful about not implicitly introducing new terms, especially if they are redundant.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
 
 The `instances` field immediately follows the `glTF` field (which may be omitted when `header.gltfByteLength` is `0`).
 
-The `instances` field contains `header.instancesLength` of tightly packed instances.  Each instance has three fields:
+The `instances` field contains `header.instancesLength` of tightly packed instances.  Each instance has the following fields:
+
+| Field name | Data type | Description | Required |
+| --- | --- | --- | --- |
+| `position.x` | `uint16` | The x-coordinate in quantized cartesian coordinates. | `yes` |
```
pjcozzi: Throughout: "cartesian" -> "Cartesian"

pjcozzi:
Add a **Resources** section at the bottom that links to the papers for quantization and oct-endcoding.  Also link to the Cesium helper functions implementation.

pjcozzi:
Just like the Cesium implementation, this needs:
* [ ] `TILES3D_SCALE` and `TILES3D_NON_UNIFORM_SCALE`, https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/33#issuecomment-227208405

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+| `position.z` | `uint16` | The z-coordinate in quantized cartesian coordinates. | `yes` |
+| `v1.x` | `uint` | The first component of the oct-encoded vector `v1`. | `yes` |
+| `v1.y` | `uint` | The second component of the oct-encoded vector `v1`. | `yes` |
+| `v2.x` | `uint` | The first component of the oct-encoded vector `v2`. | `yes` |
+| `v2.y` | `uint` | The second component of the oct-encoded vector `v2`. | `yes` |
+| `batchId` | `uint16` | ID in the range `[0, length of arrays in the Batch Table)`, which indicates the corresponding properties. | `if header.batchTableByteLength > 0`
+
+### X, Y, and Z for Translation
+
+`x`, `y`, and `z` are stored as `uint16` positions in the quantized instance region defined by the `origin` and `span` fields in the header.
```
pjcozzi: The definition here for "quantized instance region" should come earlier in the spec, and how it does/doesn't relate to the bounding volume should be explicit to avoid confusion; readers may think they are the same at first glance.

Also, "quantized volume" is a better name since instance is implicit in this context and the reader won't need to wonder if there is a difference between volume and region; it is the prefix, "bounding" or "quantized", that defines that.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+
+**Figure 2:** The instance region defined by `origin` and `span`
+
+![](figures/instance-region.png)
+
+Transforming `position` in instance region space to `position_w` in world space can be done using the formula:
```
pjcozzi: Let's not say "world space" here; "model space" would be fine even if model == world for now.  If we say world, we will need to update this later for #98.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+
+`position_w` = `origin` + [`position` * `span` / (`2^16-1`)].
+
+### V1, and V2 for Rotation
```
pjcozzi: Drop the comma.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+`v1` and `v2` are stored as two components in oct-encoded format as described in
+[*A Survey of Efficient Representations of Independent Unit Vectors* by Cigolle et al.](http://jcgt.org/published/0003/02/01/)
+The [AttributeCompression](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js)
+module in Cesium contains an implementation for encoding and decoding vectors in this fashion.
+
+In their decoded form, `v1` and `v2` establish a new orthonormal basis, effectively rotating the coordinate system.
+
+The `y` and `x` vectors in the natural basis are transformed to map onto `v1` and `v2` respectively.
+
+**Figure 3:** A 3D box in the natural basis
```
pjcozzi: The perspective is awkward in this figures.  Is it possible to generate them using an orthogonal projection?

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+### V1, and V2 for Rotation
+
+`v1` and `v2` are stored as two components in oct-encoded format as described in
+[*A Survey of Efficient Representations of Independent Unit Vectors* by Cigolle et al.](http://jcgt.org/published/0003/02/01/)
+The [AttributeCompression](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js)
+module in Cesium contains an implementation for encoding and decoding vectors in this fashion.
+
+In their decoded form, `v1` and `v2` establish a new orthonormal basis, effectively rotating the coordinate system.
+
+The `y` and `x` vectors in the natural basis are transformed to map onto `v1` and `v2` respectively.
```
pjcozzi: Wouldn't it be more natural for readers to map x/y, not y/x.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +112,54 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+**Figure 4:** A 3D box in the rotated basis
 
-_TODO: make this much more memory efficient and more general, [#33](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/33)._
+![](figures/box-rotated-basis.png)
 
-When `header.batchTableByteLength` is zero, which indicates there is not a batch table, `batchId` is omitted, so each instance contains only `longitude` and `latitude` fields.
+The mapping for `z` can be omitted since it will be the cross product of `v1` and `v2`.
 
-Each instance is in the east-north-up reference frame (`x` points east, `y` points north, and `z` points along the geodetic surface normal).
+A rotation matrix can be created from `v1`, `v2`, and `v3`=`v2`x`v1` by placing `v2` into the first column,
```
pjcozzi: Can this include simple equations like, for example, https://github.com/tsturm/glTF/blob/master/extensions/Vendor/FRAUNHOFER_materials_pbr/Appendix.md#specular-brdf

pjcozzi:
Good start here, @lasalvavida!

pjcozzi:
Also:
* [ ] Investigate how AOS (XYZ, XYZ, XYZ) vs. SOA (XXX, YYY, ZZZ) affects gzip compression

pjcozzi:
Also, change the "Spec Status" for i3dm to "Solid base, only minor changes expected"

See https://github.com/AnalyticalGraphicsInc/3d-tiles#spec-status

arneschilling:
Isn't the definition of v1 and v2 as 2D vectors too restrictive? In this case a single rotation angle would suffice, because the rotation will always be around the z axis. Can we have 6 DOF as Quaternions or Euler angles?

arneschilling:
is the UP vector and axis order somewhere defined in glTF? I know that it can be set in COLLADA specifically

lasalvavida:
> @lasalvavida did you draw all the figures? If not, what is their terms of use?

I drew all of the figures

TileFormats/Instanced3DModel/README.md
lasalvavida: They would be `up` and `right` I think.

TileFormats/Instanced3DModel/README.md
lasalvavida: The "instance region" is different from a tile. It is a volume defined by the `origin` and `span` defined in the header. This doesn't have to be the same as the tile's bounding volume or the content bounding volume, though it's probably worth adding that the instance region should be contained by the tile volume.

TileFormats/Instanced3DModel/README.md
lasalvavida: I agree, quantized volume is a better name.

TileFormats/Instanced3DModel/README.md
lasalvavida: They're done in inkscape which is not the best for generating 3D figures. I'll see what I can do.

TileFormats/Instanced3DModel/README.md
lasalvavida: Maybe. The only reason I did it this way was to map `up` first for orientation and then `right` for rotation.

lasalvavida:
> is the UP vector and axis order somewhere defined in glTF? I know that it can be set in COLLADA specifically

@arneschilling I'm pretty sure it's Y up and right hand rule.

> Isn't the definition of v1 and v2 as 2D vectors too restrictive?

v1 and v2 are not 2D vectors, they are 3D vectors. They are stored as two components each because they are oct-encoded. 

> In this case a single rotation angle would suffice, because the rotation will always be around the z axis. Can we have 6 DOF as Quaternions or Euler angles?

You can define any orientation using an orthonormal basis. It is also easy to find for the common use case. `v1` is usually just `up` relative to `WGS84` and `v2` is usually just `UNIT_Z` crossed with `v1` so that it is orthogonal to `up` and coplanar with `UNIT_X`. You can rotate `UNIT_Z` around `UNIT_Y` before crossing for simple rotations. 

That is the most common use case, and it can be stored in only 4 bytes. Other more complex orientations are still possible, you just might have to do a little more math. However, if there's a very strong argument for using some other representation, I'd be happy to hear it.

lasalvavida:
I should add, it makes the implementation dead simple:

```javascript
// Decode compressed normals
AttributeCompression.octDecode(normalOneX, normalOneY, normalUp);
AttributeCompression.octDecode(normalTwoX, normalTwoY, normalRight);

// Compute third normal
Cartesian3.cross(normalRight, normalUp, normalOut);

// Place the basis into the rotation matrix
Matrix3.setColumn(rotation, 0, normalRight, rotation);
Matrix3.setColumn(rotation, 1, normalUp, rotation);
 Matrix3.setColumn(rotation, 2, normalOut, rotation);
```

TileFormats/Instanced3DModel/README.md
lasalvavida: I'm open to whatever you think is best as far as naming goes. 

`span` is a little different from `scale` in quantization. In quantization since we store everything in a decode matrix `scale` is stored as `span` divided by `2^16-1`. However, here we can choose not do that and potentially improve precision which is what I am doing currently.

TileFormats/Instanced3DModel/README.md
pjcozzi: Double check, and let's go with those semantics since I think they provide more intuition.

pjcozzi:
> I should add, it makes the implementation dead simple...

I think it is worth including this code snippet in the spec, with a link to the implementation in the figure's description.

This is going to be one of the most common questions when someone implements i3dm.

TileFormats/Instanced3DModel/README.md
pjcozzi: So it sounds like `scale` is the right term here.  Is `origin` the standard quantization term used in the literature?

lasalvavida: By the literature,  `origin` should be `offset` and `span` should be `scale`. I'll make those changes.

lasalvavida:
> Investigate how AOS (XYZ, XYZ, XYZ) vs. SOA (XXX, YYY, ZZZ) affects gzip compression

I just tried AOS vs SOA for InstancedWithBatchTable (512 KB) and gzipped AOS is 101KB and gzipped SOA is 68KB, so it seems like SOA is probably the way to go, @pjcozzi with your go-ahead, I'll make this change in the implementation and the spec.

TileFormats/Instanced3DModel/README.md
lasalvavida: I'll change this to read as `right`/`up`/`forward` which seems to be the common notation.

pjcozzi:
> I just tried AOS vs SOA for InstancedWithBatchTable (512 KB) and gzipped AOS is 101KB, gzipped SOA is 68KB, so it seems like SOA is probably the way to go, @pjcozzi with your go-ahead, I'll make this change in the implementation and the spec.

Sounds great, thanks for investigating!

pjcozzi:
**TODO** discussed offline
* [ ] Evaluate quaternion for rotation as @bagnell suggested

lasalvavida:
>  Evaluate quaternion for rotation as @bagnell suggested

A quaternion stored as axis-angle (3 bytes) works and isn't too complex. Will update the spec.

lasalvavida:
@pjcozzi This is ready for another look

pjcozzi:
> A quaternion stored as axis-angle (3 bytes) works and isn't too complex. Will update the spec.

Do you have before/after size numbers (gzipped)?

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,30 +20,53 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
 
 A tile is composed of a header section immediately followed by a body section.
 
+Instances are organized as a structure of arrays where the fields from each instance are grouped together.
```
pjcozzi: "structure of arrays" probably change to "Structure of Arrays (SOA)"

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,30 +20,53 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
...
+| `offset.x` | `double` | The `x` coordinate of the offset for the quantized volume. |
+| `offset.y` | `double` | The `y` coordinate of the offset for the quantized volume. |
+| `offset.z` | `double` | The `z` coordinate of the offset for the quantized volume. |
+| `scale.x` | `double` | The length in the `x` direction covered by the quantized volume. |
+| `scale.y` | `double` | The length in the `y` direction covered by the quantized volume. |
+| `scale.z` | `double` | The length in the `z` direction covered by the quantized volume. |
 
-_TODO: Link to Cesium code for reading header_
+Code for reading the header can be found in
+[Instanced3DModelTileContent](https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Instanced3DModel3DTileContent.js#L170)
```
pjcozzi: "Instanced3DModelTileContent" -> "Instanced3DModelTileContent.js"

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,30 +20,53 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
...
+in the Cesium implementation of 3D Tiles.
 
 If either `gltfByteLength` or `instancesLength` equal zero, the tile does not need to be rendered.
 
 The body section immediately follows the header section, and is composed of three fields: `Batch Table`, `glTF`, and `instances`.
```
pjcozzi: This should be the last sentence in the section (since it transitions from header to body) so move the new **Quantized Volume** section above it.

TileFormats/Instanced3DModel/README.md
```diff
@@ -20,30 +20,53 @@ Instanced 3D Model maps well to the [ANGLE_instanced_arrays](https://www.khronos
...
 
+## Quantized Volume
+
+Instances are placed by their position relative to the quantized volume defined by the `offset` and `scale` fields in the header.
+
+**Figure 2:** The quantized volume defined by `offset` and `scale`.
+
+![](figures/instance-region.png)
```
pjcozzi: Wouldn't this image be better with the `x` vector coming directly from the offset?  I know that it might be harder to draw, but it will be less confusing.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+Each instance has the following fields:
+
+| Field name | Data type | Description | Required |
+| --- | --- | --- | --- |
+| `position.x` | `uint16` | The x-coordinate in quantized Cartesian coordinates. | `yes` |
+| `position.y` | `uint16` | The y-coordinate in quantized Cartesian coordinates. | `yes` |
+| `position.z` | `uint16` | The z-coordinate in quantized Cartesian coordinates. | `yes` |
+| `encodedAxis.x` | `uint` | The first component of the oct-encoded `axis` vector. | `yes` |
+| `encodedAxis.y` | `uint` | The second component of the oct-encoded `axis` vector. | `yes` |
+| `encodedAngle` | `uint` | The angle to rotate about the axis from `0` to `2π` scaled to fill a single byte. | `yes` |
```
pjcozzi: @bagnell @lasalvavida are you guys absolutely positive about this representation?  How lossy is it compared to oct-encoding two normals?  Only being able to represent 255 angles sounds **way** too limited to define precise rotations, which we must be able to faithfully represent.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
 
 The `instances` field immediately follows the `glTF` field (which may be omitted when `header.gltfByteLength` is `0`).
 
-The `instances` field contains `header.instancesLength` of tightly packed instances.  Each instance has three fields:
+The `instances` field contains `header.instancesLength` of each instance field which are stored in groups of each field.
```
pjcozzi: "are stored in groups of each field" -> this needs more explaining, perhaps a small example with a diagram would help, e.g. show `X0X1X2 Y0Y1Y2 Z0Z1Z2` maps to
* instance: X0, Y0, Z0
* instance: X1, Y1, Z1
* instance: X2, Y2, Z2

TileFormats/Instanced3DModel/README.md
pjcozzi: Also say why since this is a non-obvious representation to must developers.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+| `encodedAxis.x` | `uint` | The first component of the oct-encoded `axis` vector. | `yes` |
+| `encodedAxis.y` | `uint` | The second component of the oct-encoded `axis` vector. | `yes` |
+| `encodedAngle` | `uint` | The angle to rotate about the axis from `0` to `2π` scaled to fill a single byte. | `yes` |
+| `batchId` | `uint16` | ID in the range `[0, length of arrays in the Batch Table)`, which indicates the corresponding properties. | `if header.batchTableByteLength > 0`
+
+### Position for Translation
+
+`position.x`, `position.y`, and `position.z` are stored as `uint16` positions in the [quantized volume](#quantized-volume) defined by the `offset` and `scale` fields in the header.
+
+Transforming `position` in instance region space to `position_m` in model space can be done using the formula:
```
pjcozzi: "instance region space" -> is this old?

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
 
-|Field name|Data type|Description|
-|----------|---------|-----------|
-| `longitude` | `double` | The longitude, in radians, in the range `[-PI, PI]`. |
-| `latitude` | `double` | The latitude, in radians, in the range `[-PI / 2, PI / 2]`. |
-| `batchId` | `uint16`  | ID in the range `[0, length of arrays in the Batch Table)`, which indicates the corresponding properties. |
+This format can be easily converted to a quaternion or [a rotation matrix](#creating-a-rotation-matrix-for-an-instance).
 
-_TODO: make this much more memory efficient and more general, [#33](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/33)._
+### Scaling using TILES3D_SCALE and TILES3D_NON_UNIFORM_SCALE
```
pjcozzi: This could be a sub-title for a section called "Batch Table Semantics."  Each tile format spec will have a section like this.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
 
-When `header.batchTableByteLength` is zero, which indicates there is not a batch table, `batchId` is omitted, so each instance contains only `longitude` and `latitude` fields.
+Scaling can be done using the special batch semantics `TILES3D_SCALE` and `TILES3D_NON_UNIFORM_SCALE`.
```
pjcozzi: "special" -> "well known"

"batch semantics" -> "batch table semantics"

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
 
-Each instance is in the east-north-up reference frame (`x` points east, `y` points north, and `z` points along the geodetic surface normal).
+`TILES3D_SCALE` is a number stored for a `batchId` in the batch table.
```
pjcozzi: Precisely define the datatype: `float32` is OK.

TileFormats/Instanced3DModel/README.md
```diff
@@ -98,19 +121,51 @@ In either case, `header.gltfByteLength` contains the length of the glTF field in
...
+If present, all instances with that `batchId` will have their dimensions scaled by the number.
+
+`TILES3D_NON_UNIFORM_SCALE` is an array of numbers stored for a `batchId` in the batch table.
```
pjcozzi: Same `float32` comment.

Also, this is currently AOS, right?  With @lilleyse upcoming changes, we can make this SOA.

pjcozzi:
In the previous review, I asked for equations to show how to construct the 4x4 matrix.  We still need these and they should include the scale semantics and how the matrices are multiplied together.

TileFormats/Instanced3DModel/README.md
pjcozzi: What happens if both semantics are present?  As is, this is undefined behavior; the root of all spec evil.  We need to precisely define this.  I think it would be fine to just define the order in which the matrices are multiplied.  When only one semantic is defined (which will be the 99.9999% case), then it is as if the other is identity.  In practice, this matrix multiple would be optimized out in reasonable implementations.

TileFormats/Instanced3DModel/README.md
lasalvavida: Yes, it must have slipped past me

TileFormats/Instanced3DModel/README.md
lasalvavida: Actually the way I have it now is that they are both applied if present. So you can have a non-uniform scale and a uniform scale and they get applied on top of each other. Order doesn't matter for multiplication since they're in Jordan form.

lasalvavida: If that's agreeable, I'll update the spec to state that more explicitly.

TileFormats/Instanced3DModel/README.md
lasalvavida: We have a precision of ~1.4 degrees for rotation with axis-angle. @bagnell what do you think? If we need more precision, is it better to make the angle a `uint16` and get ~0.005 degrees or to go back to two vectors?

lasalvavida:
> A quaternion stored as axis-angle (3 bytes) works and isn't too complex. Will update the spec.
> Do you have before/after size numbers (gzipped)?

| | Raw | Gzipped |
| --- | --- | --- |
| SOA (2 Normals) | 512 KB | 101 KB |
| AOS (2 Normals) | 512 KB | 55 KB |
| AOS (axis-angle) | 503 KB | 69 KB |

So, gzipped axis-angle is actually bigger than 2 Normals. @pjcozzi How do you want to proceed?

TileFormats/Instanced3DModel/README.md
pjcozzi: Sounds perfect!

In general, look out for corner cases like this and make them explicit in the spec.  It is very common for specs to miss a lot of details, and then the "spec" becomes the first or most popular implementation.  We want to avoid - or at least minimize - this.

pjcozzi:
Why do you think axis-angle doesn't compress as well?  Did you analyze the data?  Is it because it is not as coherent (e.g., adjacent values are very different) or because of byte alignment?

Once you have a solid understanding or good hunch, it looks like we have evident that stay with 2 normals.

> We have a precision of ~1.4 degrees for rotation with axis-angle. @bagnell what do you think? If we need more precision, is it better to make the angle a uint16 and get ~0.005 degrees or to go back to two vectors?

~1.4 degrees is not nearly enough precision.  For fine-grained details, this would be drastically off.  Do we have a sense of the precision for the oct-encoded normals?  Oct-encoding is great for lighting where we can be a bit off, but we need to be precise here.  The paper has precision details, http://jcgt.org/published/0003/02/01/

bagnell:
From the paper, we use oct16P encoding which has a mean error of `0.31485` and a max error of `0.63575`. Those values are degrees.

lasalvavida:
@pjcozzi, I followed some suggestions from @bagnell about improving precision for axis-angle using `snorm`, but it still compresses to the same 69 KB.

I padded the angle to make the axis vectors byte-aligned, and it didn't change the compressed size at all. Looking at the data, it seems like there are more repeated data segments in the 2 normals as opposed to axis-angle. That could change with a different data set perhaps, but my test data set is composed of objects rotated around their `up` vector which I suspect will be the most common use case.

I think reverting back to the two-normals approach would be best; and I'll include some justification as to why in the specification.

pjcozzi:
> my test data set is composed of objects rotated around their up vector which I suspect will be the most common use case

Yes, for most geospatial use cases, this will be the case; however, it will not be the case for CAD-type assets.

> it seems like there are more repeated data segments in the 2 normals

Is this because the oct16P encoding isn't that precise and is basically dropping values into buckets?

To move forward, I suggest we determine
* [ ] How precise is @bagnell's `snorm` representation?
* [ ] Are there other encoding strategies that we can use from the JCGT paper since `[0.31485, 0.63575]` degrees is way too much.  I should have realized this from the start.  Size-wise, how do these compare to @bagnell's approach?
* [ ] Unless we strike gold, we should consider not storing the nromal/quat in the instances, and instead introduce a batch table semantic, e.g., `3DTILES_WHATEVER_ENCODED_NORMAL`, so that .i3dm allows different normal compression (maybe also include an uncompressed one).  I guess we could default to east-north-up of the instance's position if this semantic is undefined.  This would actually save a lot of data for a common case at the cost of some more CPU time at tile load.

lasalvavida:
>How precise is @bagnell's snorm representation?

I looked at the cesium code for this and it is just linearly mapping onto a byte, so it really isn't any more precise than what I was already doing.

> Are there other encoding strategies that we can use from the JCGT paper

oct24P `[0.01953, 0.03928]` and oct32P `[0.00122, 0.00246]` seem to be the best performers in the `24-bit` and `32-bit` categories

>  I guess we could default to east-north-up of the instance's position if this semantic is undefined. This would actually save a lot of data for a common case at the cost of some more CPU time at tile load.

I would do something along the lines of
`3DTILES_UP_NORMAL`
`3DTILES_UP_NORMAL_OCT16P`
`3DTILES_UP_NORMAL_OCT24P`

`3DTILES_RIGHT_NORMAL`
...

That way if you just want to do rotation about the `up` vector, you can leave `3DTILES_UP_NORMAL` undefined to be east/north/up and define `3DTILES_RIGHT_NORMAL`.

lasalvavida:
While we're talking about precision; quantization may have some issues as well; if your tiles scale in a direction is 100,000 meters, positions for instances your instances may be off by ~1.5 meters

pjcozzi:
> Small quantized volumes are fine, but as the volume grows, precision drops.

This is fine for most use cases since as the viewer zooms in, the bounding box gets smaller and smaller for volumes down the tree.

However, there are reasonable use cases for have one big `i3dm` file, e.g., for all the trees in Philadelphia, so let's add a `positionsFormat` to the header that can be `UNCOMPRESSED` or `QUANTIZED`.  Then move the offset/scale out of the header and only have it if `positionsFormat` is `QUANTIZED`.

pjcozzi:
For rotation, let's go with batch table semantics:
* One that is oct32P 
* One that is uncompressed

We can add more as prominent use cases arrive, but we shouldn't start with too many since it complicates the spec, increases the barrier to entry to write a valid loader, and will make the standardization process harder.

If no rotation semantic is defined, go with east/north/up.  We'll consider adding a semantic for the rotation around north as you suggested, but I want to keep it simple for now and not add too many fast paths until needed.

pjcozzi:
>> Small quantized volumes are fine, but as the volume grows, precision drops.

> This is fine for most use cases since as the viewer zooms in, the bounding box gets smaller and smaller for volumes down the tree.

> However, there are reasonable use cases for have one big i3dm file, e.g., for all the trees in Philadelphia, so let's add a positionsFormat to the header that can be UNCOMPRESSED or QUANTIZED. Then move the offset/scale out of the header and only have it if positionsFormat is QUANTIZED.

Later, we'll do the same for point clouds; it will be a less common use case to have a single large tile, but some users can't tolerate any lossyness.

lasalvavida:
> If no rotation semantic is defined, go with east/north/up. We'll consider adding a semantic for the rotation around north as you suggested, but I want to keep it simple for now and not add too many fast paths until needed.

I'm not sure what you're saying here. east/north/up is the orientation. I'm saying that you can override either the `up` and `right` vector in that default orientation with the semantics in the batch table. Overriding the `right` vector gives you rotation, overriding the `up` vector gives you tilt.

lasalvavida:
@pjcozzi This is ready for a look

pjcozzi:
> TileFormats/Instanced3DModel/figures/body-layout.png

Can you explicitly show the JSON and binary part of each table, e.g., like https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Points#layout

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,129 @@ A tile is composed of a header section immediately followed by a body section.
...
 
-|Field name|Data type|Description|
-|----------|---------|-----------|
+| Field name | Data type | Description |
+| --- | --- | --- |
 | `magic` | 4-byte ANSI string | `"i3dm"`.  This can be used to identify the arraybuffer as an Instanced 3D Model tile. |
 | `version` | `uint32` | The version of the Instanced 3D Model format. It is currently `1`. |
 | `byteLength` | `uint32` | The length of the entire tile, including the header, in bytes. |
-| `batchTableByteLength` | `uint32` | The length of the batch table in bytes. Zero indicates there is not a batch table. |
+| `featureTableByteLength` | `uint32` | The length of the feature table in bytes. |
```
pjcozzi: For feature and batch tables, don't we need to store both the JSON length and binary length like @lilleyse did for `pnts`?

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,129 @@ A tile is composed of a header section immediately followed by a body section.
...
+#### Instance Semantics
 
-The properties for the instance with `batchId = 0` are
-```javascript
-id[0] = 'unique id';
-displayName[0] = 'Tree species';
-yearBuilt[0] = 1999;
-location[0] = {x : 1, y : 2};
-```
+These semantics map to an array of feature values that are used to create instances. The length of these arrays must be the same for all semantics.
```
pjcozzi: Let's be explicit that length is number of instances, not byte length.

(in general this is what "length" means in 3D Tiles as opposed to "byteLength", but I think being extra explicit is good here).

TileFormats/Instanced3DModel/README.md
lasalvavida: Yes, updating now

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,129 @@ A tile is composed of a header section immediately followed by a body section.
...
-The properties for `batchId = 1` are
-```javascript
-id[1] = 'another unique id';
-displayName[1] = 'Another tree species';
-yearBuilt[1] = 2003;
-location[1] = {x : 3, y : 4};
-```
+If a semantic has a dependency on another semantic, that semantic must be defined in order to be used.
+
+| Semantic | Data Type | Dependencies | Description |
```
pjcozzi: Instead of `none` for dependencies; it would be cleaner to leave it blank.

pjcozzi: This table should indicate that `POSITION` or `POSITION_QUANTIZED` is required.

For inspiration, see tables in glTF: https://github.com/KhronosGroup/glTF/blob/master/specification/README.md#shader

In general, feel free to make judicial use of http://www.emoji-cheat-sheet.com/

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,129 @@ A tile is composed of a header section immediately followed by a body section.
...
+| --- | --- | --- | --- |
+| `POSITION` | `float32[3]` | `none` | A 3-component array of numbers containing `x`, `y`, and `z` Cartesian coordinates for the position of the instance. |
+| `POSITION_QUANTIZED` | `uint16[3]` | `QUANTIZED_VOLUME_OFFSET`, `QUANTIZED_VOLUME_SCALE` | A 3-component array of numbers containing `x`, `y`, and `z` in quantized Cartesian coordinates for the position of the instance. |
+| `NORMAL_UP` | `float32[3]` | `NORMAL_RIGHT` | A unit vector defining the `up` direction for the orientation of the instance. |
+| `NORMAL_UP_OCT32P` | `uint16[2]` | `NORMAL_RIGHT_OCT32P` | An oct-encoded unit vector with 32-bits of precision defining the `up` direction for the orientation of the instance. |
+| `NORMAL_RIGHT` | `float32[3]` | `NORMAL_UP` | A unit vector defining the `right` direction for the orientation of the instance. Must be orthogonal to `up`. |
```
pjcozzi: Probably swap this with `NORMAL_RIGHT` above so related fields are together.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,129 @@ A tile is composed of a header section immediately followed by a body section.
...
-id[1] = 'another unique id';
-displayName[1] = 'Another tree species';
-yearBuilt[1] = 2003;
-location[1] = {x : 3, y : 4};
-```
+If a semantic has a dependency on another semantic, that semantic must be defined in order to be used.
+
+| Semantic | Data Type | Dependencies | Description |
+| --- | --- | --- | --- |
+| `POSITION` | `float32[3]` | `none` | A 3-component array of numbers containing `x`, `y`, and `z` Cartesian coordinates for the position of the instance. |
```
pjcozzi: Throughout if SOA is better, we can define this semantic to use that (same for normals).

pjcozzi:
Just those comments.  I really dig this new organization.  Clean, logical, and efficient.

lasalvavida:
Updated.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+| Field name | Data type | Description |
+| --- | --- | --- |
 | `magic` | 4-byte ANSI string | `"i3dm"`.  This can be used to identify the arraybuffer as an Instanced 3D Model tile. |
 | `version` | `uint32` | The version of the Instanced 3D Model format. It is currently `1`. |
 | `byteLength` | `uint32` | The length of the entire tile, including the header, in bytes. |
-| `batchTableByteLength` | `uint32` | The length of the batch table in bytes. Zero indicates there is not a batch table. |
+| `featureTableJSONByteLength` | `uint32` | The length of the feature table JSON section in bytes. |
+| `featureTableBinaryByteLength` | `uint32` | The length of the feature table binary section in bytes. |
+| `batchTableJSONByteLength` | `uint32` | The length of the batch table JSON section in bytes. Zero indicates that there is no batch table. |
+| `batchTableBinaryByteLength` | `uint32` | The length of the batch table binary section in bytes.
```
pjcozzi: "The length of the batch table binary section in bytes." -> "The length of the batch table binary section in bytes.  If `batchTableJSONByteLength` is zero; this will also be zero."

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+These semantics map to an array of feature values that are used to create instances. The length of these arrays must be the same for all semantics and is equal to the number of instances.
 
-The properties for `batchId = 1` are
-```javascript
-id[1] = 'another unique id';
-displayName[1] = 'Another tree species';
-yearBuilt[1] = 2003;
-location[1] = {x : 3, y : 4};
-```
+If a semantic has a dependency on another semantic, that semantic must be defined in order to be used.
```
pjcozzi: "that semantic must be defined in order to be used" -> "that semantic must be defined in order to be a valid tile"

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+| `SCALE_NON_UNIFORM` | `float32[3]` | `none` | A 3-component array of numbers defining the scale to apply to the `x`, `y`, and `z` axes of the instance. | No |
+| `BATCH_ID` | `unit16` | `none` | The `batchId` of the instance that can be used to retrieve metadata from the `Batch Table`. | No |
+
+#### Global Semantics
+
+These semantics define global properties for all instances.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `INSTANCES_LENGTH`| `uint32` | The number of instances to generate. The length of each array value for an instance semantic should be equal to this. | Yes |
```
pjcozzi: Add `:white_check_mark:`

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
-The properties for `batchId = 1` are
-```javascript
-id[1] = 'another unique id';
-displayName[1] = 'Another tree species';
-yearBuilt[1] = 2003;
-location[1] = {x : 3, y : 4};
-```
+If a semantic has a dependency on another semantic, that semantic must be defined in order to be used.
+
+| Semantic | Data Type | Dependencies | Description | Required |
```
pjcozzi: Consider replacing the `Dependencies` column with `Required`, e.g., "this is required when that is defined"

Consider using a different color check.

I _think_ this will read better, but try and see.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
-If either `gltfByteLength` or `instancesLength` equal zero, the tile does not need to be rendered.
+The body section immediately follows the header section, and is composed of three fields: `Feature Table`, `Batch Table`, and `glTF`.
 
-The body section immediately follows the header section, and is composed of three fields: `Batch Table`, `glTF`, and `instances`.
+Code for reading the header can be found in
+[Instanced3DModelTileContent](https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Instanced3DModel3DTileContent.js#L170)
+in the Cesium implementation of 3D Tiles.
 
-## Batch Table
+## Feature Table
```
pjcozzi: This section could use a few examples, e.g., Feature table with
* Just positions (minimal use case)
* Quantized positions and oct-encoded normals (typical use case)
* Quantized positions, oct-encoded normals, non-uniform scale, and batch ids (full flexibility use case)

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `INSTANCES_LENGTH`| `uint32` | The number of instances to generate. The length of each array value for an instance semantic should be equal to this. | Yes |
+| `QUANTIZED_VOLUME_OFFSET` | `float32[3]` | A 3-component array of numbers defining the offset for the quantized volume. | No |
+| `QUANTIZED_VOLUME_SCALE` | `float32[3]` | A 3-component array of numbers defining the scale for the quantized volume. | No |
+
+### Instance Orientation
+
+An instance's orientation is defined by an orthonormal basis created by an `up` and `right` vector. If `NORMAL_UP` and `NORMAL_RIGHT` are not present,
+the instance will default to `east/north/up` for its position projected onto `WGS84`.
```
pjcozzi: Change to "the instance will default to the `east/north/up` reference frame's orientation for the instance's Cartographic position (`x`, `y`, `z` converted to `longitude` and `latitude` on the WGS84 ellipsoid)"

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `INSTANCES_LENGTH`| `uint32` | The number of instances to generate. The length of each array value for an instance semantic should be equal to this. | Yes |
+| `QUANTIZED_VOLUME_OFFSET` | `float32[3]` | A 3-component array of numbers defining the offset for the quantized volume. | No |
+| `QUANTIZED_VOLUME_SCALE` | `float32[3]` | A 3-component array of numbers defining the scale for the quantized volume. | No |
+
+### Instance Orientation
+
+An instance's orientation is defined by an orthonormal basis created by an `up` and `right` vector. If `NORMAL_UP` and `NORMAL_RIGHT` are not present,
```
pjcozzi: Mention `_OCT32P` versions.

pjcozzi:
@lasalvavida this looks ready to implement!  Nice work here.

I would like to wait for your implementation and do one more read of the spec, before merging.

pjcozzi:
@lasalvavida do you need to make any updates here to account for any potential changes in https://github.com/AnalyticalGraphicsInc/cesium/pull/4101?

lasalvavida:
I don't think so. We do need the feature table spec. Should that be part of this or a separate pull request?

pjcozzi:
> Should that be part of this or a separate pull request?

At this point, we don't have the generic glTF `bufferView`/`accessor`-like schema for the feature table, right?  If that is the case, it does not need to be part of this PR.

pjcozzi:
Is this ready to merge?

lilleyse:
Should the pptx and svg files be in this PR?

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+in the Cesium implementation of 3D Tiles.
 
-## Batch Table
+## Feature Table
 
-_TODO: create a separate Batch Table spec that b3dm, i3dm, etc. can reference, [#32](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/32)?_
+Contains values for `i3dm` semantics used to create instanced models.
 
-The batch table is a `UTF-8` string containing JSON.  It immediately follows the header.  It can be extracted from the arraybuffer using the `TextDecoder` JavaScript API and transformed to a JavaScript object with `JSON.parse`.
+See the [Feature Table](TODO:add link) reference for more information.
```
lilleyse: For now should we just remove this line?

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
+Scaling can be applied to instances using the `SCALE` and `SCALE_NON_UNIFORM` semantics.
+`SCALE` applies a uniform scale along all axes, and `SCALE_NON_UNIFORM` applies scaling to the `x`, `y`, and `z` axes independently.
+
+If `SCALE` and `SCALE_NON_UNIFORM` are defined for an instance, both scaling operations will be applied.
+
+## Batch Table
+
+Contains metadata organized by `batchId` that can be used for declarative styling.
+
+See the [Batch Table](TODO:add link) reference for more information.
```
lilleyse: Should this link just be https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel#batch-table for now?

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,61 +22,131 @@ A tile is composed of a header section immediately followed by a body section.
...
 
 ## glTF
```
lilleyse: The line: 

```The glTF field immediately follows the batch table (or immediately follows the header, if `header.batchTableByteLength` is zero).``` 

is no longer accurate now that there's a feature table.

**Comment by pjcozzi** [comment]
>Should the pptx and svg files be in this PR?

Yes

File: TileFormats/Instanced3DModel/README.md (hunk 34 above)
Comment (pjcozzi) [inline TileFormats/Instanced3DModel/README.md]: OK

File: TileFormats/Instanced3DModel/README.md (hunk 35 above)
Comment (pjcozzi) [inline TileFormats/Instanced3DModel/README.md]: +1 for now.

**Comment by lasalvavida** [comment]
@pjcozzi I think what @lilleyse is referring to is that I removed the .pptx in favor of separate .svg files which are a more proper way to store vector graphics.

**Comment by pjcozzi** [comment]
> @pjcozzi I think what @lilleyse is referring to is that I removed the .pptx in favor of separate .svg files which are a more proper way to store vector graphics.

Did you author the figure a tool other than PowerPoint?

**Comment by lasalvavida** [comment]
I used inkscape for the graphics I created. You could also use adobe illustrator to produce a similar result. I also exported the pptx->pdf->svg to get separate editable svgs for the layout graphics.

**Comment by pjcozzi** [comment]
At this point, I am not ready to commit to having everyone author these in Inkscape or Illustrator even if that is what the pros do.  I would rather keep these in pptx, where they are super easy for everyone to edit, and then at the final step, we'll have a designer to the "print ready" versions.

**Comment by lasalvavida** [comment]
I can make the changes to the original powerpoint for the header graphics and replace the svgs with that. But for the others, I would be putting svgs into powerpoint slides which seems really backwards.

**Comment by pjcozzi** [comment]
Ah, I see.  Just leave the others since I don't expect them to change, but the headers need to be malleable for now.

**Comment by lasalvavida** [comment]
Okay, this is ready for another look.

**Comment by lilleyse** [comment]
All looks good to me. @pjcozzi do you want to merge?

File: TileFormats/Instanced3DModel/README.md [hunk 37]
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
 
-_TODO: Link to Cesium code for reading header_
+If either `featureTableJSONByteLength` or `gltfByteLength` equal zero, the tile does not need to be rendered.
 
-If either `gltfByteLength` or `instancesLength` equal zero, the tile does not need to be rendered.
+The body section immediately follows the header section, and is composed of three fields: `Feature Table`, `Batch Table`, and `glTF`.
 
-The body section immediately follows the header section, and is composed of three fields: `Batch Table`, `glTF`, and `instances`.
+Code for reading the header can be found in
+[Instanced3DModelTileContent](https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Instanced3DModel3DTileContent.js#L170)
```
pjcozzi: Change this to:

> [Source/Scene/Instanced3DModelTileContent.js](https://github.com/AnalyticalGraphicsInc/cesium/blob/3d-tiles/Source/Scene/Instanced3DModel3DTileContent.js)

If you can give me commit access to your fork, I will make minor changes like this.

pjcozzi: Also note that I dropped the link to the exact line number.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+
+Contains values for `i3dm` semantics used to create instanced models.
+
+### Semantics
+
+#### Instance Semantics
+
+These semantics map to an array of feature values that are used to create instances. The length of these arrays must be the same for all semantics and is equal to the number of instances.
+
+If a semantic has a dependency on another semantic, that semantic must be defined in order to be a valid tile.
```
pjcozzi: > that semantic must be defined in order to be a valid tile

to

> hat semantic must be defined.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+| `NORMAL_UP` | `float32[3]`| A unit vector defining the `up` direction for the orientation of the instance. | :red_circle: No <br> :large_blue_diamond: Depends on `NORMAL_RIGHT` |
+| `NORMAL_RIGHT` | `float32[3]` | A unit vector defining the `right` direction for the orientation of the instance. Must be orthogonal to `up`. | :red_circle: No <br> :large_blue_diamond: Depends on `NORMAL_UP` |
+| `NORMAL_UP_OCT32P` | `uint16[2]` | An oct-encoded unit vector with 32-bits of precision defining the `up` direction for the orientation of the instance. | :red_circle: No <br> :large_blue_diamond: Depends on `NORMAL_UP_OCT32P` |
+| `NORMAL_RIGHT_OCT32P` | `uint16[2]` | An oct-encoded unit vector with 32-bits of precision defining the `right` direction for the orientation of the instance. Must be orthogonal to `up`. | :red_circle: No <br> :large_blue_diamond: Depends on `NORMAL_RIGHT_OCT32P` |
+| `SCALE` | `float32` | A number defining a scale to apply to all axes of the instance. | :red_circle: No |
```
pjcozzi: Somewhere, we should say that both `SCALE` and `SCALE_NON_UNIFORM` can be defined, and both scales will be applied.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+| `BATCH_ID` | `unit16` | The `batchId` of the instance that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No |
+
+#### Global Semantics
+
+These semantics define global properties for all instances.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `INSTANCES_LENGTH`| `uint32` | The number of instances to generate. The length of each array value for an instance semantic should be equal to this. | :white_check_mark: Yes |
+| `QUANTIZED_VOLUME_OFFSET` | `float32[3]` | A 3-component array of numbers defining the offset for the quantized volume. | :red_circle: No |
```
pjcozzi: This and below are required if `POSITION_QUANTIZED` is defined.  Overall, I'm not sure if the yes/depends/no column is clear.

Maybe it should be just:
* Yes
* Yes, if ...
* No

And drop the "depends on" since that becomes explicit in "Yes, if ..."

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+| `QUANTIZED_VOLUME_SCALE` | `float32[3]` | A 3-component array of numbers defining the scale for the quantized volume. |:red_circle: No |
+
+### Instance Orientation
+
+An instance's orientation is defined by an orthonormal basis created by an `up` and `right` vector. If `NORMAL_UP` and `NORMAL_RIGHT` or `NORMAL_UP_OCT32P` and `NORMAL_RIGHT_OCT32P` are not present,
+the instance will default to the `east/north/up` reference frame's orientation for the instance's Cartographic position (`x`, `y`, `z` converted to `longitude` and `latitude` on the `WGS84` ellipsoid).
```
pjcozzi: This should say the `(x, y, z)` with the transform from #98 applied.  I know the spec is updated for #98 yet, but I think it will be easier to make the update here now.

@lilleyse do we have a unit test for this?

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+![box rotated basis](figures/box-rotated-basis.png)
+
+#### Oct-encoded Normal Vectors
+
+If `NORMAL_UP` and `NORMAL_RIGHT` are not defined for an instance, its orientation may be stored as oct-encoded normals in `NORMAL_UP_OCT32P` and `NORMAL_RIGHT_OCT32P`.
+
+These define `up` and `right` using the oct-encoding described in
+[*A Survey of Efficient Representations of Independent Unit Vectors* by Cigolle et al.](http://jcgt.org/published/0003/02/01/).
+
+An implementation for encoding and decoding these unit vectors can be found in Cesium's
```
pjcozzi: Put in the same paragraph as the previous sentence.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+[AttributeCompression](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js)
```
pjcozzi: This too.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+module.
+
+### Instance Position
+
+`POSITION` defines the location where an instance should be placed in Cartesian space.
```
pjcozzi: What space?  Be precise here with respect to #98.  @lilleyse please help provide language if needed.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
 
-_TODO: create a separate Batch Table spec that b3dm, i3dm, etc. can reference, [#32](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/32)?_
+### Instance Scaling
 
-The batch table is a `UTF-8` string containing JSON.  It immediately follows the header.  It can be extracted from the arraybuffer using the `TextDecoder` JavaScript API and transformed to a JavaScript object with `JSON.parse`.
+Scaling can be applied to instances using the `SCALE` and `SCALE_NON_UNIFORM` semantics.
+`SCALE` applies a uniform scale along all axes, and `SCALE_NON_UNIFORM` applies scaling to the `x`, `y`, and `z` axes independently.
 
-Each property in the object is an array with its length equal to `header.batchLength`.  Array elements can be any valid JSON data type, including objects and arrays.  Elements may be `null`.
+If `SCALE` and `SCALE_NON_UNIFORM` are defined for an instance, both scaling operations will be applied.
```
pjcozzi: I think we should move this to the table above since that is where the question is most likely to be asked.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
 
-An instance's `batchId` is used to access elements in each array and extract the corresponding properties.  For example, the following batch table has properties for two instances:
 ```json
 {
-    "id" : ["unique id", "another unique id"],
-    "displayName" : ["Tree species", "Another tree species"],
-    "yearPlanted" : [1999, 2003],
-    "location" : [{"x" : 1, "y" : 2}, {"x" : 3, "y" : 4}]
+    "INSTANCES_LENGTH" : 4,
+    "POSITION" : [
```
pjcozzi: Does the feature table JSON support pointing to the feature table binary yet?  I'm not asking about the general purpose glTF-like `bufferView`/`accessor`.  It would be fine if it is just for well known semantics so they are not stored as bulky arrays in JSON.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,180 @@ A tile is composed of a header section immediately followed by a body section.
...
+        65535, 32768
+    ]
+}
 ```
 
+## Batch Table
+
+Contains metadata organized by `batchId` that can be used for declarative styling.
+
+See the [Batch Table](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel#batch-table) reference for more information.
```
pjcozzi: Use relative links for links inside the repo.

Also add a `TODO` to change this link to the generic batch table.

pjcozzi:
This is super close, thanks @lasalvavida.

pjcozzi:
Also explain how the transform (inverse transpose) affects normals.

TileFormats/Instanced3DModel/README.md
lasalvavida: This is stated in the `Instance Scaling` section.

TileFormats/Instanced3DModel/README.md
lilleyse: There aren't any unit tests for the transform branch yet, but I will make sure to test this.

TileFormats/Instanced3DModel/README.md
lasalvavida: I think if you do that, you lose the idea that something is not required, but if you do use it, you need something else as well.

Maybe, a no, unless would be more clear in this case.

TileFormats/Instanced3DModel/README.md
lasalvavida: Yes, it does. However, I think that should be part of the feature table specification.

lasalvavida: I can add a note here that they should be stored in the feature table binary and are just written out to make them more readable.

TileFormats/Instanced3DModel/README.md
pjcozzi: Give that a shot and see if it is less confusing.  Currently, it looks more complicated than it really is.

TileFormats/Instanced3DModel/README.md
pjcozzi: See this comment: https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/116#discussion_r75418183

lasalvavida: I think that may be a little too verbose. Let me know what you think with the latest revisions here and if you still want me to make a separate binary on for each example I will.

lasalvavida:
> Also explain how the transform (inverse transpose) affects normals.

Wouldn't that be better suited for the tileset transform specification?

Also this is updated and ready for another look.

pjcozzi:
>> Also explain how the transform (inverse transpose) affects normals.
> Wouldn't that be better suited for the tileset transform specification?

It should be mentioned here too; otherwise, someone writing a new tile loaded based on this spec may forget it.  It can be short, just something like:

> The positions and normals are defined in the tile's local coordinate system; they should be transformed as described in `TODO`.

Later, that will link to the section in the spec that described how to compute the matrices using the tile and its ancestors.

pjcozzi:
Is this ready?

lasalvavida:
Yup. Everything should be addressed now.

TileFormats/Instanced3DModel/README.md
```diff
@@ -22,65 +22,193 @@ A tile is composed of a header section immediately followed by a body section.
...
+`POSITION = POSITION_QUANTIZED * QUANTIZED_VOLUME_SCALE + QUANTIZED_VOLUME_OFFSET`
+
+### Instance Scaling
+
+Scaling can be applied to instances using the `SCALE` and `SCALE_NON_UNIFORM` semantics.
+`SCALE` applies a uniform scale along all axes, and `SCALE_NON_UNIFORM` applies scaling to the `x`, `y`, and `z` axes independently.
+
+### Examples
+
+In these examples, the semantic values are shown as JSON arrays. This is done to make the examples more human readable, and is still a valid feature table.
```
pjcozzi: My comment from before still holds.  Please revise.

pjcozzi:
The `Required` column is still confusing.  I would drop the "depends" part.  What is the concern with:
* Yes (always)
* Sometimes, if... (use a different emoji, http://www.webpagefx.com/tools/emoji-cheat-sheet/)
* No (never)

Then under the table, include a link to the example section so readers could jump to common layouts.

pjcozzi:
We need to define what happens if `NORMAL_UP` and `NORMAL_RIGHT` are defined, as well as `NORMAL_UP_OCT32P` and `NORMAL_RIGHT_OCT32P`.  Given other parts of 3D Tiles, I think it would be allowed, but we would use the uncompressed versions.

lasalvavida:
> The Required column is still confusing. I would drop the "depends" part. What is the concern with:
> * Yes (always)
> * Sometimes, if... (use a different emoji, http://www.webpagefx.com/tools/emoji-cheat-sheet/)
> * No (never)

My issue with this is that something being required but with an optional alternative, (`POSITION` and `POSITION_QUANTIZED`) is a different idea than something that is only required if something else is defined, (`NORMAL_RIGHT` and `NORMAL_UP` `QUANTIZED_VOLUME_OFFSET` and `POSITION_QUANTIZED`). 

I think applying a blanket "Sometimes, if" to both cases is confusing. If anything I think maybe:

* Yes
* Yes, unless
* No
* No, unless

lasalvavida:
Alright, updated, let me know what you think.

TileFormats/Instanced3DModel/README.md
pjcozzi: The new approach with the code example is good (actually really good for those doing writers in Node.js).  Update this paragraph since it is out-of-date.

pjcozzi:
> I think applying a blanket "Sometimes, if" to both cases is confusing. If anything I think maybe:

> Yes
Yes, unless
No
No, unless

This is pretty good, not sure that it is perfect, but let's go with it and see if we get any feedback.

pjcozzi:
Just that one small update and this is ready to merge!

lasalvavida:
Updated

pjcozzi:
Nice!

Issue #22, opened 2015-09-25
Finish Point Cloud Tile spec


mpgerlek:
The following changes to the points README would help reduce potential for misunderstanding:
- the `magic` bytes should be set to `pnts`
- the `version` should be set to `1`
- are the `positions` data to be in lat/lon, radians, or..?
- are the `positions` data to be expressed as `XYZXYZXYZ` or `XXXYYYZZZ`?
- are the `colors` data to be expressed as `RGBRGBRGB` or `RRRGGGBBB`?
- as a sanity check, note that the file size should be equal to `16 + numpoints * 15`

hobu:
Hobu, Inc. has been working on organization and streaming of large scale point clouds, and we have been working with the Potree project. See http://speck.ly and http://potree.entwine.io for examples. We have developed two open source software projects in support of this effort -- Greyhound http://github.com/hobu/greyhound and Entwine http://github.com/connormanning/entwine Greyhound is the HTTP server that answers LoD requests for a dynamic point cloud schema. Entwine is software that organizes massive point clouds (100+ billion pts) to be served through Greyhound. 

Entwine is like PotreeConverter if you are familiar with that software, except more sophisticated in that it losslessly processes the data and does so in a parallel way. PotreeConverter answers a fixed schema, whereas Entwine/Greyhound is dynamic (a basic assumption of XYZ of course). Greyhound answers requests as either uncompressed ArrayBuffer or LASzip-style compressed fields, which works in both native C/C++ and JavaScript code. We have found this LASzip-style approach to be efficient and data preserving from source all the way to client.

3D Tiles doesn't have to use the approaches we've developed, but we have been working through the problem for quite a while, and it may learn from our experience. Here's some highlights:

1. You only have to look at PCL's header files to see the consequences of fixed-field dimension types for point cloud data. There just ends up being way too many. Everyone wants their own special fields, and accommodating them all in a fixed way ends up spreading lots of pain around. We developed Greyhound to be able to answer dynamically (ask for the dimension list, ask for what you want in the format you want) to make it easier for clients. In my opinion, a static tile point cloud format must have the ability to dynamically define the content beyond the required XYZ. It doesn't have to be fancily specified, but I think it needs to be there.

2. Lossy data structures are a non-starter for many. There is a strong desire for a data organization that serves both the need of a servable LoD structure and one that does archival, at-rest storage.

[...]

PR #104, opened 2016-07-14
Update pnts tile format spec to include batch table

pjcozzi:
https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/pnts/TileFormats/Points

Add notes to keep it up-to-date with https://github.com/AnalyticalGraphicsInc/cesium/pull/4112

These are just notes; there will be more words when we flush out the spec.

Part of #22.


lilleyse:
The image has a colors[] at the end still.

lilleyse:
Maybe worth adding that a default color will be used if no color semantics are present.

pjcozzi:
> The image has a colors[] at the end still.

Good eye, updated.

pjcozzi:
> Maybe worth adding that a default color will be used if no color semantics are present.

Yes, updated.

This is ready.

lilleyse:
Did you want to update the picture considering https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/103? I'm going to make the change in Cesium.

pjcozzi:
Updated

lilleyse:
I merged this.. but is the "4 byte aligned array buffer" comment needed anymore?

pjcozzi:
It is implicit given the 28-byte header, but I still think it is good to be precise and it may be needed depending on the final format.

PR #116, opened 2016-08-18
Pnts update

lilleyse:
Will follow this PR with another that includes per-point batchID and batch table.


pjcozzi:
Initial work for #22.

pjcozzi:
Same comment about the `Required` column that I had for `i3dm`.

pjcozzi:
Same comment about strong binary data for the feature table that I had for `i3dm`.

pjcozzi:
Same comment about transforms, #98, as I had for `i3dm`.

pjcozzi:
Made a few small tweaks.  Check my commits.

pjcozzi:
@lasalvavida do you want to give this a quick read?

TileFormats/Points/README.md
```diff
@@ -2,35 +2,147 @@
...
+
+#### Point Normals
+
+Per-point normals are an optional property that can help improve the visual quality of points by enabling lighting, hidden surface removal, and other rendering techniques.  Oct-encoding is described in [*A Survey of Efficient Representations of Independent Unit Vectors* by Cigolle et al.](http://jcgt.org/published/0003/02/01/).  An implementation for encoding and decoding these unit vectors can be found in Cesium's
+[AttributeCompression](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Core/AttributeCompression.js)
+module.
+
+### Examples
+
+Note: these examples use JSON arrays for illustration purposes but for best performance per-point properties like `POSITIONS` and `RGB` should be stored in the `featureTableBinary`.
```
pjcozzi: Ah, I see.  We need to be careful with this because, at a glance, people would criticize this as a naive approach with just JSON.  Try separating each example below into `JSON` and a "binary" section.

pjcozzi:
Is this ready?

lasalvavida:
The same `no, unless` clause from the i3dm for `QUANTIZED_VOLUME_OFFSET` and `QUANTIZED_VOLUME_SCALE` should be applied here.

TileFormats/Points/README.md
lasalvavida: If we like it, maybe try what I put for i3dm. It makes this clarification and shows a json with the `byteOffset` attribute. I think its a little verbose to do a separate one for every example.

lilleyse:
> Is this ready?

I'm waiting for https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/100 to be done. Then it should be ready soon after.

lilleyse:
Updated.

pjcozzi:
For the `Quantized Positions` section, can you include the figures (and any additional wording if any) from the `i3dm` spec?

pjcozzi:
Can we add an example with oct-encoded normals?

pjcozzi:
Made some tweaks in 27bad0a777fe4c4f3fcb59eec46b3a267d52b34c.

pjcozzi:
This is very close, thanks!

TileFormats/Instanced3DModel/README.md
```diff
@@ -169,7 +169,7 @@ var featureTableBinary = new Buffer(new Float32Array([
 #### Quantized Positions and Oct-Encoded Normals
 
 In this example, the 4 instances will be placed with an orientation `up` of `[0.0, 1.0, 0.0]` and `right` of `[1.0, 0.0, 0.0]` in oct-encoded format 
-and they will be placed on the corners of a quantized volume that spans from `-250.0` to `250.0` units in the `x` and `z` directions.
+and they will be placed on the corners of a quantized volume that spans from `-500.0` to `0.0` units in the `x` and `z` directions.
```
lilleyse: I wasn't sure if these numbers were supposed to also take the offset into account, so I adjusted them.

lilleyse:
The folder name is still called `Points`. Should we change this? It may cause some broken links.

pjcozzi:
> The folder name is still called Points. Should we change this? It may cause some broken links.

I noticed that too.  Please rename and ensure that all links inside this repo are updated.  I don't think there are many external deep links.

TileFormats/Instanced3DModel/README.md
pjcozzi: @lasalvavida can you confirm this?

pjcozzi:
Looks good, please merge after renaming.

TileFormats/Instanced3DModel/README.md
lasalvavida: No, this change is not correct. The offset puts the origin at `(-250, 0, -250)`, then the scale makes the opposite corner `-250 + 500 = 250` : `(250, 0, 250)`.

TileFormats/Instanced3DModel/README.md
```diff
@@ -187,21 +187,21 @@ var featureTableJSON = {
     }
 };
 
-var positionQuantizedBinary = new Buffer(new UInt16Array([
+var positionQuantizedBinary = new Buffer(new Uint16Array([
```
lasalvavida: Thank you for catching this!

TileFormats/Instanced3DModel/README.md
lilleyse: Ah alright, I mistook the offset as the center. I'll need to update https://github.com/AnalyticalGraphicsInc/cesium/pull/4183 as well.

lilleyse:
Updated.

PR #118, opened 2016-08-22
Feature Table Specification

lasalvavida:
Feature Table specification document.

* [x] Add reference to this in Pnts when #116 gets merged.


TileFormats/FeatureTable/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+var positionArray = new Float32Array(featureTableBinary.buffer, byteOffset, featuresLength * 3); // There are three components for each POSITION feature.
+var position = positionArray.subarray(featureId * 3, featureId * 3 + 3); // Using subarray creates a view into the array data, and not a new array, which is better for performance.
+```
+
+## Implementation Notes
+
+This may vary between implementations, but in javascript, a `TypedArray` cannot be created on data unless it is byte-aligned.
+This means that a `Float32Array` must be stored in memory such that its data begins on a byte multiple of four since each `float` contains four bytes.
+The data types used in 3D Tiles have a maximum length of four bytes, so padding to a multiple of four will work for all cases, since smaller types with lengths of one and two will also be byte-aligned.
```
lilleyse: We support doubles too.

lilleyse: At least, maybe not yet but potentially.

lasalvavida: Okay, I'll make this less specific.

pjcozzi:
@lasalvavida @lilleyse when you open spec PRs, please include a link like https://github.com/lasalvavida/3d-tiles/tree/feature-table/TileFormats/FeatureTable

TileFormats/FeatureTable/README.md
```diff
@@ -0,0 +1,64 @@
+# Feature Table
+
+## Contributors
+
+* Sean Lilley, [@lilleyse](https://twitter.com/lilleyse)
```
pjcozzi: Add
```
Dan Bagnell, [@bagnell](https://github.com/bagnell)
```
The order can be Sean, you, Dan, and me.

pjcozzi: I'll make this change along with some other edits.

lilleyse:
Check out my note here which may be relevant to this PR: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/22#issuecomment-241471497

pjcozzi:
Made some tweaks in 877ffb946a025a51355c9fbffc2347a9e207c77d

pjcozzi:
> feature-table-layout.png

* Change `featureTableJSON` to `JSON header`
* Change `featureTableBinary` to `binary body`

> feature-table-binary-index.png

* Change `featureTableJSON` to `JSON header`
* Change `featureTableBinary` to `binary body`

In the context of this spec, the `featureTable` prefix is not needed and not consistent with the spec's text.

Keep `featureTableBinaryByteLength` since I imagine that is from the tile's header (explicitly mention that in the text).

pjcozzi:
> JSON Header

Include an complete-ish example like the tile format specs have.

pjcozzi:
Create a JSON schema for the JSON header.  It should be pretty small, but we need to be precise, e.g., right now `byteOffset` isn't really defined well.  What data type is it?  Can it be negative?  Is it required?  To ensure consistent implementations, we need to be precise.

For JSON schema examples, see
* https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/schema
* https://github.com/KhronosGroup/glTF/tree/master/specification/schema

Use JSON Schema 3.  You will not be able to represent all requirements/relationships, but others can be put in the description and the spec's text.

Make the schema a subdirectory of this new directory and include a link to it from the spec.

pjcozzi:
>  Add reference to this in Pnts when #116 gets merged.

Merge in master and you can do this.

TileFormats/FeatureTable/README.md
```diff
@@ -0,0 +1,71 @@
...
+```javascript
+var byteOffset = featureTableJSON.POSTION.byteOffset;
+
+var positionArray = new Float32Array(featureTableBinary.buffer, byteOffset, featuresLength * 3); // There are three components for each POSITION feature.
+var position = positionArray.subarray(featureId * 3, featureId * 3 + 3); // Using subarray creates a view into the array, and not a new array.
+```
+
+## Implementation Notes
+
+In JavaScript, a `TypedArray` cannot be created on data unless it is byte-aligned to the data type.
```
pjcozzi: Do we have to pad the JSON string as opposed to the start of the binary body?

I would like this spec to include language similar to this from b3dm:

> The body section immediately follows the header section, and is composed of two fields: Batch Table and Binary glTF.

I guess in either padding case, this is still true.

lasalvavida:
@pjcozzi Is there a particular reason we are tied to v3 of the json-schema spec? I haven't found a way to define that `POSITION` is required unless `POSITION_QUANTIZED` is present and vice-versa in v3 of the spec, but it's pretty easy to do in v4 with the `oneOf` or `not` keywords. https://github.com/json-schema/json-schema/wiki/anyOf,-allOf,-oneOf,-not

TileFormats/FeatureTable/README.md
lilleyse: Padding at the start of the binary body is not easily possible. When you add padding to the binary the `byteOffset` values in the JSON need to change, which then may extend the JSON length and breaks the alignment of the binary.

I believe this is the same reason why binary gltf appends whitespace to the JSON rather than padding the binary.

pjcozzi:
> @pjcozzi Is there a particular reason we are tied to v3 of the json-schema spec?

Tooling.  There are more tools for v3 than v4, e.g., like this tool we wrote: https://github.com/AnalyticalGraphicsInc/wetzel

With that said, the v4 features look handy and we actually don't have that much JSON schema for 3D Tiles (we are definitely not updating glTF to v4 right now).  If you want to update _all_ of the 3D Tiles JSON schema to v4 right now, go for it, but please stay focused on the end goal of 3D Tiles and don't get too caught up in schema land.  I have been there.

lasalvavida:
Sounds good to me. Changes should be pretty minimal for the schema we already have.

lasalvavida:
This should be ready for a look.

pjcozzi:
Made some tweaks in d161ccb61178dc8bb37b9849af37af92da3eda06.

TileFormats/FeatureTable/schema/featureTable.schema.json
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema" : "http://json-schema.org/draft-04/schema#",
+    "id" : "featureTable.schema.json",
+    "title" : "Feature Table",
+    "type" : "object",
+    "description" : "A set of semantics containing per-tile and per-feature values defining the position and appearance properties for features in a tile.",
+    "patternProperties" : {
```
pjcozzi: If a property is an object, does it have to have a `byteOffset` property?

For example, if a tile format wants to define a Feature Table with a global semantic like
```
"KEY" : {
   "some-property" : 0.0,
   "another-property" : [1, 2, 3]
}
```
Can it do so as the schema is written here?  (It might, I just don't know v4).

I suggest that we let the schema be flexible (and note this in the Feature Table spec) and rely on each tile format to specifically scope what is valid.  This will be forward-compatible as we and others add new tile formats.

TileFormats/FeatureTable/schema/i3dm.featureTable.schema.json
```diff
@@ -0,0 +1,62 @@
...
+  "id" : "i3dm.featureTable.schema.json",
+  "title" : "Instanced 3D Model Feature Table",
+  "type" : "object",
+  "description": "A set of Instanced 3D Model semantics that containper-feature values defining the position and appearance properties for instanced models in a tile.",
+  "allOf" : [
+    {
+      "$ref" : "featureTable.schema.json"
+    },
+    {
+      "properties": {
```
pjcozzi: Each of these need to define the value values, e.g., is it a scalar or an array, min/max, etc.

TileFormats/FeatureTable/schema/featureTable.schema.json
lasalvavida: Your example currently will fail since it doesn't have a `byteOffset`. Do you want to just make `byteOffset` optional?

TileFormats/FeatureTable/schema/pnts.featureTable.schema.json
```diff
@@ -0,0 +1,49 @@
...
+  "id" : "pnts.featureTable.schema.json",
+  "title" : "Point Cloud Feature Table",
+  "type" : "object",
+  "description": "A set of Point Cloud semantics that contains values defining the position and appearance properties for points in a tile.",
+  "allOf" : [
+    {
+      "$ref" : "featureTable.schema.json"
+    },
+    {
+      "properties": {
```
pjcozzi: Same comment.

pjcozzi:
Made a few other trivial tweaks.

Really nice progress here.

TileFormats/FeatureTable/schema/featureTable.schema.json
lasalvavida: To elaborate, a semantic value is currently defined as either an object with a `byteOffse`t referencing the binary, an array, or a single value, and I think we currently forbid using SEMANTICS not defined for a tile format, so if it is an object, I think you should always have a `byteOffset`.

pjcozzi: I read that in the spec, but this schema doesn't, for example, explicitly disallow strings.

I would make the schema for the generic Feature Table as generic as possible.  Basically all we can say is that if a key is an object, and that object has a `byteOffset` property, then `byteOffset` should be:
```
"type" : "integer",
"minimum" : 0
```
Then let each tile format specifically scope what is valid for its Feature Table.  This is no more work for an implementation with our current tile formats, and allows future tile formats maximum flexibility to define their data structures.

lasalvavida:
Updated

lasalvavida:
If you want to play around with schemas:

`npm install -g ajv-cli`

`ajv -s schema/i3dm.featureTable.schema.json -r schema/featureTable.schema.json -d examples/i3dm.featureTable.json`

And edit the example to see that it breaks when it is supposed to.

pjcozzi:
Can you put some instructions like https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/118#issuecomment-242085086 in a README.md in the /schema directory?

lasalvavida:
Sure. Right now I have the feature table schemas in `TileFormats/FeatureTable/schema`, should I move those out to the global schema folder?

lasalvavida:
Added README

pjcozzi:
Made some updates in e362488ad22048327d4bb29cc94c6d9358ef0e94.

TileFormats/FeatureTable/schema/featureTable.schema.json
```diff
@@ -0,0 +1,91 @@
...
+    },
+    "definitions" : {
+        "featureReference" : {
+            "type" : "object",
+            "required" : ["byteOffset"]
+        },
+        "featureArray" : {
+            "type" : "array",
+            "items" : {
+                "type" : "number"
```
pjcozzi: For the base Feature Table spec, we can't require that arrays only contain numbers that would prevent new tile formats from using other datatypes.

TileFormats/FeatureTable/schema/featureTable.schema.json
```diff
@@ -0,0 +1,91 @@
...
+        "feature" : {
+            "oneOf" : [
+                {
+                    "$ref" : "#/definitions/featureReference"
+                },{
+                    "$ref" : "#/definitions/featureArray"
+                }
+            ]
+        },
+        "featureGlobalScalar" : {
```
pjcozzi: This should have a more precise name like `featureGlobalLength`.

TileFormats/FeatureTable/schema/featureTable.schema.json
lasalvavida: That isn't what this does. It is a definition that I use for more specific semantics in the i3dm and pnts specifications. This only puts a restriction on the data type where it is used.

TileFormats/FeatureTable/schema/featureTable.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "type" : "object",
+            "required" : ["byteOffset"]
+        },
+        "featureArray" : {
+            "type" : "array",
+            "items" : {
+                "type" : "number"
+            }
+        },
+        "feature" : {
```
pjcozzi: perhaps `perFeatureProperty` and then `globalPropertyXXX` below.

TileFormats/FeatureTable/schema/featureTable.schema.json
```diff
@@ -0,0 +1,91 @@
...
+            "properties" : {
+                "byteOffset" : {
+                    "type" : "integer",
+                    "minimum" : 0
+                }
+            }
+        }
+    },
+    "definitions" : {
+        "featureReference" : {
```
pjcozzi: Perhaps `binaryBodyReference`?

TileFormats/FeatureTable/schema/featureTable.schema.json
pjcozzi: Actually, what exactly is this?

TileFormats/FeatureTable/schema/featureTable.schema.json
lasalvavida: I can make it `featureNumericArray` if you'd like, but in general if you wanted to make a schema for a property that had something besides numbers in an array, you would need to make a new definition for that anyway.

TileFormats/FeatureTable/schema/i3dm.featureTable.schema.json
```diff
@@ -0,0 +1,71 @@
+{
+    "$schema" : "http://json-schema.org/draft-04/schema#",
+    "id" : "i3dm.featureTable.schema.json",
+    "title" : "Instanced 3D Model Feature Table",
+    "type" : "object",
+    "description" : "A set of Instanced 3D Model semantics that contains values defining the position and appearance properties for instanced models in a tile.",
+    "allOf" : [
+        {
+            "$ref" : "featureTable.schema.json"
+        },
```
pjcozzi: Minor whitespace here and throughout (but will save a level of indent), change lines like this to:
```
}, {
```

And above and throughout, can be like
```
[{
```

which are both what we do in JavaScript.

TileFormats/FeatureTable/schema/featureTable.schema.json
lasalvavida: It is a general definition for scalar values. It gets applied to `INSTANCES_LENGTH` and `POINTS_LENGTH`. It says that a single scalar can be stored via reference in the binary, can be an array with only one element, or can be an integer >= 0.

TileFormats/FeatureTable/schema/featureTable.schema.json
pjcozzi: Gotcha, let's rename this to something more precise like `numericArray` or similar.

pjcozzi:
Thanks for the JSON schema validation instructions, I made some edits for conciseness in 06b1305d1a9da55e6d20518011ba4140d862b28a and also had to mention that it is not full 3D Tiles validation, but I am digging the v4 schema.

lasalvavida:
Edits on the schema naming and whitespace are done and are ready for a look.

pjcozzi:
> should I move those out to the global schema folder?

Yes that is probably so it is all in one place.  Please link to the schema file or directory from the README.md's.

TileFormats/FeatureTable/schema/featureTable.schema.json
pjcozzi: > I can make it featureNumericArray if you'd like, but in general if you wanted to make a schema for a property that had something besides numbers in an array, you would need to make a new definition for that anyway.

That is OK, we just don't want this name to be too generic because it sounds like it is for all arrays.

TileFormats/FeatureTable/schema/featureTable.schema.json
pjcozzi: We can't say that all scalars are integers >= 0, perhaps break this into two definitions.

lasalvavida:
Updated

pjcozzi:
One more question: `i3dm`, for example, defines the `POSITION` semantic, which I believe can be either an inline array or reference the binary body.  Are we sure we want to allow this much flexibility?  Why not have the semantic allow only referencing the binary body?  It helps make tilesets "efficient by default" and it is easier for a renderer to implement.  The only drawback is quick-and-dirty writers would require slightly more work, but we want to discourage creating inefficient tilesets so we don't get into "Cesium isn't slow; the data isn't optimized" situations.

lasalvavida:
If we're going to explicitly forbid it, should the Cesium implementation throw an exception if the values are in the JSON?

Or alternatively continue allowing it but throw a warning?

pjcozzi:
> If we're going to explicitly forbid it, should the Cesium implementation throw an exception if the values are in the JSON?

The validator, https://github.com/AnalyticalGraphicsInc/3d-tiles-tools/issues/5, would do full validation.  The runtime could just throw an exception if the semantic is not an object with a `byteOffset`.

Each semantic should also be explicit that it needs to be binary.

lasalvavida:
Updated

schema/boundingVolume.schema.json
```diff
@@ -1,6 +1,7 @@
 {
-    "$schema" : "http://json-schema.org/draft-03/schema",
-    "title" : "boundingVolume",
+    "$schema" : "http://json-schema.org/draft-04/schema",
```
pjcozzi: In JSON schema v4, is there a way to say that only one of the three properties below can be present?  I changed this in #123.

pjcozzi:
Very nice, @lasalvavida.

@lilleyse can you do a final read and merge?  That will also help you prep for the Batch Table spec.

lasalvavida:
> In JSON schema v4, is there a way to say that only one of the three properties below can be present? I changed this in #123.

The schema has now been updated to reflect this.

pjcozzi:
> The schema has now been updated to reflect this.

Nice!

lilleyse:
Everything looks good to me. This will be good to go after clarifying the byte alignment for the embedded glTF.

lasalvavida:
Made the edit, let me know what you think.

lilleyse:
Add a note that this is suggested to keep proper btye alignment for the embedded glTF for the b3dm and i3dm formats.

lasalvavida:
Updated

lilleyse:
Looks good to me, @pjcozzi do you want to do a quick read? Then I can merge.

TileFormats/FeatureTable/README.md
```diff
@@ -70,4 +70,4 @@ In JavaScript, a `TypedArray` cannot be created on data unless it is byte-aligne
 For example, a `Float32Array` must be stored in memory such that its data begins on a byte multiple of four since each `float` contains four bytes.
 
 The string generated from the JSON header should be padded with space characters in order to ensure that the binary body is byte-aligned.
-The binary body should also be padded if necessary when there is data following the Feature Table.
+The binary body should also be padded to a multiple of eight bytes. This ensures byte-alignment for the embedded glTF in Batched 3D Models and Instanced 3D Models.
```
pjcozzi: Are we sure about this?  I thought the guarantees would be:
* Typed arrays in the Feature Table's binary body are byte-aligned to the size of their component type, e.g., 4 bytes for 
`float`.
* Typed arrays in the Batch Table's binary body are byte-aligned to the size of their component type, e.g., 4 bytes for `float`.
* In `b3dm` and `i3dm` (put the wording in their specs), the Binary glTF is 8-byte aligned, so glTF's byte-alignment guarantees can be meet with respect to the tile's entire arraybuffer.

lasalvavida:
Updated. Is that better?

lasalvavida:
Actually I'm not sure about this now. For the i3dm it works fine since the header is 32 bytes. For batched it's only 20 right now. If you don't have a batch table to pad, how are you supposed to start it on a multiply of 8?

pjcozzi:
> Actually I'm not sure about this now. For the i3dm it works fine since the header is 32 bytes. For batched it's only 20 right now. If you don't have a batch table to pad, how are you supposed to start it on a multiple of 8?

Good catch.  I hate to say that the header (both `b3dm` and `i3dm`) need an explicit gltfOffset to allow "padding the header" (or really anything).  @lilleyse do you have a better suggestion?

Also, on the implementation side, I would like to merge all breaking b3dm changes in one PR since b3dm is so widely used.

lasalvavida:
We could make it implicit that if there's no batch table, the glTF starts at the next multiple of 8 (24)

lasalvavida:
Or alternatively pad the beginning and read until you hit the glTF magic.

lilleyse:
Well the simplest thing to do right now is nothing since the b3dm header is about to change to 24 bytes.

lasalvavida:
Alrighty, then this is probably good to merge, unless anyone has anything else?

pjcozzi:
Nice, big progress here!

PR #123, opened 2016-08-25
Per-tile transforms

### LINKED ISSUE #98: Per-tileset or per-tile transform

pjcozzi:
Fixes #98.

See
* [Coordinate System and Units](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/transform#coordinate-system-and-units)
* [Tile transform](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/transform#tile-transform)
* [External Tilesets](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/transform#external-tilesets)

The new figures will be updated later for consistency with the rest of the spec.


README.md
```diff
@@ -162,10 +164,110 @@ The file extension of `content.url` defines the [tile format](#tileFormats).  Th
...
+        var child = children[k];
+        var childToRoot = defined(child.transform) ? Matrix4.fromArray(child.transform) : Matrix4.IDENTITY;
+        childToRoot = Matrix4.multiplyTransformation(transformToRoot, childToRoot, childToRoot);
+        stack.push({
+            tile : child,
+            transformToRoot : childToRoot
+        });        
+    }
+}
+```
```
lilleyse: I think a recursive example may be easier to grasp here for most readers.

lilleyse:
Wow this covers a lot of small but important cases. Looks good to me!

README.md
pjcozzi: Ah, yes, it is much shorter.  If you have a Cesium unit test handy, could you quickly run this?

pjcozzi: It is here: f557f562ad2777bc3173e8905ea9ff09c10d4f55

README.md
```diff
@@ -162,10 +164,102 @@ The file extension of `content.url` defines the [tile format](#tileFormats).  Th
...
+    var children = tile.children;
+    var length = children.length;
+    for (var k = 0; k < length; ++k) {
+        var child = children[k];
+        var childToRoot = defined(child.transform) ? Matrix4.fromArray(child.transform) : Matrix4.IDENTITY;
+        childToRoot = Matrix4.multiplyTransformation(transformToRoot, childToRoot, childToRoot);
+        computeTransform(child, childToRoot);
+    }
+}
+```
```
lilleyse: Had to make two edits. Here's the working code.

```javacscript
        function computeTransforms(tileset) {
            var t = tileset.root;
            var transformToRoot = defined(t.transform) ? Matrix4.fromArray(t.transform) : Matrix4.IDENTITY;

            computeTransform(t, transformToRoot);
        }

        function computeTransform(tile, transformToRoot) {
            // Apply 4x4 transformToRoot to this tile's positions and bounding volumes

            var inverseTransform = Matrix4.inverse(transformToRoot, new Matrix4());
            var normalTransform = Matrix4.getRotation(inverseTransform, new Matrix3());
            normalTransform = Matrix3.transpose(normalTransform, normalTransform);
            // Apply 3x3 normalTransform to this tile's normals

            var children = tile.children;
            var length = children.length;
            for (var k = 0; k < length; ++k) {
                var child = children[k];
                var childToRoot = defined(child.transform) ? Matrix4.fromArray(child.transform) : Matrix4.clone(Matrix4.IDENTITY);
                childToRoot = Matrix4.multiplyTransformation(transformToRoot, childToRoot, childToRoot);
                computeTransform(child, childToRoot);
            }
        }
```

lilleyse: I'll just commit the change.

pjcozzi:
Thanks @lilleyse!  Is this ready?

Does the spec need to cover anything else based on your implementation experience?

lilleyse:
It covers everything I can think of.

Issue #24, opened 2015-10-02
Vector Tile: Polygons


pjcozzi:
See the originial notes for [Vector Data](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/VectorData/README.md).

Considering only polygons and no styling:

```
Header
------
magic            // uchar[4], "vctr"
version          // uint32,   1
byteLength       // uint32,   length, in bytes, of the entire tile.  See #23
batchTableLength // uint32,   length, in bytes, of the batch table.  Can be zero.

Batch Table
-----------
// Same as Batched 3D Model.  Optional.

Polygons
--------
polygonsLength    // uint32, number of polygons in the tile

There are polygonsLength of these (each is variable length):
* batchId         // uint32, index into batchTable, zero when Header.batchTableLength === 0
* minimumHeight   // float,  for shadow volume extrusion optimization
* maximumHeight   // float,  for shadow volume extrusion optimization
* positionsLength // uint32, number of vertex positions in this polygon
* x, y, z ...     // doubles in WGS84 coordinates
* indicesLength   // uint32, number of triangle indices
* i0, i1, i2, ... // uint16, every three indices form a triangle
```

This design assumes server-side triangulation and client-side subdivision and shadow volume extrusion.  It includes metadata to optimize shadow volume rendering (which makes a huge difference).

Polygons with holes can be represented as multiple polygons (with or without the same `batchId`) or as one polygon with disconnected triangles.

The geometry is well behaved: there are no degenerate triangles nor are the duplicate vertex positions.

Cesium should render these tiles with RTC instead of GPU RTE.

**Easy Questions**

* [x] `batchId` is the first field here, but the last field in i3dm (#23).  Which do we prefer?
* [x] Should indices be `uint16` or `uint32`?  When using `uint16`, a big polygon could be represented as several polygons with the same `batchId`.
* [x] `minimumHeight` and `maximumHeight` are for rendering optimization.  Most likely, they should be a semantic in the Batch Table instead.
* [x] `minimumHeight` and `maximumHeight` just need to be conservative.  Is `uint16` more appropriate?

**Less Easy Questions**

* [x] Tight representation for vertex positions, e.g., RTC, [EXT_quantized_attributes](https://github.com/mlimper/glTF/tree/spec-1.0/extensions/MultiVendor/EXT_quantized_attributes), delta encoding, Open3DGC SOA vs. AOS, cartographic, etc.
[...]

PR #124, opened 2016-08-26
Vector tiles

lasalvavida:
@bagnell @pjcozzi, this should be ready for a look.

This is largely based on the existing vector tiles implementation, but I have made some tweaks to make it more consistent with the  way that the `i3dm` and `pnts` specifications work.

**TODO**
- [ ] Write schema after this gets looked over a bit.


lasalvavida:
[link](https://github.com/lasalvavida/3d-tiles/tree/vector-tiles/TileFormats/VectorData)

pjcozzi:
I am part way through this.  Made some edits in 47d6089d2a9998318f22c07aba9f61cbd8b0f7af.

Please update the status section of the main README.md.

> POLYGON_COUNT

Perhaps better named `POLYGON_LENGTHS` or maybe `POLYGON_POINTS_LENGTHS`

> POLYLINE_COUNT

Similar comment.

> This refers to POLYGON_INDICES if it is defined, otherwise it refers to POSITION or POSITION_QUANTIZED.

I can see the use case for polylines, but are we sure about this for polygons?  CC @bagnell

bagnell:
>I can see the use case for polylines, but are we sure about this for polygons? CC @bagnell

We need to know the individual indices/positions of the polygons for picking and/or dynamically changing the polygons (we need to re-batch or reorder draw calls).

pjcozzi:
Ah, OK

pjcozzi:
> TODO: Write schema after this gets looked over a bit.

I added this to a tasklist at the top of this PR.  Please add any other items that need to be addressed before this is merged, e.g., review by @bagnell, etc.

TileFormats/VectorData/README.md
```diff
@@ -1,20 +1,207 @@
...
+
+TODO: Write schema after this gets looked over a bit.
+
+### Semantics
+
+If a semantic has a dependency on another semantic, that semantic must be defined as well.
+If both `POSITION` and `POSITION_QUANTIZED` are defined, the higher precision `POSITION` will be used.
+Per-feature semantics specific to a feature type are prefixed with the name of the feature type. e.g. `POLYGON` for polygons and `POLYLINE` for polylines.
+
+At least one global `LENGTH` semantic must be defined.
```
pjcozzi: Here and below in the `Global Semantics` table, I don't think we need to require this.  Throughout 3D Tiles if length is zero (e.g., no feature table), the tile doesn't need to be rendered.  Requiring that one of these be defined still does not guarantee that one is non-zero so it is not really a meaningful requirement.

pjcozzi:
>>> This refers to POLYGON_INDICES if it is defined, otherwise it refers to POSITION or POSITION_QUANTIZED.

>> I can see the use case for polylines, but are we sure about this for polygons? CC @bagnell

> We need to know the individual indices/positions of the polygons for picking and/or dynamically changing the polygons (we need to re-batch or reorder draw calls).

Actually, my question is if this is for supporting triangle soup, and if so why can't we just require indices for polygons?

pjcozzi:
We'll also need to change the pnts and i3dm spec, but the `Data Type` column is confusing.  For example, `uint32` means "array of uint32" in the `Vector Semantics` table, but it means "unit32" in the `Global Semantics`.

Perhaps

* Rename this to "Component Type" (we use this term elsewhere in 3D Tiles, right?  It is from WebGL/glTF) in the `Vector Semantics` table with a sentence beforehand to say that these are arrays, even if it is redundant with the Feature Table spec.
* Rename "Vector Semantics" to "Per-Feature Semantics"

pjcozzi:
Also consider renaming `POLYGON` and `POLYLINE` prefixes for per-feature semantics to `POLYGONS` and `POLYLINES`.

For example, `POLYLINE_BATCH_ID` sounds like one batch id, and `POLYLINE_BATCH_IDS` sounds like several batch ids for one polyline.  `POLYLINES_BATCH_ID` sounds like what we intend.

pjcozzi:
> CLAMP_TO_GROUND

As discussed offline, I would drop this for now until we have the full game plan for extrusions, etc.

TileFormats/VectorData/README.md
```diff
@@ -1,20 +1,207 @@
...
+With this flag enabled, the Cartesian `x`, `y`, and `z` coordinates should be mapped to the longitude and latitude of the surface.
+
+The `MINIMUM_HEIGHT` and `MAXIMUM_HEIGHT` semantics place bounds on the height from the surface for the projection. 
+For example, `MINIMUM_HEIGHT` could be set to `10.0` to have the rendered vector graphics float uniformly `10.0` units off of a flat surface.
+ 
+This behavior is disabled by default, but is useful for rendering things like roads and borders onto a globe.
+
+### Examples
+
+This example will define positions for polylines, but the same procedure is applied for polygons as well.
```
pjcozzi: I don't expect this to be true.

Polygons should always have indices.

Polylines should optionally have indices.

pjcozzi: > Polylines should optionally have indices.

Or not have indices and just be strips.

TileFormats/VectorData/README.md
```diff
@@ -1,20 +1,207 @@
...
+
+**Figure 2**: Example polyline plot.
+
+![vector polyline plot](figures/vector-polyline-plot.png)
+
+A data layout using indices is shown in Figure 3. This is typically the best approach since duplicate points do not need to have their data repeated.
+
+**Figure 3**: Example polyline data layout.
+
+![vector polyline diagram](figures/vector-polyline-diagram.png)
```
pjcozzi: The boxes for the `POSITION` array are not aligned.

TileFormats/VectorData/README.md
```diff
@@ -1,20 +1,207 @@
...
+For example, `MINIMUM_HEIGHT` could be set to `10.0` to have the rendered vector graphics float uniformly `10.0` units off of a flat surface.
+ 
+This behavior is disabled by default, but is useful for rendering things like roads and borders onto a globe.
+
+### Examples
+
+This example will define positions for polylines, but the same procedure is applied for polygons as well.
+The only difference is that polygons automatically connect the last point to the start point to create a closed shape, whereas with polylines, that relationship needs to be defined explicitly to achieve the same result.
+
+In this example, four features are defined, a line, two triangles connected to each line endpoint, and a single point on the XY-Plane as shown in Figure 2.
```
pjcozzi: Let's discuss using a new example.

pjcozzi:
Made a few tweaks in 09f1222b174038cf2c640c0e33b9e160e9b7bd64, and also removed `CLAMP_TO_GROUND`, which had some misinformation.

pjcozzi:
Can we make this explicit from #24?

* The geometry is well behaved: there are no degenerate triangles nor are the duplicate vertex positions.

We should also be explicit that polygons do not have self-intersections (in WGS84 coordinates).

pjcozzi:
Can we allow per-feature indices arrays to either be `uint16` or `uint32`?

pjcozzi:
> Can we allow per-feature indices arrays to either be uint16 or uint32?

Same comment for batch ids throughout the spec.

lasalvavida:
> Actually, my question is if this is for supporting triangle soup, and if so why can't we just require indices for polygons?

During my discussion with @bagnell, I did bring this up, but he said it was reasonable to support both. If that changes, I'm happy to adjust the spec.

TileFormats/VectorData/README.md
lasalvavida: What did you have in mind?

pjcozzi:
> During my discussion with @bagnell, I did bring this up, but he said it was reasonable to support both. If that changes, I'm happy to adjust the spec.

Given the "fast by default" spirit of 3D Tiles, I see no reason to encourage triangle soup.

pjcozzi:
We also need to define if polygons and polylines are sub-sampled as geodesics or rhumb lines.  This can be a global semantic as I think this needs to be an optional instead of just, for example, always going with rhumb lines.

pjcozzi:
@bagnell after discussing offline with @lasalvavida, can we explore a more concise representation for positions for ground clamped positions?  I think we can do even better than quantized x, y, z by taking advantage of the fact that the positions are on the ground.

Longitude/latitude is the obvious choice, but would require a lot of precision and runtime conversion.

Perhaps something in the tile's tangent plane or angular measurements from the tile's center.  The key is to not require many bits, but still have good precise by using a local origin, and to have a fast conversion to WGS84 for rendering.

bagnell:
>Given the "fast by default" spirit of 3D Tiles, I see no reason to encourage triangle soup.

I agree. 

@lasalvavida We discussed this offline, but what are `POLYGON_COUNT` and `POLYGON_INDICES`. Assuming that indices are now required, they are the number of positions and number of indices?

bagnell:
>angular measurements from the tile's center

Isn't that the same as latitude/longitude or do you mean polar coordinates in the tangent plane?

>Perhaps something in the tile's tangent plane

I like this idea, but I would be concerned with points on the plane with large distances to the ellipsoid. We would also run into the same problems that we have with triangulating polygons on the tangent plane in Cesium. Can we restrict the tile regions so they can never get that big?

Another idea is to store them in 2D (`GeographicProjection` or `WebMercatorProjection`). This would be expensive to convert back to WGS84.

pjcozzi:
>> angular measurements from the tile's center

> Isn't that the same as latitude/longitude or do you mean polar coordinates in the tangent plane?

Polar coordinates in the tangent plane

>> Perhaps something in the tile's tangent plane

> I like this idea, but I would be concerned with points on the plane with large distances to the ellipsoid. We would also run into the same problems that we have with triangulating polygons on the tangent plane in Cesium. Can we restrict the tile regions so they can never get that big?

> Another idea is to store them in 2D (GeographicProjection or WebMercatorProjection). This would be expensive to convert back to WGS84.

All good points, the question still remains:

> can we explore a more concise representation for positions for ground clamped positions? I think we can do even better than quantized x, y, z by taking advantage of the fact that the positions are on the ground.

Perhaps profile to see how bad Cartographic-to-Cartesian is in practice.  Perhaps there are optimizations if the points are all close by (there are definitely optimizations, for example, when the points are on the same line of longitude or latitude).

No rush, but we should explore or decide to punt before finalizing the initialize spec.

bagnell:
Thanks for the initial work here @lasalvavida! I'm closing this since the format has changed significantly. I merged your commits to the `vector-tiles` branch in the main repo. I may draw from the history when we support arbitrary volumes.

PR #134, opened 2016-09-28
Change conditions to array of pairs

lilleyse:
For https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/2#issuecomment-249891323


pjcozzi:
Nice, thanks @lilleyse for the update and @twpayne for the catch!

PR #135, opened 2016-10-10
Change equality comparisons to == and !=

lilleyse:
Related to this discussion: https://github.com/AnalyticalGraphicsInc/cesium/pull/4336#issuecomment-252718422

>If we decide to be loose about type comparisons in the JS implementation we may also want to change === to == to force casting. Right now a style with 5 < "6" would return true but 5 === "5" would be false.

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -64,7 +64,7 @@ The following style assigns the default show and color properties to each featur
 Instead of showing all features, `show` can be an expression dependent on a feature's properties, for example:
 ```json
 {
-    "show" : "${ZipCode} === '19341'"
+    "show" : "${ZipCode} == '19341'"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


pjcozzi:
I think I understand that you want to add `==` and `!=` so that they behave similar to `<` and friends with type coercion, but why are we removing the JavaScript best practice `===` and `!==` operators?  Because they can't be implemented with the GLSL transpiler?  If so, please explain.

lilleyse:
Oh I was thinking we would not support both == and === at the same time, but that's definitely doable.

pjcozzi:
OK, let's do it, thanks.

pjcozzi:
@lilleyse can you please update this PR and also update the styling spec to include changes needed for point clouds, e.g., `pointSize` in this PR or a new one?

lilleyse:
Updated. I'll open a new PR for the other styling and point cloud changes.

pjcozzi:
Looks good.

PR #138, opened 2016-10-21
Add point cloud styling section

lilleyse:
Added `pointSize`, `${POSITION}`, `${COLOR}`, `${NORMAL}`

@pjcozzi I'm in the process of writing a section for shader styling. Did we decide on a final solution for the current GLSL styling limitations? I have the current issues in https://github.com/AnalyticalGraphicsInc/cesium/issues/3241 under the `Point Cloud Shader Styling` section:

> - [ ] **Point Cloud Shader Styling**
>   - [ ] Unsupported:
>     - [ ] `isNan`, `isFinite`(supported in later versions of GLSL with `isnan` and `isinf`)
>     - [ ] Strings, not supported in the GLSL styling engine or the batch table binary
>     - [ ] Regular expressions
>     - [ ] Arrays that aren't length 2, 3, or 4.
>     - [ ] null and undefined - use a default value like 0?
>     - [ ] type mismatching - comparing floats to bools to vecn will result in shader compilation errors

and some notes here: https://github.com/AnalyticalGraphicsInc/cesium/pull/4336#issuecomment-250238675

I am also going to write a section for the new `ColorBlendMode` types in https://github.com/AnalyticalGraphicsInc/cesium/pull/4451. Technically this is not part of the declarative styling language though, so should it go in the `Styling` section or the main README ?

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -23,9 +23,8 @@ Example: Creating a color ramp based on building height.
 * Gabby Getz, [@ggetz](https://github.com/ggetz)
 * Matt Amato, [@matt_amato](https://twitter.com/matt_amato)
 * Tom Fili, [@CesiumFili](https://twitter.com/CesiumFili)
-* Patrick Cozzi, [@pjcozzi](https://twitter.com/pjcozzi)
 * Sean Lilley, [@lilleyse](https://github.com/lilleyse)
-
+* Patrick Cozzi, [@pjcozzi](https://twitter.com/pjcozzi)
 
 Contents:
 
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -472,7 +472,7 @@ ${temperatures['values'][0]} === 70 // Same as (temperatures[values])[0] and tem
 
 ### Point Cloud
 
-In addition to evaluating a point's `color` and `show` properties, a point cloud style may evaluate `pointSize`, or the size of each point in pixels. The default `pointSize` is `1.0`.
+A [Point Cloud](../TileFormats/PointCloud/README.md) is a collection of points that may be styled like other features. In addition to evaluating a point's `color` and `show` properties, a point cloud style may evaluate `pointSize`, or the size of each point in pixels. The default `pointSize` is `1.0`.
 ```json
 {
     "color" : "color('red')",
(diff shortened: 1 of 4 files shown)


pjcozzi:
> Did we decide on a final solution for the current GLSL styling limitations?

No.  The plan is to continue to work on the implementation and see how impactful these limitations are, then we will be in a better position to decide.

Can you submit an issue to this repo labeled "Draft 1.0" with the above info and then link to it from the Cesium roadmap issue?

pjcozzi:
> Technically this is not part of the declarative styling language though, so should it go in the Styling section or the main README ?

The spec is about the spec, not the Cesium implementation.  I think the spec should say that the computed color from the style can be applied in an application-specific way, e.g., replace, modulate, mix.  We just have to carefully word so that a valid implementation can't just set the color to black and say it conforms.

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+
+In addition to evaluating a point's `color` and `show` properties, a point cloud style may evaluate `pointSize`, or the size of each point in pixels. The default `pointSize` is `1.0`.
+```json
+{
+    "color" : "color('red')",
+    "pointSize" : "${Temperature} * 0.5"
+}
+```
+
+Implementations may clamp the evaluated `pointSize` to the system's supported point size range. For example, WebGL renderers may query `ALIASED_POINT_SIZE_RANGE` to get the system limits when rendering with `POINTS`.
```
pjcozzi: Perhaps say that a pointSize of 1.0 must be supported.

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+
+Point cloud styles may also reference per-point semantics including position, color, and normal to allow for more flexible styling of the source data.
+* `${POSITION}` is an array of three values representing the xyz coordinates of the point before the `RTC_CENTER` and tile transform are applied. When the positions are quantized, `${POSITION}` refers to the position after the `QUANTIZED_VOLUME_SCALE` is applied, but before `QUANTIZED_VOLUME_OFFSET` is applied.
```
pjcozzi: Precisely say what coordinate system that is using the same terms in the point cloud spec.

pjcozzi: Also precisely say the type of each component, e.g., `float`.

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+* `${COLOR}` evaluates to a Color type, where each of the rgba color components are in the range `0.0` to `1.0`.
```
pjcozzi: Precisely state component type.

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+* `${NORMAL}` is an array of three values representing the normal of the point before the tile transform is applied. When normals are oct-encoded `${NORMAL}` refers to the decoded normal.
```
pjcozzi: Precisely state coordinate system and component type.

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+```json
+{
+    "color" : "color('red')",
+    "pointSize" : "${Temperature} * 0.5"
+}
+```
+
+Implementations may clamp the evaluated `pointSize` to the system's supported point size range. For example, WebGL renderers may query `ALIASED_POINT_SIZE_RANGE` to get the system limits when rendering with `POINTS`.
+
+Point cloud styles may also reference per-point semantics including position, color, and normal to allow for more flexible styling of the source data.
```
pjcozzi: "semantics from the Feature Table"

Styling/README.md
```diff
@@ -469,6 +470,37 @@ ${temperatures.values[0]} === 70
...
+* `${POSITION}` is an array of three values representing the xyz coordinates of the point before the `RTC_CENTER` and tile transform are applied. When the positions are quantized, `${POSITION}` refers to the position after the `QUANTIZED_VOLUME_SCALE` is applied, but before `QUANTIZED_VOLUME_OFFSET` is applied.
+* `${COLOR}` evaluates to a Color type, where each of the rgba color components are in the range `0.0` to `1.0`.
+* `${NORMAL}` is an array of three values representing the normal of the point before the tile transform is applied. When normals are oct-encoded `${NORMAL}` refers to the decoded normal.
+
+For example:
+
+```json
+{
+    "color" : "${COLOR} * color('red')'",
+    "show" : "${POSITION}[0] > 0.5",
```
pjcozzi: Should these be arrays or .xyz and .rgba?  I get that the later is more work since we have to introduce a vec3 type to the styling language, but it is more natural.

pjcozzi:
Bump.  @lilleyse let's not forget about this one.

lilleyse:
Updated with reference to https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/166

pjcozzi:
@lilleyse what else is need here after #166?

lilleyse:
That should be it, unless you see any areas to change.

lilleyse:
The `Point Cloud Shader Styling` section is still a TODO because we haven't fully fleshed out the rules for typechecking and other things.

pjcozzi:
> The Point Cloud Shader Styling section is still a TODO because we haven't fully fleshed out the rules for typechecking and other things.

Please submit an issue for this labeled `1.0` and `styling` by flushing out the tasklist and Cesium PR linked to at the top here.

In the meantime, it is OK to merge this with the TODO.

lilleyse:
This issue should have all the current GLSL limitations: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/140

PR #143, opened 2016-10-31
Added abs function

lilleyse:
@pjcozzi 

Cesium implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/4545

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -40,6 +40,7 @@ Contents:
       * [RegExp](#regexp)
    * [Conversions](#conversions)
    * [Variables](#variables)
+   * [Built-in Functions](#built-in-functions)
    * [Notes](#notes)
 * [File Extension](#file-extension)
 * [MIME Type](#mime-type)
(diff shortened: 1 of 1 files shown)


pjcozzi:
Part of #2

pjcozzi:
In GLSL, `abs` takes `genType`, https://www.opengl.org/sdk/docs/man4/html/abs.xhtml, so, for example, it works with vectors.

Do we want to do that now or do you want to start with just scalars for this and similar functions?

pjcozzi:
When you open a PR, please include a direct link to the part of the spec updated, e.g., in this case: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/abs/Styling#built-in-functions

CC @leesafini @Dylan-Brown

pjcozzi:
Add some intro text to the built-ins section, and also a sentence or two to https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/abs/Styling#overview

lilleyse:
> In GLSL, abs takes genType, https://www.opengl.org/sdk/docs/man4/html/abs.xhtml, so, for example, it works with vectors.

> Do we want to do that now or do you want to start with just scalars for this and similar functions?

Maybe wait until vectors are in the spec.

pjcozzi:
> Maybe wait until vectors are in the spec.

OK, please add to the roadmap, #2.

lilleyse:
> OK, please add to the roadmap, #2.

Done

lilleyse:
Ready for another look.

PR #144, opened 2016-10-31
batchId type update

pjcozzi:
As discussed in https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/141#issuecomment-257335562.

@lilleyse does this require any Cesium implementation updates?

Note that `pnts`/`i3dm` are slightly different than `b3dm` since `b3dm` would allow float batchids.  I do not think that is appropriate in the context of `pnts`/`i3dm`.

### DIFF
--- a/TileFormats/Batched3DModel/README.md
+++ b/TileFormats/Batched3DModel/README.md
@@ -42,7 +42,7 @@ in the Cesium implementation of 3D Tiles.
 
 ## Batch Table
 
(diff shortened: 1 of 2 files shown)


lilleyse:
Just this change: https://github.com/AnalyticalGraphicsInc/cesium/pull/4561

pjcozzi:
Merged https://github.com/AnalyticalGraphicsInc/cesium/pull/4561.

PR #146, opened 2016-11-03
Allow more flexible naming for the batchId attribute

lilleyse:
Instead of requiring the `batchId` attribute to be named `a_batchId`, this allows it to be named anything as long as the parameter semantic is `BATCHID`.

Cesium implementation here: https://github.com/AnalyticalGraphicsInc/cesium/pull/4583


pjcozzi:
Perfect, thanks!

PR #149, opened 2016-11-03
updated link in i3dm tile spec

JudyWeng:
The link for "tile transform" was broken in i3dm's tile spec under Instance Orientation


pjcozzi:
Thanks for the fix, @JudyWeng!

PR #150, opened 2016-11-04
updated readme for sqrt documentation

leesafini:
AnalyticalGraphicsInc/cesium/pull/4606

- Added documentation for `sqrt`
- Added self to Acknowledgments

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -476,6 +476,7 @@ The following built-in functions are supported by the styling language:
 
 * [`abs`](#abs)
 * [`cos`](#cos)
+* [`sqrt`](#sqrt)
 
 #### abs
 
@@ -498,6 +499,18 @@ Returns the cosine value of the number (in radians).
 ```json
 {
     "show" : "cos(${temperature}) > 0.0"
+}
+```
+
+#### sqrt 
+
+`sqrt(value : Number) : Number`
+
+Returns the square root value of the number.
+
+```json
+{
+    "show" : "sqrt(${temperature}) > 0.0"
 }
 ```
 
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -511,11 +511,11 @@ Returns the square root of value if value >= 0. Returns NaN when value < 0.
 ```json
 {
     "color" : {
-    "expression" : "sqrt(${temperature}",
-    "conditions" : [
-        ["${expression} >= 0.5", "color('#00FFFF')"],
-        ["${expression} >= 0.0", "color('#FF00FF')"]
-    ]
+        "expression" : "sqrt(${temperature}",
+        "conditions" : [
+            ["${expression} >= 0.5", "color('#00FFFF')"],
+            ["${expression} >= 0.0", "color('#FF00FF')"]
+        ]
     }
 }
 ```
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -511,13 +511,11 @@ Returns the square root of value if value >= 0. Returns NaN when value < 0.
 ```json
 {
     "color" : {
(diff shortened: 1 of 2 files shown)


Styling/README.md
```diff
@@ -501,6 +502,18 @@ Returns the cosine value of the number (in radians).
 }
 ```
 
+#### sqrt 
+
+`sqrt(value : Number) : Number`
```
pjcozzi: What is the range of `value`?  `>= 0.0`?  What is the behavior if, for example, `-1.0` is provided?

Styling/README.md
```diff
@@ -501,6 +502,18 @@ Returns the cosine value of the number (in radians).
...
+
+Returns the square root value of the number.
+
+```json
+{
+    "show" : "sqrt(${temperature}) > 0.0"
```
pjcozzi: Perhaps use a more realistic example like use `sqrt` to create a color ramp using the `"color"` property instead of `"show"`.

leesafini:
Made changes based on your comments @pjcozzi. Was this what you had in mind for the color ramp?

pjcozzi:
> Was this what you had in mind for the color ramp?

Yes, but use `expression` so it is more efficient.  Also, only two or three conditions are needed to get the point across; it's a good practice to make the example as simple as possible, but no more simple.  Here's an example from the spec:

```
{
    "color" : {
        "expression" : "regExp('^1(\\d)').exec(${id})",
        "conditions" : [
            ["${expression} === '1'", "color('#FF0000')"],
            ["${expression} === '2'", "color('#00FF00')"],
            ["true", "color('#FFFFFF')"]
        ]
    }
}
```

> Also, should I be adding myself to the contributers list for this specific page?

Let's stick with the main acknowledgements for now; we'll likely update each page to have editors and contributors like the [glTF spec](https://github.com/KhronosGroup/glTF/blob/master/specification/README.md) and then we would move your name there; right now "contributors" kind of means "editors."

leesafini:
@lilleyse updated!

Styling/README.md
```diff
@@ -501,6 +502,25 @@ Returns the cosine value of the number (in radians).
...
+Returns the square root of value if value >= 0. Returns NaN when value < 0.
+
+```json
+{
+    "color" : {
+    "expression" : "sqrt(${temperature}",
+    "conditions" : [
+        ["${expression} >= 0.5", "color('#00FFFF')"],
+        ["${expression} >= 0.0", "color('#FF00FF')"],
+        ["true", "color('#FFFFFF')"]
```
lilleyse: You can remove this last line since it should always meet the two previous conditions.

leesafini: done

Styling/README.md
```diff
@@ -501,6 +502,24 @@ Returns the cosine value of the number (in radians).
...
+
+```json
+{
+    "color" : {
+    "expression" : "sqrt(${temperature}",
+    "conditions" : [
+        ["${expression} >= 0.5", "color('#00FFFF')"],
+        ["${expression} >= 0.0", "color('#FF00FF')"]
+    ]
+    }
```
lilleyse: The indentation is still a bit off.

leesafini: What is it supposed to be?

lilleyse:
```
"color" : {
    "expression" : "sqrt(${temperature}",
    "conditions" : [
        ["${expression} >= 0.5", "color('#00FFFF')"],
        ["${expression} >= 0.0", "color('#FF00FF')"]
    ]
}
```

lilleyse:
Thanks @leesafini!

PR #154, opened 2016-12-05
Change BATCHID type from 5123 to 5126

lilleyse:
From discussion:
https://github.com/AnalyticalGraphicsInc/cesium/pull/4684#discussion_r90712835
https://github.com/KhronosGroup/glTF/issues/789


pjcozzi:
What is the point of this change?  Just to update the example, but not change the requirements for a batch id?

lasalvavida:
Currently in some files, we have the `batchId` listed as a `float` in the technique for the shader, even though it is an integer. I think correcting that is the point of this change.

pjcozzi:
Ah, OK.

PR #156, opened 2016-12-12
Fix links

jbo023:
fixing links to tilesetjson section, and we moved our demo applications to a new domain.


lilleyse:
Thanks @jbo023

PR #160, opened 2016-12-16
Prefix BATCHID semantic with an underscore

lasalvavida:
@pjcozzi, from https://github.com/AnalyticalGraphicsInc/gltf-pipeline/pull/191.

[Link](https://github.com/lasalvavida/3d-tiles/tree/batchid-prefix-underscore/TileFormats/Batched3DModel)

Also do we want to change the semantic to the more correct `_BATCH_ID -> batchId` with an underscore between `BATCH` and `ID` since we're making this change anyway?


pjcozzi:
> Also do we want to change the semantic to the more correct _BATCH_ID -> batchId with an underscore between BATCH and ID since we're making this change anyway?

No, the glTF semantics do not do this: https://github.com/lexaknyazev/glTF/tree/master-test/specification/1.1#semantics

pjcozzi:
@lasalvavida can you also please open a standalone PR into Cesium's `3d-tiles` branch for this, including backwards compatibility?

PR #165, opened 2017-01-03
Header length update

### LINKED ISSUE #164: Batched 3d model spec readme error
There is an error on the Batched 3d model spec readme file. 
On the second paragraph of the binary glTF section says "Binary glTF immediately follows the batch table. It begins **20** + batchTableByteLength bytes from the start of the arraybuffer " when it really starts at **24** + batchTableByteLength bytes from the start

lilleyse:
Fixes #164


pjcozzi:
Good eye!

PR #166, opened 2017-01-09
Vector types: vec2, vec3, vec4

lilleyse:
For #2 

https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/138 will build off this.

Cesium implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/4759

To Do:
* [ ] Need to update the built-in functions with vector support. Currently the built-in functions are on https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/math-functions which will be merged relatively soon. There are just a few more unary operations remaining.

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -284,11 +284,20 @@ The styling language includes 2, 3, and 4 component floating-point vector types:
 
 ##### Vector usage
 
-`vec2` components may be accessed with `.x`, `.y`; `.r`, `.g`; or `[0]`, `[1]`.
+`vec2` components may be accessed with
+* `.x`, `.y`
+* `.r`, `.g`
+* `[0]`, `[1]`
 
-`vec3` components may be accessed with `.x`, `.y`, `.z`; `.r`, `.g`, `.b`; or `[0]`, `[1]`, `[2]`.
+`vec3` components may be accessed with
+* `.x`, `.y`, `.z`
+* `.r`, `.g`, `.b`
+* `[0]`, `[1]`, `[2]`
 
-`vec4` components may be accessed with `.x`, `.y`, `.z`, `.w`; `.r`, `.g`, `.b`, `.a`; or `[0]`, `[1]`, `[2]`, `[3]`.
+`vec4` components may be accessed with
+* `.x`, `.y`, `.z`, `.w`
+* `.r`, `.g`, `.b`, `.a`
+* `[0]`, `[1]`, `[2]`, `[3]`
 
 Unlike GLSL, the styling language does not support swizzling. For example `vec3(1.0).xy` is not supported.
 
@@ -305,7 +314,7 @@ Operations between vectors of different types will not evaluate component-wise, 
 
 #### Color
 
-Colors are created with one of the following functions:
+Colors are implemented as `vec4` and are created with one of the following functions:
 * `color() : Color`
 * `color(keyword : String, [alpha : Number]) : Color`
 * `color(6-digit-hex : String, [alpha : Number]) : Color`
@@ -335,10 +344,10 @@ Colors defined with `rgba` or `hsla` have a fourth argument that is an alpha com
 * `rgba(100, 255, 190, 0.25)`
 * `hsla(1.0, 0.6, 0.7, 0.75)`
 
-Colors are implemented as `vec4` and share the same functions, operators, and component accessors. Color components are stored in the range `0.0` to `1.0`.
+Colors are equivalent to the `vec4` type and share the same functions, operators, and component accessors. Color components are stored in the range `0.0` to `1.0`.
 
 For example:
-* `color('red').x`, `color('red').r`, and `color('red')[0]` evaluate to `1.0`.
+* `color('red').x`, `color('red').r`, and `color('red')[0]` all evaluate to `1.0`.
 * `color('red').toString()` evaluates to `(1.0, 0.0, 0.0, 1.0)`
 * `color('red') * vec4(0.5)` is equivalent to `vec4(0.5, 0.0, 0.0, 1.0)`
 
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -23,6 +23,7 @@ Example: Creating a color ramp based on building height.
 * Gabby Getz, [@ggetz](https://github.com/ggetz)
 * Matt Amato, [@matt_amato](https://twitter.com/matt_amato)
 * Tom Fili, [@CesiumFili](https://twitter.com/CesiumFili)
+* Sean Lilley, [@lilleyse](https://github.com/lilleyse)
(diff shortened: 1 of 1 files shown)


pjcozzi:
> [ ] Should it be possible to do component-wise operations between Color and vec4?

What does this mean `add(color, vector)`?

> [ ] Similarly can a style that evaluates color be set to a vec4? In the Cesium GLSL implementation this automatically works because both types are represented by vec4. E.g. color : vec4(1.0)
> [ ] If we don't go with the above two, we should at least support casting between colors and vectors. Like color(vec4(1.0)) and vec4(rgba(0,0,0,1))

Do we need a `Color` type at all?  Is it reasonable to implement it as `vec4` and be just as easy to use?

If not, in the GLSL world, these would be separate types so I suppose they should **not** have implicit conversion and instead should require explicit construction (`color(vec4(1.0))` and `vec4(rgba(0,0,0,1))`).  Is that hard to implement?

Styling/README.md
```diff
@@ -281,13 +291,66 @@ Colors store rgba components internally, where each component is in the range `0
...
+* `vec4(Number)` - initialize each component with the number
+* `vec4(Number, Number, Number, Number)` - initialize with four numbers
+* `vec4(vec4)` - initialize with another `vec4`
+* `vec4(vec2, Number, Number)` - initialize with a `vec2` and two numbers
+* `vec4(Number, vec2, Number)` - initialize with a `vec2` and two numbers
+* `vec4(Number, Number, vec2)` - initialize with a `vec2` and two numbers
+* `vec4(vec3, Number)` - initialize with a `vec3` and number
+* `vec4(Number, vec3)` - initialize with a `vec3` and number
+
+`vec2` components may be accessed with `.x`, `.y` and `[0]`, `[1]`.
```
pjcozzi: Looks like this is in the **vec4** section.  Either create a new section here or move some of these sentences to be in the right vector type section just above.

Styling/README.md
```diff
@@ -281,13 +291,66 @@ Colors store rgba components internally, where each component is in the range `0
...
+
+`vec4` components may be accessed with `.x`, `.y`, `.z`, `.w` and `[0]`, `[1]`, `[2]`, `[3]`.
+
+Unlike GLSL, the styling language does not support swizzling. For example `vec3(1.0).xy` is not supported.
+
+Vectors support the following unary operators: `-`, `+`.
+
+Vectors support the following binary operators by performing component-wise operations: `===`, `==`, `!==`, `!=`, `+`, `-`, `*`, `/`, and `%`.  For example `vec4(1.0) === vec4(1.0)` is true since the x, y, z, and w components are equal.  This is not the same behavior as a JavaScript `Object`, where, for example, reference equality would be used.  Operators are essentially overloaded for `vec2`, `vec3`, and `vec4`.
+
+Operations between vectors of different types will not evaluate component-wise, but instead will evaluate as JavaScript objects. For example `vec2(1.0) === vec3(1.0)` is false and `vec2(1.0) * vec4(1.0)` is `NaN`. See the [Conversions](#conversions) section for more details.
```
pjcozzi: This is awkward.  Can we do better?

> `vec2(1.0) === vec3(1.0)`

What does this do in GLSL?  Based on the constructor overloads, is one converted to the other?

Instead of returning `false`/`NaN` for these examples, is it reasonable to implement if these are type mismatch errors?

pjcozzi:
Just those comments.

Styling/README.md
lilleyse: GLSL will not compile with that, you need to cast explicitly. 

Equality checks could be made to work without too much trouble, the multiply would be harder (is the result vec2 or vec4?). Originally all this would work if `vec2`, `vec3`, and `vec4` were treated as the same `vec4` type, but I thought the point of moving away from that was to prevent expressions like these from working.

lilleyse:
> What does this mean add(color, vector)?

Expressions like `color() + vec4(1.0)` or `color() * vec4(1, 0, 0, 1)`

> If not, in the GLSL world, these would be separate types so I suppose they should not have implicit conversion and instead should require explicit construction (color(vec4(1.0)) and vec4(rgba(0,0,0,1))). Is that hard to implement?

This shouldn't be too hard to implement.

> Do we need a Color type at all? Is it reasonable to implement it as vec4 and be just as easy to use?

I think its nice to have a color type in case we want to add special functions on colors. At the moment though its nearly the same as a `vec4`.

Styling/README.md
pjcozzi: Right, is it reasonable to implement if these are type mismatch errors?

pjcozzi:
I _think_ the right thing to do is to use `vec4` for colors.  This is not a huge spec update.  How big is the code update?

If we need color-specific functions, they would just be stand alone functions, not member functions.

GLSL is a pretty good example of getting good milage out of using `vec4` for colors.  You could also evaluate the Cesium color and Cartesian types to see if this is reasonable.

As simple as possible, but no more simple.

Styling/README.md
lilleyse: Oh I think I misunderstood your comment.

It shouldn't be hard to detect type mismatch errors in the JS backend, as usual it's more difficult in the GLSL backend, but still possible to just catch a shader failure and throw a generic type mismatch error (as talked about offline).

lilleyse:
> I think the right thing to do is to use vec4 for colors. This is not a huge spec update. How big is the code update?

The code update is not too bad at all. This approach sounds good with me too, more flexible.

Styling/README.md
pjcozzi: Great, let's go with the type mismatch.

pjcozzi:
>> I think the right thing to do is to use vec4 for colors. This is not a huge spec update. How big is the code update?

> The code update is not too bad at all. This approach sounds good with me too, more flexible.

Great, let's do it!

lilleyse:
I opened a PR for handling colors as vec4. https://github.com/AnalyticalGraphicsInc/cesium/pull/4849

I'll also have a PR for checking type mismatching.

pjcozzi:
@lilleyse given https://github.com/AnalyticalGraphicsInc/cesium/pull/4867, is this ready to finish?

lilleyse:
This is ready for another look. The new changes describe colors in terms of vectors.

There will be a separate PR for type mismatching since that affects pretty much everything including vectors.

>  Need to update the built-in functions with vector support. Currently the built-in functions are on https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/math-functions which will be merged relatively soon. There are just a few more unary operations remaining.

I'm going to open a PR soon for the built-in math functions, and in that PR it will support vector types.

pjcozzi:
Looks great!

PR #171, opened 2017-01-18
Batch Table Hierarchy

lilleyse:
For https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/66
Cesium implementation : https://github.com/AnalyticalGraphicsInc/cesium/pull/4625

This is a rough draft for the batch table binary.

To-do:

* [x] Batch Table Hierarchy schema


lilleyse:
This is ready for review now. The Batch Table Hierarchy section in Batch Table is pretty long - maybe it belongs in its own file.

pjcozzi:
@arneschilling @pmconne @jbo023 your input is very welcomed here!

pjcozzi:
**Direct links**
* Batch Table: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/batch-table-hierarchy/TileFormats/BatchTable#hierarchy
* Styling functions: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/batch-table-hierarchy/Styling#batch-table-hierarchy

pjcozzi:
Can you add a TOC to TileFormats/BatchTable/README.md?

Styling/README.md
```diff
@@ -46,6 +46,7 @@ Contents:
    * [Variables](#variables)
    * [Built-in Variables](#built-in-variables)
    * [Built-in Functions](#built-in-functions)
+   * [Batch Table Hierarchy](#batch-table-hierarchy)
```
pjcozzi: I don't think the TOC indentation is right here; a lot of this should not be under **Expressions**.  Can you reorganize?

pjcozzi:
**Class** and **Feature** should be precisely defined; keep in mind this is a spec, not a tutorial.

pjcozzi:
> Class and Feature should be precisely defined; keep in mind this is a spec, not a tutorial.

Also define **Instance**.

Use a diagram if it is helpful.

pjcozzi:
> `classes` is an array of objects, where each object contains the following properties:
> * `name` - the name of the class
> * `length` - the number of instances of the class
> * `instances` - metadata for the instances. This section is similar to a standard batch table; properties may be stored as an array of values or a reference to data in the binary body.

Although the schema is the authority, here and throughout, precisely define the types like the spec does elsewhere, e.g., "A string representing the name of the class"

TileFormats/BatchTable/README.md
```diff
@@ -150,6 +150,423 @@ var cartographicArray = new Float64Array(batchTableBinary.buffer, byteOffset, ca
...
+  }
+}
+```
+
+`classes` is an array of objects, where each object contains the following information:
+* `name` - the name of the class
+* `length` - the number of instances of the class
+* `instances` - metadata for the instances. This section is similar to a standard batch table; properties may be stored as an array of values or a reference to data in the binary body.
+
+`instancesLength` is the number of instances. In this example `instancesLength` equals `batchLength`, however this is not always the case (as will be seen below).
```
pjcozzi: Precisely say why or link to why.  The spec needs to be rigorous.

TileFormats/BatchTable/README.md
```diff
@@ -150,6 +150,423 @@ var cartographicArray = new Float64Array(batchTableBinary.buffer, byteOffset, ca
...
+    "byteOffset" : 0,
+    "componentType" : "UNSIGNED_SHORT"
+};
+```
+
+### Instance Hierarchy
+
+Another limitation of the standard batch table is the difficulty in expressing metadata hierarchies.
+
+For example consider a tile that represents a city block. The block itself contains metadata, the individual buildings contain metadata, and the building components (doors, walls, roofs) contain metadata.
```
pjcozzi: Perhaps a screenshot or a series of screenshots with different features highlighted from the Sandcastle example would be useful here.

pjcozzi:
The tutorial style content is good, but we need to better separate what is precisely written spec and what is an example.  Using the example to define the spec is atypical.

pjcozzi:
Made some tweaks in 9c2679d21e3032fd7b1865170e0f045b7c36e337.

pjcozzi:
Made a few more tweaks in 22a44e33d84774ce5bbf7ae36318120ccaf66728.

pjcozzi:
That's all my comments; this looks good.  Let me know when the updates and schema are ready.

lilleyse:
Thanks for the first round of reviewing.

> The tutorial style content is good, but we need to better separate what is precisely written spec and what is an example. Using the example to define the spec is atypical.

Definitely agree here, I'll work on this as well as creating some diagrams.

arneschilling:
"@arneschilling @pmconne @jbo023 your input is very welcomed here!"

We can definitely work with this approach, thank you very much. Something is not quite clear to me. The attribute names in the classes section seem to have prefixes, e.g. "door_name" and "building_name" for classes Door and Building. Is this a naming convention or can we use the original attribute names, in our case just "name" for each class? Adding a prefix to each would alter the way we present batch table information to the user, because we cannnot always implement a dictionary mapping these strings values to more user friendly text. 

From a Logical Point of view, I have one comment.
The hierarchy ("parentIds" field) and the classes seem to be independant from each other, even though classes is embedded in "HIERARCHY". One could in fact implement classes without using the parentIds as in the first example with "Lamp", "Car" and "Tree". If somebody has a collection of independant object classes without interrelations between them, the hierarchy is not needed. On the other hand, the "parentIds" field could be used in combination with a standard batch table without classes (currently this is our implementation). Why not separate these 2 things?

lilleyse:
Thanks for the feedback.

The attribute names don't have to have prefixes, this was just for example purposes. In my next round of edits I will clarify that.

If someone is using the concept of parents with the standard batch table (no classes) I think they would be better off storing parent id's as a regular batch table property and have hierarchy logic be part of the app.

pjcozzi:
@lilleyse what's the latest here?  Is this ready?

pmconne:
Having finally gotten a chance to play around with this in earnest, I have a couple of questions:

1. Does this work if the features have no properties of their own, only those which they inherit from their parents? e.g.:
    "classes": [
        {
        "name": "Wall",
        "length": 3,
        "instances": { /* I have no properties... */ }
        }

2. Would it be worth optimizing for the case in which every feature has the same class ID (and - at least in our case - same number of parents), rather than having the first N entries in the classIds and parentCounts arrays be identical?

I think we may have an unusual use case - we want to use this purely for classification of features, with all the properties residing on the abstract "classification" classes. We can do that with the existing spec - redundant data notwithstanding. If the answer to question 1 is "yes" then the redundancy mentioned in question 2 will not be much of an issue.

pmconne:
To answer my own question...yes, instances with no properties are fine.
This is working wonderfully - thanks.

lilleyse:
Second draft updated. I still think it may be worth splitting it off into a separate file - what do you think @pjcozzi.

The batch table schema is also updated.

Styling/README.md
lilleyse: It fits ok, as those sections describe expressions related to the hierarchy and point clouds. I'm fine with placing them in a different section too.

pjcozzi: Yes, I think this should be in a top-level section, not under Expressions.

TileFormats/BatchTable/README.md
```diff
@@ -147,6 +164,344 @@ var cartographicArray = new Float64Array(batchTableBinary.buffer, byteOffset, ca
...
+
+* `isExactClass`
+* `isClass`
+* `getExactClassName`
+
+More detailed descriptions are provided in the [Styling Spec](../../Styling/README.md#batch-table-hierarchy).
+
+### Notes
+
+* In some cases a feature may have multiple ancestors of the same class, or ancestors of different classes with the same property names. It is up to the implemantation to decide how to handle overloaded properties.
```
pjcozzi: Can you provide a concrete example here.

> It is up to the implemantation to decide how to handle overloaded properties.

In general, we need to avoid undefined behavior as much as possible.  Are we sure we can't define reasonably behavior here?

TileFormats/BatchTable/README.md
```diff
@@ -147,6 +164,344 @@ var cartographicArray = new Float64Array(batchTableBinary.buffer, byteOffset, ca
...
+{
+  "Height" : [...],
+  "Longitude" : [...],
+  "Latitude" : [...],
+  "HIERARCHY" : {...}
+}
+```
+
+* The batch table hierarchy is self-contained within the tile. It is not possible to form metadata hierarchy across different tiles in the tileset.
+
```
pjcozzi: Below in

> TileFormats/BatchTable/figures/batch-table-hierarchy-block.png

"Labelled" should be "Labeled"

pjcozzi: And in batch-table-hierarchy-parking-lot.png.

pjcozzi:
Looks good, this is very close.

TileFormats/BatchTable/README.md
lilleyse: An instance has two parents of the same class. Anytime the instance gets a property it will have to choose between which parent to get it from. Cesium's behavior is to get it from the first parent.

An instance may also have two different ancestors of different classes that happen to use the same property name, like `id`. In Cesium it will just return the property from whichever ancestor it encounters first, so similar to the situation above.

I left this open ended at first, but I think this behavior could be defined pretty easily here.

lilleyse: Once I started to rewrite that section I realized it may be hard to define the exact behavior, which would be equivalent to defining how the hierarchy should be traversed. I did make the wording less open-ended - so it is not allowed to return an array of values, or just do nothing.

pjcozzi:
Thanks!

PR #174, opened 2017-01-31
Type Mismatching

lilleyse:
This PR makes the styling language stricter about types and less reliant on JavaScript conventions.

* Removed `!=` and `==`
* Added a section about operator rules. Specifies which operators accept which types
* Added a section about string conversions.
* Added the Array type to more areas, I think it was just oversight that it was missing before
* Removed `Matrix` types from batch table binary - they were never implemented in the shader backend in Cesium and I can't think of a great use case for them anymore

Cesium implementation : https://github.com/AnalyticalGraphicsInc/cesium/pull/4860 and https://github.com/AnalyticalGraphicsInc/cesium/pull/4940

### DIFF
--- a/Styling/README.md
+++ b/Styling/README.md
@@ -41,7 +41,9 @@ Contents:
       * [vec4](#vector)
       * [Color](#color)
       * [RegExp](#regexp)
-   * [Conversions](#conversions)
+   * [Operator Rules](#operator-rules)
+   * [Type Conversions](#type-conversions)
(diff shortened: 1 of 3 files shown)


pjcozzi:
All looks good, just one question: remind me where arrays are currently used in practice?

lilleyse:
They are allowed as properties in the batch table json. Otherwise they aren't used for any special purposes.

PR #178, opened 2017-02-10
Length to half-length for box schema

lilleyse:
The values in the box should be half lengths, not full lengths of the axis directions.


pjcozzi:
Good eye, thanks @lilleyse!

PR #182, opened 2017-02-15
Added gltfUpAxis property

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/175

Cesium implementation : https://github.com/AnalyticalGraphicsInc/cesium/pull/5005#issuecomment-279902013


pjcozzi:
Looks good.  Is there is need for -X, -Y, -Z as discussed offline?

lilleyse:
> Is there is need for -X, -Y, -Z as discussed offline?

I don't think so. This doesn't seem very common in practice.

pjcozzi:
@shehzan10 this is likely to be of interest to you.

PR #200, opened 2017-03-28
Add feature table to b3dm

lilleyse:
This adds feature table support to b3dm tiles. With future enhancements like [explicit tiling schemes](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/92), [per-feature bounding volumes](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/127), and others down the road, it will be important for b3dm to have a feature table to store per-feature information that cannot be contained within the glTF model. This means `batchIds` are still part of the glTF, but per-feature bounding volumes (for example) would be part of the feature table.

New layout:

layout

The `batchLength` item in the header is removed and now moved to a `BATCH_LENGTH` property of the feature table. This is similar to `INSTANCES_LENGTH` for i3dm and `POINTS_LENGTH`/`BATCH_LENGTH` for pnts.

There was some discussion in the past regarding glTF byte alignment, since 28 is not on an even 8-byte split. The normal recommendation would still apply here, just pad the feature table or, if it exists, the batch table.

In the Cesium implementation we will continue to detect legacy headers, so old tilesets will still work for the time being.


pierotofy:
Looks good to me!

lilleyse:
Cesium implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/5165 ... and more to come if the changes here look good.

pjcozzi:
@pmconne heads up for this small change to the b3dm header.

Issue #90, opened 2016-04-18
Generating tilesets dynamically


pjcozzi:
What is the best approach for tileset.json when generating tilesets dynamically like in https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/22#issuecomment-209977268?

pjcozzi:
CC #89 

> It would then build tileset.json on the fly from that data. This would allow for tileset.json subsets to be generated from larger datasets and possibly other types of dynamic data generation (like merging to tilesets on the fly?)

elfprince13:
I would like to see one of two things, both of which are premised on lazy generation (the real win for generating tilesets dynamically should be not having to store the whole structure in memory at once...):

- Allow the spec of a child node to be given by a URL instead of inline data. This would essentially allow a filesystem-like approach to be taken, but would require cooperation from the server.
- The ability to specify a generator function to be invoked as a callback when the next child is needed. This could trivially subsume the functionality of the previous suggestion, but would require more careful planning around API design.

pjcozzi:
Thanks for the input, @elfprince13!

> [edit]
I think my first suggestion is more-or-less external tilesets...

Agreed.

> The ability to specify a generator function to be invoked as a callback when the next child is needed. This could trivially subsume the functionality of the previous suggestion, but would require more careful planning around API design.

Can you provide a concrete use case?  Is this purely a client-side implementation?  Or does this impact the spec?

elfprince13:
I think it would best be done as part of the spec. Rather than specifying:
```
"root": {
    /* Various other properties here */
    "children": [..]
}
```
You would specify something akin to this:
```
"root": {
    /* Various other properties here */
    "lazy-children": {
        "callback" : "name-of-function-to-call",
        "context" : /* Any value may appear here. It's purpose is only to be passed opaquely  to the callback */,
         "numchildren" : 8 /* Maybe this is an oct tree */
    }
}
```
To access a specific child, the callback could then be invoked by the implementation using something akin to [this](http://stackoverflow.com/a/359910), and passed both the context value and the index of the desired child (and the resulting object could be memoized for efficiency). The meta-goal is simulating lazy evaluation with closures (the pattern here is essentially a generalization of [this tutorial](http://blog.jeremyfairbank.com/javascript/functional-javascript-streams-2/) to acyclic datastructures with more than one child per node). A trivial concrete use case would be generating an oct-tree as needed without having to load the entire structure into memory.

[...]

PR #213, opened 2017-04-21
Remove tile extension restriction

lilleyse:
Addresses https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/90#issuecomment-246207288

Cesium Implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/5206


pjcozzi:
Each tile format spec (including the PR for vector tiles) should be updated to explicitly mention it is optional; validation implementations ignore it and use `magic`.

For example: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Instanced3DModel#file-extension

lilleyse:
Updated, if the wording seems fine I'll update vector tiles.

pjcozzi:
Looks good, please update vector tiles.

Leaving #90 open for its other topics.

lilleyse:
Updated there.

Issue #99, opened 2016-07-06
Tile expiration


pjcozzi:
To support some types of dynamic tilesets that may change on the server, consider adding a per-time expiration time stamp (explicit date/time) or duration (in seconds), which indicate when a tile and its sub-tree become invalid and should be re-requested.

pjcozzi:
@lilleyse what do you think about this new per-tile `expire` object property?

Use `expire.date`:

```javascript
{
  "expire": {
    "date" : "2012-04-30T12:02Z" // Julian date like CZML, https://github.com/AnalyticalGraphicsInc/czml-writer/wiki/CZML-Structure
  },
  "boundingVolume": {
    // ...
  },
  "geometricError": 43.88464075650763,
  "refine" : "add",
  "content": {
    "boundingVolume": {
      // ...
    },
    "url": "2/0/0.b3dm"
  },
  "children": [...]
}
```

Or use `expire.duration`:

```javascript
{
  "expire": {
    "duration" : 30 // seconds  from when the tile is downloaded (not requested, not first rendered, not generated, etc.)
  },
  "boundingVolume": {
    // ...
  },
  "geometricError": 43.88464075650763,
  "refine" : "add",
  "content": {
    "boundingVolume": {
      // ...
    },
    "url": "2/0/0.b3dm"
  },
  "children": [...]
}
```

`expire` can have either a `date` sub-property or a `duration` sub-property (if both are provided, the soonest one is used).  The runtime can convert it to an expiration date, and then request a new tile (and eventually sub-tree) when it expires.  The runtime can still show the expired tile while it waits for the new tile request.  If the request doesn't come back, e.g., because that tile is no longer part of the tileset, then the expired tile should be unloaded (and, of course, unloaded when the request is successful).

For future work for the Cesium implementation, we could also proactively unload expired tiles in our cache that are not visible, but I'll make this a roadmap item.

From the server's perspective, `date` allows the use case for a static file server to serve a tileset with a know expiration date; where as `duration` allows a dynamic server to generate a tile that is valid for a particular amount of time.

lilleyse:
Just so I'm clear, does expire referring the refreshing the whole subtree, or just the content of a single tile? Is this a proeprty worth adding to the `expire` block?

pjcozzi:
I would expire the tile and its sub-tree.  I don't have a use case for expiring just the tile, but it can be added in a backwards-compatible way if good use cases arise.

pjcozzi:
[...]

Issue #92, opened 2016-05-05
Implicit tiling schemes


pjcozzi:
The general spatial data structure in tileset.json does not allow for random access, e.g., "give me tile x,y at level z" because the tiling scheme is not fixed; instead the data structure needs to be searched top-down.

If random access is needed, we could allow a tileset to advertise itself as using a particular typing scheme, e.g.,

``` json
"tilingScheme" : "quadtree"
```

pjcozzi:
Could also be useful for #90.

pjcozzi:
The tileset.json would have to specify the tile format - or each tile could say the formats of its children - and each tiling scheme may have some specific data like a bounding volume if it isn't global, e.g.,
```javascript
{
  "asset": { /* ... */ },
  "tilingScheme": "quadtree",
  "format": "b3dm",
  "boundingVolume" : {
    "region": [ /* ... */ ]  // could default to global
  }
}
```

* Geometric error would either be implicit or part of a new section of the tile's payload
* The tile may also need a new section for children information like child contents' bounding volumes

pjcozzi:
This might have to get pushed post 1.0, but I would at least like to prototype beforehand since this is an important feature for global-scale datasets.

@austinEng perhaps you can help with this once you are ramped up.

pjcozzi:
This should probably be called _implicit tiling scheme_, not _explicit tiling scheme_.  Renamed.

pjcozzi:
For point clouds, also see https://github.com/connormanning/ept

pjcozzi:
**A few potential use cases / interoperability to consider**
* quadtree for global or local terrain or vector data.  Dual-quadtree for Cesium's current [quantized-mesh](https://github.com/AnalyticalGraphicsInc/quantized-mesh) terrain
* octree for interior CAD/BIM, exterior CAD/BIM, point clouds, photogrammetry, 3D buildings, etc.  Local with Cartesian AABB/OBBs and global with geographic regions
* Interop well with GeoPackage, https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/89#issuecomment-465330742
* Interop well with CDB, https://www.opengeospatial.org/standards/cdb

**Implementation notes**
* See https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/92#issuecomment-234416818
* How much better can this make skip LODs, prefetch, etc?
[...]

PR #217, opened 2017-05-01
Tile expiration

lilleyse:
For #99
Cesium implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/4136

One point I have not addressed yet is

> Time expiration should be in the tile (header / feature json) so it can change from request to request.

I wish this didn't have to happen but I can't think of another way.

Each tile's feature table could have a `TILE` object which mimics the layout of a tile in tileset.json. This will also be helpful for bounding volumes in #92 too.


lilleyse:
I started some work on a `TILE_METADATA` section in the feature table. @pjcozzi do you think this is the right idea?

pjcozzi:
> I started some work on a TILE_METADATA section in the feature table. @pjcozzi do you think this is the right idea?

I think we'll need something like this to get full mileage out of explicit tiling schemes, but we might also want it to have child tile metadata for efficient culling.

We have two options
* Push tile expiration post 1.0.  I think it can be added in a backwards compatible way and it is not yet a popular feature.
* Pseudo-design just enough of #92 to make sure we are going down the right path here.

Perhaps we wait to decide based on how fast the rest of the 1.0 work goes?

lilleyse:
> Perhaps we wait to decide based on how fast the rest of the 1.0 work goes?

Sounds good, if expiration doesn't go in that would be ok with me.

lilleyse:
Closing due to inactivity

PR #223, opened 2017-05-22
Added virtualgis demo

pierotofy:
Adds a demo that shows the use of 3D tiles for displaying a long pipeline route.


pjcozzi:
Very nice, thanks for sharing @pierotofy!

PR #227, opened 2017-05-24
Seconds to milliseconds

lilleyse:


lilleyse:
@SunBlack heads up about this change.

Styling/README.md
```diff
@@ -621,7 +621,9 @@ The prefix `tiles3d_` is reserved for built-in variables. The following built-in
 
 #### tiles3d_tileset_time
 
-Gets the time, in seconds, since the tileset was first loaded. This is useful for creating dynamic styles that change with time.
+Gets the time, in milliseconds, since the tileset was first loaded. This is useful for creating dynamic styles that change with time.
+
+_Implementation note: If this value is stored as a 32-bit floating point number, be aware of increasing loss of precision over time._
```
pjcozzi: Why not more precisely mention 4.6 hours as you did in the code PR?

lilleyse: I realized later that it's kind of arbitrary. That's when integer precision is lost, but most of the time `tiles3d_tileset_time` will be some random float. Maybe this note is not worth including as it is.

pjcozzi: It's needs some context.  As is, readers will see it as too scary to use or just ignore the warning.

lilleyse:
I removed the implementation note as I don't think it's needed. Ready.

Issue #25, opened 2015-10-02
Vector Tile Roadmap


pjcozzi:
Notes on [vector data](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/TileFormats/VectorData/README.md)

* [ ] Polygons, #24
* [ ] Polylines
* [ ] Points
   * Batch Table semantics for labels and billboards?  Declutter?

Selection of Cesium [geometries](http://cesiumjs.org/2013/11/04/Geometry-and-Appearances/):

* [ ] Polygon: extrusion and at height
* [ ] Polyline: extrusion (wall) and at height
* [ ] Rectangle: on terrain, extrusion, and at height 
* [ ] Ellipse: on terrain, extrusion, and at height 
* Others as requested

pjcozzi:
Initial Cesium implementation for polygons: https://github.com/AnalyticalGraphicsInc/cesium/pull/4186

pjcozzi:
Started spec in #124.

Pushing post 1.0 to narrow the scope, this would come soon after in a follow-up spec release since the spec and Cesium implementation are far along.

pjcozzi:
* [ ] In declarative styling, should `point.image` be able to reference pixels glTF buffer/bufferView style?

pjcozzi:
* [ ] Ability to change polyline width (and other polyline properties) via declarative styling

pjcozzi:
* [ ] Polyline style

pjcozzi:
* [ ] What classification volumes - and other higher order surfaces like ellipses - should be able to be just rendered like polygons - as opposed to used for classifying another tileset?

ggetz:
Suggestion from the forum: https://groups.google.com/forum/#!topic/cesium-dev/3zTt9rZVX08

Ability to style what happens when geometries intersect or overlap (for instance, when two points overlap, ability to define the resulting overlap color)

lilleyse:
* [ ] Make polyline heights optional. When heights are omitted the polyline should be clamped to the surface by the client.

lilleyse:
* [ ] Support textured polygons.

pjcozzi:
## Related

* [flatgeobuf](https://github.com/bjornharrtell/flatgeobuf) - "A performant binary encoding for geographic data based on flatbuffers"

jailln:
Hello, 
Do you have an idea on when the vector tile format will be available in 3D Tiles ?
Thanks :)

[...]

PR #228, opened 2017-06-05
Vector Tiles

bagnell:
Updated vector tiles draft. This reflects the current implementation in the `vector-tiles` branch of Cesium. 

A few things we may want to add:
- [ ] Meter width lines (like corridors on terrain/3D tiles)
- [ ] Points with meter width on terrain/3D Tiles.
- [ ] TOC to spec, #229
- [ ] This is very Earth centric?  Is that OK for now?  What about general 3D scenes.


pjcozzi:
@bagnell please target this to a new branch: `3d-tiles-next`

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+
+TODO: `uint16` indices.
+
+The indices are a buffer of `uint32` values. The byte length is given by `indicesByteLength` in the header. Each count in `POLYGON_INDEX_COUNT` represents a contiguous section of the array that represents a triangulated polygon. 
+For example, let the first two polygons have 6 and 12 for their index counts. The first polygon has 6 indices starting at byte offset `0` and ending at byte offset `6 * byteSize - 1`.
+The second polygon has 12 indices starting at byte offset `6 * byteSize` and ending at `6 * byteSize + 12 * byteSize`.
+
+**Figure 2**: Example index buffer.
+
+![indices](figures/indices.jpg)
```
pjcozzi: The arrow from `POLYGON_INDEX_COUNT[1]` is not clear.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+
+The number of indices must be a multiple of three. Each consecutive list of three indices is a triangle that must be ordered counter-clockwise. Each index is from the start of the buffer, **NOT** from the offset of the first position of the polygon.
+
+### Positions
+
+The positions buffer contains up to three sub-buffers for the polygons, polylines and points. The positions are encoded according to the quantized-mesh-1.0 format[1].
```
pjcozzi: Use serial comma.  "polylines,"

pjcozzi: Just link to quantized-mesh-1.0 format in a Resources section.  I don't think it's a good intro here.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+
+The bounding volume for the tile must be a tile bounding region containing the north, south, east, and west bounds of the tile. The positions are represented by u, v, and height values that are quantized and delta encoded.
```
pjcozzi: Throughout, order lower-left (x,y) to upper-right (x,y).  west, south, east, north.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+
+| Field | Meaning |
+| --- | --- |
+| u | The horizontal coordinate of the vertex in the tile. When the u value is 0, the vertex is on the Western edge of the tile. Then the value is 32767, the vertex is on the Eastern edge of the tile. For other values, the vertex's longitude is a linear interpolation between the longitudes of the Western and Eastern edges of the tile. |
```
pjcozzi: Throughout this table, are you sure words like Western should be capitalized?  Use bold if they need focus.

pjcozzi: Use consistent wording: vertex vs. position.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+    
+    uBuffer[i] = u;
+    vBuffer[i] = v;
+    heightBuffer[i] = height;
+}
+```
+
+#### Polygon positions
+
+The first section of the positions buffer, from offset 0 to `polygonPositionsByteLength`, contains the polygon positions. Polygon positions only have u and v values. The u values are from offset `0` to `polygonPositionsByteLength / 2`. The v values are from offset `polygonPositionsByteLength /2` to `polygonPositionsByteLength`. The number of positions for each polygon is determined by the value of its `POLYGON_COUNT`.
```
pjcozzi: "polygonPositionsByteLength /2" -> "polygonPositionsByteLength / 2"

pjcozzi: For each of these three sections, perhaps replace/augment the first sentence with a small table showing the start byte length and end byte length.

pjcozzi:
Bump @bagnell

Styling/README.md
```diff
@@ -1309,6 +1309,20 @@ For example:
...
 **TODO : add note about GLSL implementations requires strict type comparisons among other things: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/140**
 
+#### Vector Data
+
+A [Vector Tile](../TileFormats/VectorData/README.md) is a collection of points, polygons, and polylines. All vector data can be styled similar to other features with `color` and `show`. Points have several other styling options:
+
+| Style | Description | Default |
+| --- | --- | --- |
+| pointSize | The size of the point in pixels. | 8.0 |
+| image | A URL to an image to be displayed instead of a point. | " " |
```
bagnell: I could be a data URI. 

Would another option for the future be to have an atlas as part of the payload and the image be a subregion of the atlas? That way implementations wouldn't have to rebuild an atlas.

pjcozzi: Atlas - exactly.  Can you submit an issue for this for future work?

bagnell: Submitted #239

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
 
-Could also name this a `Geometry` tile.
+Each point, poyline, and polygon is a _feature_ in the 3D Tiles specification language. 
+
+## Layout
+
+A tile is composed of two sections: a header immediately followed by a body.
+
+**Figure1**: Vector tile layout.
+![layout](figures/layout.jpg)
```
bagnell: The positions are stored separately because the polygon vertices need to be contiguous in the buffer in counter clockwise order. The indices are for the triangulated polygon.

Another reason is polylines and points won't need indices this way.

>Is this because they are all mapped to different Cesium primitives at runtime?

This used to be the case. The buffers were sent directly to the GPU and rendered, but that isn't the case anymore.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+
+## Header
+
+The 28-byte header contains the following fields:
+
+| Field name | Data type | Description |
+| --- | --- | --- |
+| `magic` | 4-byte ANSI string | `"vctr"`. This can be used to identify the arraybuffer as a Vector tile. |
+| `version` | `uint32` | The version of the Vector Data format. It is currently `1`. |
```
bagnell: I don't know what the practice is here. @lilleyse, what did you do for the other tile formats?

lilleyse: They are all `1`.

TileFormats/VectorData/README.md
pjcozzi: How is this going to look when we scale to other geometries that need to be backed by buffer data as opposed to just a few parameters (like a sphere or cylinder).

Doesn't this seem too specific compared to something like glTF?  Why couldn't a glTF-like approach still allow polygon vertices to be contiguous and counter-clockwise?

bagnell: Right now, each subbuffer is delta +ZigZag encoded. We can grab the subbuffer, hand it to the webworker where it gets decoded.

If we make it more glTF-like, we could encode the whole buffer instead of each section for polylines, polygons and points. That would add a new step to decode the whole buffer client side. But, what about polygons which don't have heights?

When you quantize a glTF vertex, it doesn't depend on any other vertex. With delta + ZigZag encoding, each value depends on the previous value. So you encode the u, v, and height in separate buffers. Polygon vertices in counter-clockwise order will compress well. Likewise, you want polygons near each other to be consecutive in the buffer for better compression. For polygons, we can index into the u and v buffers no problem. Every other geometry type will have a height value, but how do we index into the height buffer? Fill the polygon heights with empty values? We could make it gzip compress well, but there still might be a lot of empty data when decompressed.

bagnell: Also, for general meshes, we would should prefer to quantize like glTF.

When I was experimenting with different formats, the current form compressed better. But since we want to add general meshes soon, is it better to switch back to the same quantization that glTF uses for consistency? Then we could make the buffer glTF-like as well.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,198 @@
...
+#### Global Semantics
+
+The semantics define global properties for all vector elements.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `POLYGONS_LENGTH` | `uint32` | The number of polygons in the tile. | :white_check_mark: Yes, unless one of `POLYLINES_LENGTH` or `POINTS_LENGTH` is defined. |
+| `POLYLINES_LENGTH` | `uint32` | The number of polylines in the tile. | :white_check_mark: Yes, unless one of `POLYGONS_LENGTH` or `POINTS_LENGTH` is defined.  |
+| `POINTS_LENGTH` | `uint32` | The number of points in the tile. | :white_check_mark: Yes, unless one of `POLYGONS_LENGTH` or `POLYLINES_LENGTH` is defined.  |
+| `MINIMUM_HEIGHT` | `float32` | The minimum terrain height for this tiles' region in meters above the WGS84 ellipsoid. | :white_check_mark: Yes. |
```
bagnell: It's required for decoding height values.

bagnell:
@pjcozzi This is ready for another look.

TileFormats/VectorData/README.md
pjcozzi: We discussed offline, and are going to keep this simple and as is-ish for now and get some early feedback before considering some of the changes I suggested, which may start to make this too much like a general model format.

TileFormats/VectorData/README.md
pjcozzi: Gotcha.

Styling/README.md
```diff
@@ -51,6 +51,7 @@ Contents:
    * [Notes](#notes)
 * [Batch Table Hierarchy](#batch-table-hierarchy)
 * [Point Cloud](#point-cloud)
+* [Vector Data](#vector-data)
```
pjcozzi: Add yourself to the contributors list in this file.

pjcozzi:
Tweaked the styling spec in 13e14ca853655abf091dea5e0f521ae4c85db953.

Styling/README.md
```diff
@@ -1344,6 +1345,58 @@ For example:
...
+| --- | --- | --- | --- |
+| `show` | `Boolean` | Whether or not to show the feature. | `true` |
+| `color` | `vec4` | The color of the feature. | `color('#FFFFFF')` |
+
+The following style properties apply to point features in vector tiles:
+
+| Style | Type | Description | Default |
+| --- | --- | --- | --- |
+| `show` | `Boolean` | Whether or not to show the feature. | `true` |
+| `color` | `vec4` | The color of the feature. | `color('#FFFFFF')` |
```
pjcozzi: What does this mean?  The color multiple with the image and label?  How does this relate to `pointColor`?

Styling/README.md
```diff
@@ -1344,6 +1345,58 @@ For example:
...
+| `pointSize` | `Number` | The size of the point in pixels. `pointSize` is ignored when `image` is defined. | `8.0` |
+| `pointColor` | `vec4` | The color of the point. `pointColor` is ignored when `image` is defined. | `color('#FFFFFF')` |
+| `pointOutlineColor` | `vec4` | The color of the point outline. `pointOutlineColor` is ignored when `image` is defined. | `color('#000000')` |
+| `pointOutlineWidth` | `Number` | The width, in pixels, of the point outline. `pointOutlineWidth` is ignored when `image` is defined. | `0.0` |
+| `labelText` | `String` | The text to display for the point | `undefined` |
+| `labelStyle` | `Number` | The label style: **fill** (`0`), **outline** (`1`), or **both** (`2`). This is ignored when `labelText` is undefined. | `0` |
+| `labelOutlineColor` | `vec` | The color of the text outline. This is ignored when `labelText` is undefined. | `color('#FFFFFF')` |
+| `labelOutlineWidth` | `Number` | The width of the text outline. This is ignored when `labelText` is undefined. | `1.0` |
+| `font` | `String` | The font of the displayed text. This is ignored when `labelText` is undefined. | `"30px sans-serif"` |
```
pjcozzi: I thought I mentioned this before but the format of the font string needs to be referenced, e.g., like this spec does for CSS colors.

Styling/README.md
```diff
@@ -1344,6 +1345,58 @@ For example:
...
+| `distanceDisplayCondition` | `vec2` | Sets the condition specifying at what distance from the camera that this feature will be displayed. The `x` value is the smallest distance in the interval where the feature is visible. The `y` value is the largest distance in the interval where the object is visible. If undefined, `distanceDisplayCondition` will be disabled. | `undefined` |
+| `heightOffset` | `Number` | The distance, in meters, to offset the height. This is ignored if the valuse is less than or equal to zero. | `0.0` |
+| `anchorLineEnabled` | `Boolean` | Whether or not to display a line from the feature to the point on terrain. This is ignored unless `heightOffset` is greater than zero. | `false` |
+| `anchorLineColor` | `vec4` | The color of the line from the feature to terrain. This is ignored unless `heightOffset` is greater than 0.0. | `color('#FFFFFF')` |
+| `image` | `String` | A URL to an image or a data URI to be displayed instead of a point. | `undefined` |
+
+For example:
+
+```json
```
pjcozzi: I thought I mentioned this before, but please add more examples showing more of the vector tile-specific properties.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+| `batchTableJSONByteLength` | `uint32` | The length of the batch table JSON section in bytes. Zero indicates that there is no batch table. |
+| `batchTableBinaryByteLength` | `uint32` | The length of the batch table binary section in bytes. If `batchTableJSONByteLength` is zero, this will also be zero. | 
+| `polygonIndicesByteLength` | `uint32` | The length of the polygon indices buffer. |
+| `polygonPositionsByteLength` | `uint32` | The length of the polygon positions buffer. |
+| `polylinePositionsByteLength` | `uint32` | The length of the polyline positions buffer. |
+| `pointPositionsByteLength` | `uint32` | The length of the point positions buffer. |
+
+If `featureTableJSONByteLength` equals zero, the tile does not need to be rendered.
+
+The body section immediately follows the header section, and is composed of four fields: `Feature Table`, `Batch Table`, `Indices`, and `Positions`.
```
pjcozzi: Here - and/or maybe in the ** Positions** section below - we should explicitly state the order and how to compute the offsets for the indices and positions for each type of feature.  We did to be **explicit** as opposed to relying on the order of the fields in the header and making an assumption.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+
+If a semantic has a dependency on another semantic, that semantic must be defined as well.
+
+#### Global Semantics
+
+The semantics define global properties for all vector elements.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `POLYGONS_LENGTH` | `uint32` | The number of pre-triangulated polygons in the tile. | :white_check_mark: Yes, unless one of `POLYLINES_LENGTH` or `POINTS_LENGTH` is defined. |
```
pjcozzi: Sounds like we require that at least one polygon, polyline, or point is defined.  Should this be required?  Does this mirror b3dm, i3dm, and pnts?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+| `POLYLINES_LENGTH` | `uint32` | The number of polylines in the tile. | :white_check_mark: Yes, unless one of `POLYGONS_LENGTH` or `POINTS_LENGTH` is defined.  |
+| `POINTS_LENGTH` | `uint32` | The number of points in the tile. | :white_check_mark: Yes, unless one of `POLYGONS_LENGTH` or `POLYLINES_LENGTH` is defined.  |
+| `MINIMUM_HEIGHT` | `float32` | The minimum height for this tiles' region in meters above the WGS84 ellipsoid. | :white_check_mark: Yes. |
+| `MAXIMUM_HEIGHT` | `float32` | The maximum height for this tiles' region in meters above the WGS84 ellipsoid. | :white_check_mark: Yes. |
+
+#### Vector Semantics
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `POLYGON_COUNTS` | `uint32[]` | The number of vertices that belong to each polygon. This refers to the polygon section of the positions buffer in the body. Each polygon count refers to a contiguous number of vertices in the position buffer that represents the polygon.  | :white_check_mark: Yes, unless `POLYGONS_LENGTH` is zero or not defined. |
```
pjcozzi: Here and similar below, this is probably better stated as

> Yes, when `POLYGONS_LENGTH` is greater than zero.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+| `POLYGON_INDEX_COUNTS` | `uint32[]` | The number of indices that belong to each polygon. This refers to the indices buffer of the body. Each index count refers to a contiguous number of indices that represent the triangulated polygon. | :white_check_mark: Yes, unless `POLYGONS_LENGTH` is zero or not defined. |
+| `POLYGON_MINIMUM_HEIGHTS` | `float32[]` | The minimum height of each polygon in meters above the WGS84 ellipsoid. | :red_circle: No. If the minimum height for each polygon is not specified, the global `MINIMUM_HEIGHT` will be used. |
+| `POLYGON_MAXIMUM_HEIGHTS` | `float32[]` | The maximum height of each polygon in meters above the WGS84 ellipsoid. | :red_cricle: No. If the maximum height for each polygon is not specified, the global `MAXIMUM_HEIGHT` will be used. |
```
pjcozzi: Should be `: red_circle:`

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+| `POLYLINE_BATCH_IDS` | `uint16[]` | The `batchId` of the polyline that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `POINT_BATCH_IDS` | `uint16[]` | The `batchId` of the point that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+
+## Batch Table
+
+The _Batch Table_ contains application-specific metadata, indexable by `batchId`, that can be used for declarative styling and application-specific use cases such as populating a UI or issuing a REST API request.
+
+See the [Batch Table](../BatchTable/README.md) reference for more information.
+
+### Indices
```
pjcozzi: Rename to **Polygon Indices**

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,210 @@
...
+    
+    uBuffer[i] = u;
+    vBuffer[i] = v;
+    heightBuffer[i] = height;
+}
+```
+
+#### Polygon positions
+
+**Figure 3**: Polygon layout.
```
pjcozzi: Nice diagrams!

bagnell:
@pjcozzi This is ready for another look.

pjcozzi:
Very nice @bagnell!!!

@lilleyse did you want to take a first pass over this?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+Each `u, v, height` triple is a single point.
+
+### Meshes
+
+**Figure 6**: Mesh layout
+
+![](figures/mesh.jpg)
+
+The indices are a buffer of `uint32` values. The indices for each mesh must be contiguous in the buffer. For example, the first mesh would have indices starting at the byte offset given by the first element in `MESH_INDEX_OFFSETS` and end after the first element in `MESH_INDEX_COUNTS`. After the indices is a `float32[]` containing positions of tall of the meshes. The number of positions in the buffer is given by `MESH_POSITION_COUNT`. Unlike polygons, there are no restrictions on where a mesh can index into the position buffer.
```
lilleyse: `tall` -> `all`

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+## Implementation Notes
+
+In JavaScript, a `TypedArray` cannot be created on data unless it is byte-aligned to the data type.
+For example, a `Float32Array` must be stored in memory such that its data begins on a byte multiple of four since each `float` contains four bytes.
+
+The string generated from the JSON and each array should be padded is necessary to ensure byte alignment.
+ 
+## Resources
+1. [quantized-mesh-1.0 terrain format](https://cesiumjs.org/data-and-assets/terrain/formats/quantized-mesh-1.0.html)
```
lilleyse: I may have missed it, where is this referenced from?

pjcozzi:
Please bump to me when ready.

pjcozzi:
@bagnell is the tasklist at the top of this up-to-date?

pjcozzi:
Update the vector tile link and description in the main README.md - 

image

pjcozzi:
@bagnell can you update the tasklist in #25?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| Field name | Data type | Description |
+| --- | --- | --- |
+| `magic` | 4-byte ANSI string | `"vctr"`. This can be used to identify the arraybuffer as a Vector tile. |
+| `version` | `uint32` | The version of the Vector Data format. It is currently `1`. |
+| `byteLength` | `uint32` | The length of the entire tile, including the header, in bytes. |
+| `featureTableJSONByteLength` | `uint32` | The length of the feature table JSON section in bytes. |
+| `featureTableBinaryByteLength` | `uint32` | The length of the feature table binary section in bytes. If `featureTableJSONByteLength` is zero, this will also be zero. |
+| `batchTableJSONByteLength` | `uint32` | The length of the batch table JSON section in bytes. Zero indicates that there is no batch table. |
+| `batchTableBinaryByteLength` | `uint32` | The length of the batch table binary section in bytes. If `batchTableJSONByteLength` is zero, this will also be zero. | 
+| `polygonIndicesByteLength` | `uint32` | The length of the polygon indices buffer. |
```
pjcozzi: Here and below "in bytes."

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `polygonPositionsByteLength` | `uint32` | The length of the polygon positions buffer. |
+| `polylinePositionsByteLength` | `uint32` | The length of the polyline positions buffer. |
+| `pointPositionsByteLength` | `uint32` | The length of the point positions buffer. |
+
+If `featureTableJSONByteLength` equals zero, the tile does not need to be rendered.
+
+The body section immediately follows the header section, and is composed, in order, of four fields: `Feature Table`, `Batch Table`, `Indices`, `Positions`, and `Meshes`.
```
pjcozzi: This is **five** fields now.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+Contains values for `vctr` semantics used to render features.  The general layout of a Feature Table is described in the [Feature Table specification](../FeatureTable).
+
+The `vctr` Feature Table JSON schema is defined in [vctr.featureTable.schema.json](../../schema/vctr.featureTable.schema.json).
+
+### Semantics
+
+Per-feature semantics specific to a feature type are prefixed with the name of the feature type. e.g. `POLYGON` for pre-triangulated polygons, `POLYLINE` for polylines and `POINT` for points.
+
+At least one global `LENGTH` semantic must be defined. 
+If `POLYGONS_LENGTH` is not defined, or zero, no polygons will be rendered.
```
pjcozzi: Make each of these "If POLYGONS_LENGTH is not defined..." statements a bullet; otherwise, the "formatted" paragraph is hard to read.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+If a semantic has a dependency on another semantic, that semantic must be defined as well.
+
+#### Global Semantics
+
+The semantics define global properties for all vector elements.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `POLYGONS_LENGTH` | `uint32` | The number of pre-triangulated polygons in the tile. | :red_circle: No. |
```
pjcozzi: Here and throughout I think the cleanest thing to do is to not have the period for simple "Yes" and "No" and only have it when it is a full sentence.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `POLYLINES_LENGTH` | `uint32` | The number of polylines in the tile. | :red_circle: No. |
+| `POINTS_LENGTH` | `uint32` | The number of points in the tile. | :red_circle: No. |
+| `MESHES_LENGTH` | `uint32` | The number of meshes in the tile. | :red_circle: No. |
+| `BOXES_LENGTH` | `uint32` | The number of boxes in the tile. | :red_circle: No. |
+| `CYLINDERS_LENGTH` | `uint32` | The number of cylinders in the tile. | :red_circle: No. |
+| `ELLIPSOIDS_LENGTH` | `uint32` | The number of ellipsoids in the tile. | :red_circle: No. |
+| `SPHERES_LENGTH` | `uint32` | The number of spheres in the tile. | :red_circle: No. |
+| `MINIMUM_HEIGHT` | `float32` | The minimum height for this tiles' region in meters above the WGS84 ellipsoid. | :white_check_mark: Yes. |
```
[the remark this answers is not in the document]
pjcozzi: Should min and max height be required?  These are just for potential rendering optimizations, right?  I think it is fine to make them optional.

Perhaps that will also play better if/when this tile is used for local coordinate systems.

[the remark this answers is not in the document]
pjcozzi: Should here and below have a `_ABOVE_WGS84` suffix?  They would not in Cesium, but in 3D Tiles, we don't really use "height."

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `MAXIMUM_HEIGHT` | `float32` | The maximum height for this tiles' region in meters above the WGS84 ellipsoid. | :white_check_mark: Yes. |
+| `RECTANGLE` | `float32[]` | The rectangle containing the geometry in the tile. It is an array with the four elements west, south, east, and north (in that order). | :white_check_mark: Yes. |
```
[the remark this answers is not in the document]
pjcozzi: Agreed this needs to be precisely defined.

For consistency with `boundingVolume.region`, can we just remove min/max properties above, rename this to `region`, and make this optional?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `POLYGON_INDEX_COUNTS` | `uint32[]` | The number of indices that belong to each polygon. This refers to the indices buffer of the body. Each index count refers to a contiguous number of indices that represent the triangulated polygon. | :white_check_mark: Yes, when the global `POLYGONS_LENGTH` is greater than zero. |
+| `POLYGON_MINIMUM_HEIGHTS` | `float32[]` | The minimum height of each polygon in meters above the WGS84 ellipsoid. | :red_circle: No. If the minimum height for each polygon is not specified, the global `MINIMUM_HEIGHT` will be used. |
+| `POLYGON_MAXIMUM_HEIGHTS` | `float32[]` | The maximum height of each polygon in meters above the WGS84 ellipsoid. | :red_circle: No. If the maximum height for each polygon is not specified, the global `MAXIMUM_HEIGHT` will be used. |
+| `POLYGON_BATCH_IDS` | `uint16[]` | The `batchId` of the polygon that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
```
pjcozzi: Are these often in increasing order?  It is worth having a fast path where if these are not defined, it is assume to be auto-incrementing from zero with a defined order of each feature type?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `POINT_BATCH_IDS` | `uint16[]` | The `batchId` of the point that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `MESH_INDEX_OFFSETS` | `uint32[]` | The offset of the first index in the indices buffer of each mesh. | :white_check_mark: Yes, when the global `MESHES_LENGTH` is greater than zero. |
+| `MESH_INDEX_COUNTS` | `uint32[]` | The number of indices for each mesh. | :white_check_mark: Yes, when the global `MESHES_LENGTH` is greater than zero. |
+| `MESH_POSITION_COUNT` | `uint32` | The number of positions of all meshes. | :white_check_mark: Yes, when the global `MESHES_LENGTH` is greater than zero. |
+| `MESH_BATCH_IDS` | `uint16[]` | The `batchId` of the mesh that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `BOXES` | `float32[]` | The boxes in the tile. The length of the array will be `19 * BOXES_LENGTH`. The first three elements of each box are the x, y, and z dimensions. The following 16 elements are the model matrix of the box. | :white_check_mark: Yes, when the global `BOXES_LENGTH` is greater than zero. |
```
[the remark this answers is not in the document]
pjcozzi: Also, is this the best format for instanced rendering?  Or should these be separated to avoid runtime splicing?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+**Figure 2**: Example index buffer.
+
+![](figures/indices.jpg)
+
+The number of indices must be a multiple of three. Each consecutive list of three indices is a triangle that must be ordered counter-clockwise. Each index is from the start of the buffer, **NOT** from the offset of the first position of the polygon.
+
+### Positions
+
+The positions buffer contains up to three sub-buffers for the polygons, polylines, and points.
```
pjcozzi: Shouldn't this apply to all positions including those for classification volumes?

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
pjcozzi: Same comment for the other volumes below.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `CYLINDER_BATCH_IDS` | `uint16[]` | The `batchId` of the cylinder that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `ELLIPSOIDS` | `float32[]` | The ellipsoids in the tile. The length of the array will be `19 * ELLIPSOIDS_LENGTH`. The first three elements are the radii in the x, y, and z axes. The following 16 elements are the model matrix. | :white_check_mark: Yes, when the global `ELLIPSOIDS_LENGTH` is greater than zero. |
+| `ELLIPSOID_BATCH_IDS` | `uint16[]` | The `batchId` of the ellipsoid that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `SPHERES` | `float32[]` | The spheres in the tile. The length of the array will be `17 * SPHERES_LENGTH`. The first element is the radius. The following 16 elements are the model matrix. | :white_check_mark: Yes, when the global `SPHERES_LENGTH` is greater than zero. |
```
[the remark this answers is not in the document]
pjcozzi: Yes, but spheres are more concise - probably a worthwhile fast path.

However, spheres do not need the entire 4x4 matrix - they only need the translation - the rotation doesn't matter and the scale can be baked into the radius.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+## Batch Table
+
+The _Batch Table_ contains application-specific metadata, indexable by `batchId`, that can be used for declarative styling and application-specific use cases such as populating a UI or issuing a REST API request.
+
+See the [Batch Table](../BatchTable/README.md) reference for more information.
+
+### Polygon Indices
+
+TODO: `uint16` indices?
```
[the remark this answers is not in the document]
pjcozzi: Does it have a big impact on file size?  Is this consistent with b3dm?

`uint32` implies using the WebGL extension (which is probably fine) or a lot of runtime slicing.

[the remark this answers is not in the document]
pjcozzi: > Does it have a big impact on file size? 

After gzip...

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| `SPHERES` | `float32[]` | The spheres in the tile. The length of the array will be `17 * SPHERES_LENGTH`. The first element is the radius. The following 16 elements are the model matrix. | :white_check_mark: Yes, when the global `SPHERES_LENGTH` is greater than zero. |
+| `SPHERE_BATCH_IDS` | `uint16[]` | The `batchId` of the sphere that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+
+## Batch Table
+
+The _Batch Table_ contains application-specific metadata, indexable by `batchId`, that can be used for declarative styling and application-specific use cases such as populating a UI or issuing a REST API request.
+
+See the [Batch Table](../BatchTable/README.md) reference for more information.
+
+### Polygon Indices
```
pjcozzi: Also for meshes?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+
+### Positions
+
+The positions buffer contains up to three sub-buffers for the polygons, polylines, and points.
+
+The bounding volume for the tile must be a tile bounding region containing the west, south, east, north bounds of the tile. The positions are represented by u, v, and height values that are quantized and delta encoded.
+
+| Field | Meaning |
+| --- | --- |
+| u | The horizontal coordinate of the vertex in the tile. When the u value is 0, the vertex is on the western edge of the tile. Then the value is 32767, the vertex is on the eastern edge of the tile. For other values, the vertex's longitude is a linear interpolation between the longitudes of the western and eastern edges of the tile. |
```
pjcozzi: With just a few wording tweaks and maybe a property tweak or two - couldn't we word this so that it is not geo specific?  For example, I suspect the code doesn't care that something is "west" or "east", it is just "min" and "max."

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
+| v | The vertical coordinate of the vertex in the tile. When the v value is 0, the vertex is on the southern edge of the tile. When the value is 32767, the vertex is on the northern edge of the tile. For other values, the vertex's latitude is a linear interpolation between the latitudes of the southern and northern edges of the tile. |
+| height | The height of the vertex of the tile. When the height value is 0, the vertex's height is equal to `MINIMUM_HEIGHT` from the feature table. When the value is 32767, the vertex's height is equal to `MAXIMUM_HEIGHT` from the feature table. For other values, the vertex's height is a linear interpolation of the minimum and maximum heights. |
+
+The values are then delta and ZigZag encoded. The delta encoding ensures the values are small integers. The ZigZag encoding ensure the values are positive integers. Example encoding code is listed below:
```
pjcozzi: "The ZigZag encoding ensure" -> "ensures"

Styling/README.md
```diff
@@ -1344,6 +1347,65 @@ For example:
...
+| Style | Type | Description | Default |
+| --- | --- | --- | --- |
+| `show` | `Boolean` | Whether or not to show the feature. | `true` |
+| `color` | `vec4` | The color of the feature. | `color('#FFFFFF')` |
+
+The following style properties apply to point features in vector tiles:
+
+| Style | Type | Description | Default |
+| --- | --- | --- | --- |
+| `show` | `Boolean` | Whether or not to show the feature. | `true` |
```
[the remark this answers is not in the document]
bagnell: I added `pointColor` and `labelColor` so that they can have different colors from the polygond/polylines. Should they have the same color? Should the only way to make them different from using a style that uses the metadata from the batch table?

Styling/README.md
```diff
@@ -1344,6 +1347,65 @@ For example:
...
+    "labelStyle" : "2",
+    "image" : "'http://example.com/url/to/image.jpg'",
+    "origin" : "1",
+    "labelOrigin" : "-1"
+}
+```
+
+| Fill | Outline | Fill and Outline |
+| --- | --- | --- |
+| ![](figures/label_fill.jpg) | ![](figures/label_outline.jpg) | ![](figures/label_fill_outline.jpg) |
```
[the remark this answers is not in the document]
bagnell: FXAA was disabled. I'll see what they look like when drawing the text to a canvas.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,248 @@
...
-* Concise representations for Cesium's [set of geometries](https://cesiumjs.org/tutorials/Geometry-and-Appearances/), including extrusions, and billboards and labels.
-   * Need to carefully select the representation for the best trade-off between conciseness and runtime processing.  For example, polygons will likely be pre-triangulated since it only adds indices to the payload, but will be subdivided at runtime since subdivision is fast and increases the vertex payload significantly.
-   * RTC positions for high precision-rendering.
-   * Context-aware compression.
-   * Bounding volume may need to be adjusted at runtime for terrain clamping.
-* Metadata for cracking, morphing, and perhaps label declutter.
+The _Vector_ tile format allows streaming of vector datasets including points, polylines, and polygons.  Points can be represented with a combination of billboard, label, and point graphics primitives.
 
-Could also name this a `Geometry` tile.
+Each point, poyline, and polygon is a _feature_ in the 3D Tiles specification language.
```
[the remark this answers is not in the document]
bagnell: The other feature types will be in the geometry tile (except the mesh which was removed).

Should I mention the classification use case in the B3DM spec or is that an implementation detail?

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
bagnell: The min and max heights are required over the entire tile because they are used to interpolate the height values for points and polylines.

Is the suffix needed if it is defined to be above the ellipsoid in the description?

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
bagnell: I can rename it to region, but it would still be required. It is used to interpolate the positions.

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
bagnell: >Why isn't this from 0 to 65525? Is it because of the delta and ZigZag encoding?

Yes, see the quantized mesh 1.0 spec.

TileFormats/VectorData/README.md
bagnell: It isn't referenced anywhere. The spec for the position layout is similar. @pjcozzi, should I remove this?

Styling/README.md
```diff
@@ -1344,6 +1347,65 @@ For example:
...
+| `font` | `String` | The font of the displayed text. This is ignored when `labelText` is undefined. | `"30px sans-serif"` |
+| `backgroundColor` | `vec4` | The label background color. This is ignored when `labelText` is undefined. | `rgba(42, 42, 42, 0.8)` |
+| `backgroundPadding` | `vec2` | The background padding, in pixels, of this label.  The `x` value controls horizontal padding, and the `y` value controls vertical padding. This is ignored when `labelText` is undefined. | `vec2(7, 5)` |
+| `backgroundEnabled` | `Boolean` | Whether or not to display the label background. This is ignored when `labelText` is undefined. | `false` |
+| `scaleByDistance` | `vec4` | Sets near and far scaling properties of a feature based on the features's distance from the camera. A feature's scale will interpolate between the `y` and `w` values while the camera distance falls within the upper and lower bounds of the specified `x` and `z` value. Outside of these ranges, the features's scale remains clamped to the nearest bound.  If undefined, `scaleByDistance` will be disabled. | `undefined` |
```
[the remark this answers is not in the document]
bagnell: @oterral added `translucencyByDistance` and `distanceDisplayCondition` to the implementation in Cesium (maybe even scaleByDistance).

I can imagine having a root tile over a country with labels for the major cities. You could have some priority in the style where some of the cities are always visible and some fade in as the camera gets closer to reduce cluttering. You could get that effect with better tiling but the labels would pop in as opposed to fading in.

Styling/README.md
```diff
@@ -1344,6 +1347,65 @@ For example:
...
+| `translucencyByDistance` | `vec4` | Sets near and far translucency properties of a feature based on the feature's distance from the camera. A feature's translucency will interpolate between the `y` and `w` while the camera distance falls within the upper and lower bounds of the specified `x` and `z`. Outside of these ranges the feature's translucency remains clamped to the nearest bound.  If undefined, `translucencyByDistance` will be disabled. | `undefined` |
```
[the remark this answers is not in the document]
bagnell: I'm OK with either translucency or alpha.

bagnell:
@pjcozzi @lilleyse This is ready for another look. I'll update the implementation for the few minor changes after you have another pass. The geometry and mesh were removed. Geometry tiles will be a separate PR.

Styling/README.md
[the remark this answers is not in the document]
pjcozzi: Maybe I am confused; what does `color` do when applied to a point feature?  Is it valid to use?  Or must `pointColor` be set instead?

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
pjcozzi: Mention the classification use case just like, for example, the top of the i3dm tile spec gives examples for how i3dm will be used.  Classification is a use case but is not required by the spec.

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
pjcozzi: Maybe not.  How can we generalize this so it is not specific to WGS84, e.g., if we just wanted to use it for a BIM model?  Should there be a tile-level enum that says WGS84 or a local coordinate?  Seems like this is becoming a common enough case.

TileFormats/VectorData/README.md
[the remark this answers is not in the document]
pjcozzi: OK.

Same question about local coordinates.

TileFormats/VectorData/README.md
pjcozzi: Up to you.

Styling/README.md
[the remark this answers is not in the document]
pjcozzi: Sounds good, but please still document interpolation. https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/228#discussion_r153024762

Styling/README.md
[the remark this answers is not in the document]
pjcozzi: OK, then up to @lilleyse I suppose.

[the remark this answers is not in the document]
lilleyse: Alpha is fine with me.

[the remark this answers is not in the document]
pjcozzi: OK, captain obvious, but please make it 0 - transparent, 1 - opaque, and then map it to Cesium's translucency.

Styling/README.md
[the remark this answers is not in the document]
bagnell: `color` isn't applied to a point. Its applied to polygons and polylines. `pointColor` is applied to points only.

bagnell:
Instead of a separate PR, I added the geometry spec to this.

Styling/README.md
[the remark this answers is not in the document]
pjcozzi: OK, so question still is can't `color` apply to a point?  So that point features "inherit" the standard properties instead of ignoring some and adding new ones?

pjcozzi:
> Instead of a separate PR, I added the geometry spec to this.

Add a link to it from the main README.md.

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+| `featureTableJSONByteLength` | `uint32` | The length of the feature table JSON section in bytes. |
+| `featureTableBinaryByteLength` | `uint32` | The length of the feature table binary section in bytes. If `featureTableJSONByteLength` is zero, this will also be zero. |
+| `batchTableJSONByteLength` | `uint32` | The length of the batch table JSON section in bytes. Zero indicates that there is no batch table. |
+| `batchTableBinaryByteLength` | `uint32` | The length of the batch table binary section in bytes. If `batchTableJSONByteLength` is zero, this will also be zero. | 
+
+If `featureTableJSONByteLength` equals zero, the tile does not need to be rendered.
+
+The body section immediately follows the header section, and is composed, in order, of two fields: `Feature Table` and  `Batch Table`.
+
+Code for reading the header can be found in
```
pjcozzi: Throughout just like to Cesium's master branch - later this can be a tag so it never breaks.  The PR will be merged _soon_.

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Feature Table
+
+Contains values for `geom` semantics used to render features.  The general layout of a Feature Table is described in the [Feature Table specification](../FeatureTable).
+
+The `geom` Feature Table JSON schema is defined in [geom.featureTable.schema.json](../../schema/geom.featureTable.schema.json).
+
+### Semantics
+
+Per-feature semantics specific to a feature type are prefixed with the name of the feature type. e.g. `BOXES` for boxes, `CYLINDERS` for cylinders and `ELLIPSOIDS` for ellipsoids.
```
pjcozzi: We use the "serial comma" - put a comma after "cylinders."

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+At least one global `LENGTH` semantic must be defined. 
+If `BOXES_LENGTH` in not defined, or zero, no boxes will be rendered.
```
pjcozzi: Perhaps make each of this "If" statements, a bullet.

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+If `SPHERES_LENGTH` in not defined, or zero, no spheres will be rendered.
+Multiple feature types may be defined in a single Geometry tile using multiple `LENGTH` semantics, and in that case, all specified feature types will be rendered.
+
+If a semantic has a dependency on another semantic, that semantic must be defined as well.
+
+#### Geometry Semantics
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `BOXES` | `float32[]` | The boxes in the tile. The length of the array will be `19 * BOXES_LENGTH`. The first three elements of each box are the x, y, and z dimensions. The following 16 elements are the 4x4 model matrix in column-major order. | :white_check_mark: Yes, when the global `BOXES_LENGTH` is greater than zero. |
```
pjcozzi: Let's be super explicit - are the dimensions the full length or half length?

pjcozzi: Do we even need this since it could just be the translation part of the 4x4 matrix?  Just all boxes are unit boxes transformed by the matrix?

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+| `BOX_BATCH_IDS` | `uint16[]` | The `batchId` of the box that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `CYLINDERS` | `float32[]` | The cylinders in the tile. The length of the array will be `18 * CYLINDERS_LENGTH`. The first element is the radius of the cylinder. The second element is the length along the z axis. The following 16 elements are the 4x4 model matrix in column-major order. | :white_check_mark: Yes, when the global `CYLINDERS_LENGTH` is greater than zero. |
```
pjcozzi: Similar comment for cylinders.

TileFormats/GeometryData/README.md
```diff
@@ -0,0 +1,108 @@
...
+| `CYLINDER_BATCH_IDS` | `uint16[]` | The `batchId` of the cylinder that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `ELLIPSOIDS` | `float32[]` | The ellipsoids in the tile. The length of the array will be `19 * ELLIPSOIDS_LENGTH`. The first three elements are the radii in the x, y, and z axes. The following 16 elements are the 4x4 model matrix in column-major order. | :white_check_mark: Yes, when the global `ELLIPSOIDS_LENGTH` is greater than zero. |
```
pjcozzi: Same comment.

TileFormats/GeometryData/README.md
bagnell: I'm OK with just a model matrix, that's how the implementation works.

bagnell: Should it mention anything about scaling the unit geometry? The only one I thought worth mentioning was the cylinder for which axes scaled the radius and which one was the length.

Styling/README.md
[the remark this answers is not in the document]
bagnell: Updated.

bagnell:
@pjcozzi @lilleyse This is ready for another look.

TileFormats/GeometryData/README.md
pjcozzi: +1 for the cylinder axes.  I don't think anything else is needed for the box.

pjcozzi:
@lilleyse this is all you when you have time.

@bagnell @lilleyse 
1. Please make sure that everything discussed here - but not currently done - is in an issue/roadmap
2. That we strive to make this as non-Earth centric as possible either now or in the roadmap.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
 
 **TODO : add note about GLSL implementations requires strict type comparisons among other things: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/140**
 
+## Vector Data
+
+A [Vector Tile](../TileFormats/VectorData/README.md) is a collection of vector features such as points, polygons, and polylines. Vector features can be styled similar to other features with `color` and `show`. Points have several other styling options:
+
+The following style properties apply to geometry features in vector tiles:
```
lilleyse: Probably clearer to say `polygons` and `polylines` here.

Styling/schema/vctr.style.schema.json
```diff
@@ -0,0 +1,243 @@
...
+      "pointSize" : {
+        "oneOf" : [{
+          "$ref" : "style.numberExpression.schema.json"
+        }, {
+          "$ref" : "style.condition.schema.json"
+        }],
+        "description" : "The size of points in pixels.",
+        "default" : 8.0
+      },
+      "pointColor" : {
```
lilleyse: This was removed.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `show` | `Boolean` | Whether or not to show the feature. | `true` |
+| `color` | `vec4` | The color of the point. `color` is ignored when `image` is defined. | `color('#FFFFFF')` |
+| `pointSize` | `Number` | The size of the point in pixels. `pointSize` is ignored when `image` is defined. | `8.0` |
+| `pointOutlineColor` | `vec4` | The color of the point outline. `pointOutlineColor` is ignored when `image` is defined. | `color('#000000')` |
+| `pointOutlineWidth` | `Number` | The width, in pixels, of the point outline. `pointOutlineWidth` is ignored when `image` is defined. | `0.0` |
+| `labelText` | `String` | The text to display for the point. `'\n'` can be added to separate lines of text. | `undefined` |
+| `labelColor` | `vec4` | The color of the label's text. | `color('#FFFFFF')` |
+| `labelStyle` | `Number` | The label style: **fill** (`0`), **outline** (`1`), or **both** (`2`). See **Table1**.| `0` |
+| `labelOutlineColor` | `vec4` | The color of the text outline. | `color('#FFFFFF')` |
+| `labelOutlineWidth` | `Number` | The width of the text outline. | `1.0` |
```
lilleyse: Include `in pixels` here and in the schema.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `font` | `String` | The font of the displayed text. | `"30px sans-serif"` |
```
lilleyse: Mention that this is a CSS font string, here and in the schema.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `backgroundColor` | `vec4` | The label background color. | `rgba(42, 42, 42, 0.8)` |
```
lilleyse: All the other color defaults are specified as css color strings, this should probably as well. (Also in schema).

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `backgroundPadding` | `vec2` | The background padding, in pixels, of this label.  The `x` value controls horizontal padding, and the `y` value controls vertical padding. The number of `x` pixels is added to both sides. The number of `y` pixels is added to the top and bottom. | `vec2(7, 5)` |
+| `backgroundEnabled` | `Boolean` | Whether or not to display the label background. | `false` |
+| `scaleByDistance` | `vec4` | Sets near and far scaling properties of a feature based on the features's distance from the camera. A feature's scale will interpolate between the `y` and `w` values while the camera distance falls within the lower, `x`, and upper, `z`, bounds. Outside of these ranges, the features's scale remains clamped to the nearest bound.  If undefined, `scaleByDistance` will be disabled. | `undefined` |
```
lilleyse: For `scaleByDistance`, `translucencyByDistance`, and `distanceDisplayCondition` say that they are in meters.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `heightOffset` | `Number` | The distance, in meters, to offset the height from the original height of the point. This is ignored if the value is less than or equal to zero. | `0.0` |
+| `anchorLineEnabled` | `Boolean` | Whether or not to display a line from the feature to the point's original position. This is ignored unless `heightOffset` is greater than zero. | `false` |
+| `anchorLineColor` | `vec4` | The color of the line from the feature to point's original position. This is ignored unless `heightOffset` is greater than 0.0. | `color('#FFFFFF')` |
+| `image` | `String` | A URL to an image or a data URI to be displayed instead of a point. | `undefined` |
+| `disableDepthTestDistance` | `Number` | The distance from the camera at which to disable the depth test to, for example, prevent clipping against terrain. When set to zero or `undefined`, the depth test is always applied. When set to Number.POSITIVE_INFINITY, the depth test is never applied. | `undefined` |
```
lilleyse: Include `in meters`.

Styling/README.md
```diff
@@ -1344,6 +1347,66 @@ For example:
...
+| `distanceDisplayCondition` | `vec2` | Sets the condition specifying at what distance from the camera that this feature will be displayed. The `x` value is the smallest distance in the interval where the feature is visible. The `y` value is the largest distance in the interval where the object is visible. If undefined, `distanceDisplayCondition` will be disabled. | `undefined` |
+| `heightOffset` | `Number` | The distance, in meters, to offset the height from the original height of the point. This is ignored if the value is less than or equal to zero. | `0.0` |
+| `anchorLineEnabled` | `Boolean` | Whether or not to display a line from the feature to the point's original position. This is ignored unless `heightOffset` is greater than zero. | `false` |
+| `anchorLineColor` | `vec4` | The color of the line from the feature to point's original position. This is ignored unless `heightOffset` is greater than 0.0. | `color('#FFFFFF')` |
```
lilleyse: `feature to the point's`

Styling/README.md
lilleyse: I haven't tested this, but `Number.POSITIVE_INFINITY` probably won't get parsed correctly by the styling language. We'll need to include this number in the [constants section](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/596d30eb327203f87f443615626ba2acb2a70e0e/Styling/README.md#constants)  and also added to Cesium.

Styling/schema/vctr.style.schema.json
```diff
@@ -0,0 +1,243 @@
...
+        "oneOf" : [{
+          "$ref" : "style.colorExpression.schema.json"
+        }, {
+          "$ref" : "style.condition.schema.json"
+        }],
+        "description" : "The color of the point.",
+        "default" : "Color('#FFFFFF')"
```
lilleyse: These all should be lowercase `color`

Styling/schema/vctr.style.schema.json
```diff
@@ -0,0 +1,243 @@
...
+        "default" : 0.0
+      },
+      "labelText" : {
+        "oneOf" : [{
+          "type" : "string"
+        }, {
+          "$ref" : "style.condition.schema.json"
+        }],
+        "description" : "The label text to display.",
+        "default" : "Undefined"
```
lilleyse: These should be lower case `undefined`

Styling/schema/vctr.style.schema.json
```diff
@@ -0,0 +1,243 @@
...
+        "default" : "false"
+      },
+      "scaleByDistance" : {
+        "oneOf" : [{
+          "$ref" : "style.vec4Expression.schema.json"
+        }, {
+          "$ref" : "style.condition.schema.json"
+        }],
+        "description" : "Sets near and far scaling properties of a feature based on the features's distance from the camera. A feature's scale will interpolate between the `y` and `w` values while the camera distance falls within the upper and lower bounds of the specified `x` and `z` value. Outside of these ranges, the features's scale remains clamped to the nearest bound.  If undefined, `scaleByDistance` will be disabled.",
+        "default" : "Undefined"
```
lilleyse: You'll have to check if it is valid for "Undefined" to be a default value for some of these whose type is not string.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Code for reading the header can be found in
+[Geometry3DTileContent.js](https://github.com/AnalyticalGraphicsInc/cesium/blob/master/Source/Scene/Geometry3DTileContent.js)
+in the Cesium implementation of 3D Tiles.
+
+## Feature Table
+
+Contains values for `geom` semantics used to render features.  The general layout of a Feature Table is described in the [Feature Table specification](../FeatureTable).
+
+The `geom` Feature Table JSON schema is defined in [geom.featureTable.schema.json](../../schema/geom.featureTable.schema.json).
```
lilleyse: `geom.featureTable.schema.json` does not exist.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Semantics
+
+Per-feature semantics specific to a feature type are prefixed with the name of the feature type. e.g. `BOXES` for boxes, `CYLINDERS` for cylinders, and `ELLIPSOIDS` for ellipsoids.
```
lilleyse: Also include spheres.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+At least one global `LENGTH` semantic must be defined. 
+* If `BOXES_LENGTH` in not defined, or zero, no boxes will be rendered.
+* If `CYLINDERS_LENGTH` in not defined, or zero, no cylinders will be rendered.
+* If `ELLIPSOIDS_LENGTH` in not defined, or zero, no ellipsoids will be rendered.
+* If `SPHERES_LENGTH` in not defined, or zero, no spheres will be rendered.
+Multiple feature types may be defined in a single Geometry tile using multiple `LENGTH` semantics, and in that case, all specified feature types will be rendered.
```
lilleyse: Add a newline before this. Right now it is part of the `SPHERES_LENGTH` bullet.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+If a semantic has a dependency on another semantic, that semantic must be defined as well.
+
+#### Geometry Semantics
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `BOXES` | `float32[]` | The boxes in the tile. The length of the array will be `16 * BOXES_LENGTH`. The 16 elements are the 4x4 model matrix in column-major order. | :white_check_mark: Yes, when the global `BOXES_LENGTH` is greater than zero. |
```
lilleyse: Should it be mentioned somewhere that only affine transformations are supported?

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+| `CYLINDER_BATCH_IDS` | `uint16[]` | The `batchId` of the cylinder that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `ELLIPSOIDS` | `float32[]` | The ellipsoids in the tile. The length of the array will be `16 * ELLIPSOIDS_LENGTH`. The 16 elements are the 4x4 model matrix in column-major order. | :white_check_mark: Yes, when the global `ELLIPSOIDS_LENGTH` is greater than zero. |
+| `ELLIPSOID_BATCH_IDS` | `uint16[]` | The `batchId` of the ellipsoid that can be used to retrieve metadata from the `Batch Table`. | :red_circle: No. |
+| `SPHERES` | `float32[]` | The spheres in the tile. The length of the array will be `3 * SPHERES_LENGTH`. The 3 elements are the translation of the sphere. | :white_check_mark: Yes, when the global `SPHERES_LENGTH` is greater than zero. |
```
lilleyse: Spheres should contain scale.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+The semantics define global properties for all geometry elements.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `BOXES_LENGTH` | `uint32` | The number of boxes in the tile. | :red_circle: No. |
+| `CYLINDERS_LENGTH` | `uint32` | The number of cylinders in the tile. | :red_circle: No. |
+| `ELLIPSOIDS_LENGTH` | `uint32` | The number of ellipsoids in the tile. | :red_circle: No. |
+| `SPHERES_LENGTH` | `uint32` | The number of spheres in the tile. | :red_circle: No. |
+| `RTC_CENTER` | `float32[]` | The center used for RTC rendering. | :red_circle: No. If no center is supplied, the center of the global `RECTANGLE` will be used. |
```
lilleyse: `RECTANGLE` is not included in `GEOM`.

TileFormats/Geometry/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+The _Geometry_ tile format allows streaming of geometry datasets including boxes, cylinders, ellipsoids, and spheres.
+
+Each box, cylinder, ellipsoid, and sphere is a _feature_ in the 3D Tiles specification language.
+
+In addition to rendering the geometry, the geometry can be used for classifying geometry of other 3D Tiles tilesets.
+
+## Layout
+
+A tile is composed of two sections: a header immediately followed by a body.
```
lilleyse: Geom should should have a layout diagram like the other formats.

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,213 @@
...
+
+The semantics define global properties for all vector elements.
+
+| Semantic | Data Type | Description | Required |
+| --- | --- | --- | --- |
+| `POLYGONS_LENGTH` | `uint32` | The number of pre-triangulated polygons in the tile. | :red_circle: No |
+| `POLYLINES_LENGTH` | `uint32` | The number of polylines in the tile. | :red_circle: No |
+| `POINTS_LENGTH` | `uint32` | The number of points in the tile. | :red_circle: No |
+| `REGION` | `float32[]` | An array of six numbers that define the bounding geographic region in WGS84 / EPSG:4326 coordinates with the order `[west, south, east, north, minimum height, maximum height]`. Longitudes and latitudes are in radians, and heights are in meters above (or below) the [WGS84 ellipsoid](http://earth-info.nga.mil/GandG/publications/tr8350.2/wgs84fin.pdf). | :white_check_mark: Yes |
+| `RTC_CENTER` | `float32[]` | The center used for RTC rendering. | :red_circle: No. If no center is supplied, the center of the global `RECTANGLE` will be used. |
```
lilleyse: Should this be `REGION`?

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,213 @@
...
+
+The number of positions for each polygon is determined by the value of its `POLYGON_COUNT`. For example, let the first polygon count be 5. The first polygon's u values start at offset `0` and end at `5 * sizeof(uint16)`. Its v values start at `polygonPositionsByteLength / 2` and end at `polygonPositionsByteLength / 2 + 5 * sizeof(uint16)`.
+
+The positions of the polygons must be the outer ring positions listed in counter-clockwise order.
+
+#### Polyline positions
+
+**Figure 4**: Polyline layout.
+
+![](figures/polyline.jpg)
```
lilleyse: In the diagram `polylinePositionsByteLength / 2` should be `polylinePositionsByteLength / 3`

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,213 @@
...
+
+The number of positions for each polyline is determined by the value of its `POLYLINE_COUNT`. From the first point on the polyline, each successive point creates a segment connected to the previous.
+
+#### Point positions
+
+**Figure 5**: Point layout.
+
+![](figures/point.jpg)
```
lilleyse: In the diagram `pointPositionsByteLength / 2` should be `pointPositionsByteLength / 3`

TileFormats/VectorData/README.md
```diff
@@ -1,19 +1,213 @@
...
+    vBuffer[i] = v;
+    heightBuffer[i] = height;
+}
+```
+
+#### Polygon positions
+
+**Figure 3**: Polygon layout.
+
+![](figures/polygon.jpg)
```
lilleyse: Should this diagram also include height values?

schema/vctr.featureTable.schema.json
```diff
@@ -0,0 +1,149 @@
...
+    "properties" : {
+      "MINIMUM_HEIGHT" : {
+        "$ref" : "featureTable.schema.json#/definitions/globalPropertyScalar"
+      },
+      "MAXIMUM_HEIGHT" : {
+        "$ref" : "featureTable.schema.json#/definitions/globalPropertyScalar"
+      },
+      "RECTANGLE" : {
+        "type" : "array",
+        "description" : "A floating-point 4 element array that defines a rectangle bounding all of the vector data. The elements in the array are, in order, west, south, east, and north.",
```
lilleyse: This description is out of date compared to the spec text.

schema/vctr.featureTable.schema.json
```diff
@@ -0,0 +1,149 @@
...
+      "POLYGONS_LENGTH" : {
+        "$ref" : "featureTable.schema.json#/definitions/globalPropertyScalar"
+      },
+      "POLYLINES_LENGTH" : {
+        "$ref" : "featureTable.schema.json#/definitions/globalPropertyScalar"
+      },
+      "POINTS_LENGTH" : {
+        "$ref" : "featureTable.schema.json#/definitions/globalPropertyScalar"
+      },
+      "MESHES_LENGTH" : {
```
lilleyse: Remove meshes, boxes, etc.

Styling/README.md
bagnell: This is mentioned here: https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/228/files#diff-406e6affb04e8aa11fb643fb5ed2fc9fR259

TileFormats/Geometry/README.md
bagnell: Isn't that implicit in the definition of `model matrix`? I could change it to `4x4 affine transformation matrix`.

TileFormats/Geometry/README.md
bagnell: `RECTANGLE` and `REGION` aren't required by `geom`.

TileFormats/VectorData/README.md
bagnell: Polygons don't have height values.The heights are either given by th epolygon minimum and maximum heightsd in the feature table or the global minimum and maximum for the tile. Only points and polylines need ber-vertex heights.

bagnell:
@lilleyse Updated. This is ready for another look.

TileFormats/Geometry/README.md
lilleyse: It might implicit, though  `4x4 affine transformation matrix` seems more exact.

TileFormats/Geometry/README.md
lilleyse: What I mean is, `RECTANGLE` should be removed from the `RTC_CENTER` description because it doesn't exist for the `geom` format.

TileFormats/Geometry/README.md
lilleyse: Bump

lilleyse: Oh actually the file exists but there is a typo in the name: `schema/geom.featureTable.scheme.json` should be `schema/geom.featureTable.schema.json`

Styling/schema/vctr.style.schema.json
```diff
@@ -0,0 +1,262 @@
...
+        "description" : "Whether or not to display the label background.",
+        "default" : "false"
+      },
+      "scaleByDistance" : {
+        "oneOf" : [{
+          "$ref" : "style.vec4Expression.schema.json"
+        }, {
+          "$ref" : "style.condition.schema.json"
+        }],
+        "description" : "Sets near and far scaling properties of a feature based on the features's distance from the camera. A feature's scale will interpolate between the `y` and `w` values while the camera distance falls within the upper and lower bounds of the specified `x` and `z` value. Outside of these ranges, the features's scale remains clamped to the nearest bound.  If undefined, `scaleByDistance` will be disabled.",
```
lilleyse: There are minor differences between the descriptions in the schema and in the spec text, it's probably worth going through all of these to make sure they are the same.

For example this one does not include "in meters".

lilleyse:
That's all my comments. This is really close now.

TileFormats/Geometry/README.md
bagnell: Ah, I missed that. Thanks.

lilleyse:
The updates look good. I'll check over the Cesium PR quickly and then this will be good to go.

lilleyse:
👍

PR #262, opened 2017-09-30
fix typo

PrincessGod:
Is there should be `content.boundingVolume` ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ A file extension is not required for `content.url`.  A content's [tile format](#
 
 ![](figures/contentsBox.png)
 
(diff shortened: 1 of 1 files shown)


lilleyse:
Thanks, good catch.

While this is a minor fix, in order to merge we'll need a CLA from you: https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CONTRIBUTING.md#contributor-license-agreement-cla.

If this seems like too much trouble I can also make this fix directly.

PrincessGod:
@pjcozzi  Thanks, I just want to make sure if this is true, maybe you fix it is easier.

lilleyse:
Ok, fixed here: https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/264. Thanks again!

PR #274, opened 2017-12-22
Remove extends from viewerRequestVolume in schema

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/273

@pjcozzi do you remember if there was a reason for using `extends`?


pjcozzi:
Probably just my mistake.

PR #279, opened 2018-01-23
Vector Tiles from 3d-tiles-next

lilleyse:
This branch contains all the vector tiles work. @pjcozzi should we merge now or wait until post `1.0`?


pjcozzi:
We should keep the `3d-tiles-next` branch parked until it is ready.  I suggest closing this.

PR #301, opened 2018-04-17
First Round of 3D Tiles 1.0 updates

### LINKED ISSUE #8: How are extensions handled?
Extensions in both tiles.json and the tiles themselves.  We'll most likely take the same approach as glTF (for tiles.json) and quantized-mesh (for tiles).

Does tiles.json support both `extensions` and `extras` like glTF?  Perhaps `extra` is useful for application-specific per-tile metadata that the app could process with a load event.  We need concrete use cases.

### LINKED ISSUE #60: MIME types for tile formats
Cesium's Node.js development server is using `application/octet-stream` for now.

Related glTF discussion: https://github.com/KhronosGroup/glTF/issues/412

### LINKED ISSUE #94: Reference glTF 2.0 spec
For example, update [here](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Batched3DModel#binary-gltf) and elsewhere to be glTF 2.0, not glTF 1.0.

glTF 2.0 updates: https://github.com/KhronosGroup/glTF/issues/605

Also consider if there are any other glTF extensions, like unsigned int indices, that we want to include.

### LINKED ISSUE #140: Shader styling considerations
Point cloud styling engines may often use a shader implementation. Currently the spec is written with a bias towards JavaScript and its implicit type conversions and ability to check equality between different types, which are not possible in a language like GLSL 1.0. Shader implementations will also have major issues with strings and regular expressions. We need to decide what aspects of the styling language can be supported for stricter languages like GLSL and C++.
- [ ] **Shader Styling**
  - [ ] Unsupported:
    - [ ] `isNan`, `isFinite`(supported in later versions of GLSL with `isnan` and `isinf`)
    - [ ] Strings, not supported in the GLSL styling engine or the batch table binary
      - [ ] This includes accessing color components like `color()["r"]` or vectors with `vec4(1.0)["x"]`
    - [ ] Regular expressions
    - [ ] Arrays that aren't length 2, 3, or 4 (technically supported but cumbersome to use)
    - [ ] null and undefined - use a default value like 0? https://github.com/AnalyticalGraphicsInc/cesium/issues/4637
    - [ ] type mismatching - comparing floats to bools to vectors will result in shader compilation errors
    - [ ] Index out of bounds for arrays/vectors/colors will result in shader compilation errors whereas it returns undefined in JS

More discussion here: https://github.com/AnalyticalGraphicsInc/cesium/pull/4336#issuecomment-250238675

### LINKED ISSUE #184: Consider making a top-level tileset.json the required entry point
Currently, any JSON file anywhere in the hierarchy can be the entry point of a tileset, there is also no requirement for it to be named `tileset.json`.  This makes it impossible to write a generic loader because it requires whomever created the tileset to identify which JSON file to use.

Requiring a `tileset.json` at the top level of the hierarchy to always be the entry point fixes this problem and shouldn't negatively impact any use cases I can think of.  In fact, the way the spec is worded, I always thought this was already the case (because we use the term `top-level` in a few places.

This is somewhat related to #89, since you want to be able to read a .3dtiles database without any external knowledge, but I thought it warranted it's own issue since it has come up a few times in my attempt to read tilesets generated by someone other than myself.

### LINKED ISSUE #211: Relax external tileset restrictions
From the spec:

```
root.geometricError === tile.geometricError,
root.refine === tile.refine, and
root.boundingVolume === tile.content.boundingVolume (or root.boundingVolume === tile.boundingVolume when tile.content.boundingVolume is undefined).
root.viewerRequestVolume === tile.viewerRequestVolume or root.viewerRequestVolume is undefined.
```

Are these restrictions actually needed? This makes serving dynamic external tilesets harder.

### LINKED ISSUE #225: Padding
From https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/222#issuecomment-303121639

>I definitely see the benefit of making padding rules part of the spec, not just an implementation note. I'll open a separate PR for that.

via @lilleyse

### LINKED ISSUE #263: Move CESIUM_RTC to the feature table for b3dm
Now that b3dm tiles have a feature table, it would probably make sense for  `RTC_CENTER` to exist in the feature table. It's still worth supporting `CESIUM_RTC` as a glTF extension, but the feature table `RTC_CENTER` seems preferable from a spec point of view.

### LINKED ISSUE #272: Relative URL : ambiguity on URL root
In the `tileset.json` spec, it is specified that tile URLs can either be relative or absolute. In the case of a relative URL that happens in a non-root tileset, it is unclear whether the URL should be created starting from the root tileset, or the non-root tileset.

Example:
```
tileset.json
   anotherTileset.json
      tile.b3dm
```
Given that the URL to `tileset.json` is  `www.example.com/tileset.json`, what is the absolute URL of tile.b3dm ? Is it `www.example.com/tile.b3dm` or `www.example.com/anotherTileset/tile.b3dm` ?

### LINKED ISSUE #280: Bounding Volume coordinate system clarification
Hi,

I'm working on a 3D Tiles exporter and renderer.  I'm trying to understand the relationship between the `3D Tiles coordinate frame` and the `glTF coordinate frame`, especially in relation to bounding volumes and the gltfUpAxis property.

As an example, lets say we have a box stored in a glTF file with Y as the up axis.  Our goal is to create a tileset containing this box. In our tileset, we need to store the bounding volume for the box.  My understanding is that the bounding volume needs to be stored in `3D Tiles Frame` (z-up) NOT in `glTF frame` (y-up).  So in this case, in our exporter we first compute the glTF files bounds in `glTF frame` and then apply a `glTF to 3D Tiles` transform to get it into `3D Tiles frame`.  In the renderer, we then need to apply the opposite of this transform when we read the glTF file to convert it into `3D Tiles frame` at runtime.  

I've made the following diagram to illustrate the process of going from `glTF Frame` to `3d Tiles Frame` as I understand it.
image

In the diagram, I assume we want to render the box n the given orientation regardless of how we define the coordinate frame.  However, there are many configurations for a right handed coordinate system in which Z is up.  I've created two examples (A) and (B).  Depending on which of these we choose, our `glTF to 3D Tiles` transform will be different.  This presents a problem because the tileset generator and renderer need to use the same convention.  Based on [this changeset](https://github.com/AnalyticalGraphicsInc/cesium/pull/4962/files), it looks like cesium uses (B) but I don't see this documented in the specification anywhere.

I haven't fully worked this out, but it seems like things would only get more confusing once the options for gltfUpAxis get added to the mix.

Please let me know if I've misunderstood how this works.

I think it would be very useful to add some additional text to the README about these transformations and when they should be applied (both during tileset creation and rendering).  I'm assuming this should be applied right after loading the glTF file before applying the Tile.transform property. 

Overall, I've found it pretty confusing that `3D Tiles` uses a different frame then the underlying content storage mechanism.  It almost feels like each tile or the tileset could specify a `content to tileset` matrix to be used when converting the content into tileset frame.   

Thanks for the help!

Cheers,
-Alex

### LINKED ISSUE #295: Computing final ECEF position of vertex arrays inside B3DM/GLTF
Been struggling with this for days now.  Can someone please explain to me whether or not this is correct.

I have binary GLTF data that has a CESIUM_RTC value.  I've interpreted that to mean that the data itself is in the ECEF coordinate system.  I've confirmed that by rotating the data to ENU via multiplying an inverse orientation matrix who's columns are formed by the local ENU vectors (computed from ECEF CESIUM_RTC converted to lat/long)

However, when converting the per-vertex ECEF values to LLH (using algorithm )https://www.mathworks.com/matlabcentral/fileexchange/7941-convert-cartesian--ecef--coordinates-to-lat--lon--alt?focused=5062924&tab=function), I get wildly incorrect height values and arbitrarily rotated 3D data with respect to LAT/LON plane.  

**My question**:  is the proper way to determine the ECEF position of an embedded vertex inside B3DM/GLTF to simply add the cesiumRTC position to it?  And If so, why the heck am I getting wrong values when converting cesiumRTC+pos values to wgs84 LLH??

### LINKED ISSUE #298: Oct encoding clarification for pnts and i3dm
In [pnts](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/PointCloud) and [i3dm](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/TileFormats/Instanced3DModel) make it clearer that oct encoded values are stored in unsigned unnormalized range [0, 65535] or [0, 255] and then converted to a signed normalized range [-1, 1] at runtime.

ggetz:
First round of edits for 3D Tiles 1.0, tried to minimize organizational and formatting changes to be tackled later.

Fixes #298
Fixes #295
Fixes #280
Fixes #272
Fixes #263
Fixes #225 
Fixes #211 
Fixes #184 
Fixes #140 
Fixes #94 
Fixes #60 
Fixes #8


pjcozzi:
Please target a new 1.0 branch.

lilleyse:
For #8, add `extras` and `extensions` to `boundingVolume.schema.json`, `batchTable.schema.json/hierarchy/items`, and `featureTable.schema` (even though b3dm, i3dm, and pnts derive from this and include these properties themselves, it would be good to be explicit in the feature table schema).

Can you also open a PR making any spec changes into the vector tiles branch?

Styling/README.md
```diff
@@ -1340,19 +1342,11 @@ For example:
...
-**TODO : add note about GLSL implementations requires strict type comparisons among other things: https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/140**
-
-## File Extension
-
-TBA
-
-## MIME Type
-
-_TBA, [#60](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/60)_
+> Implementation Note: Point cloud styling engines may often use a shader (GLSL) implementation, however some features of the expression language are not possible in pure a GLSL implementation. Features that must account for these inconsistencies include evaluation of `isNan` and `isFinite` (GLSL 2.0+ supports `isnan` and `isinf` for these functions respectively); the types `null` and `undefined`; strings, including accessing object properties (`color()['r']`) and batch table values; regular expressions; arrays of lengths other than 2, 3, or 4; mismatched type comparisons (`float` to `bool`); and handling of "index out of bounds" errors.
```
lilleyse: Is it possible to have a bulleted list within a quoted section? That might be easier to read.

Styling/README.md
```diff
@@ -1340,19 +1342,11 @@ For example:
...
 
-`application/json`
+## File Extensions and MIME Types
+ 
+* Tileset styles use the `.json` extension and the `application/json` mime type.
```
lilleyse: Same note as before where the extension is not required.

TileFormats/Batched3DModel/README.md
```diff
@@ -119,16 +120,20 @@ The vertex shader can be modified at runtime to use `a_batchId` to access indivi
 
 When a Batch Table is present or the `BATCH_LENGTH` property is greater than `0`, the `batchId` attribute (with the parameter semantic `_BATCHID`) is required; otherwise, it is not.
```
lilleyse: The gltf snipped above is still based on 1.0 techniques/shaders. For 2.0 all that's required is a primitive contains a `_BATCHID` attribute.

TileFormats/Batched3DModel/README.md
```diff
@@ -119,16 +120,20 @@ The vertex shader can be modified at runtime to use `a_batchId` to access indivi
...
+### Coordinate Reference System (CRS)
 
-## File Extension
+Vertex positions are defined according to a right-handed coordinate system where Y-axis is up. [glTF Coordinate System and Units](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#coordinate-system-and-units)
 
-`.b3dm`
+Vertex positions may be defined relative-to-center for high-precision rendering [1]. If defined, `RTC_CENTER` specifies the center position and all vertex positions are treated as relative to this value.
 
-The file extension is optional. Valid implementations ignore it and identify a content's format by the `magic` field in its header.
+> Implementation Note: Clients may also use the glTF [CESIUM_RTC](https://github.com/KhronosGroup/glTF/tree/master/extensions/1.0/Vendor/CESIUM_RTC) extension and define points relative to center.
```
lilleyse: We are phasing away from this extension, so I wouldn't include this note.

TileFormats/Batched3DModel/README.md
```diff
@@ -119,16 +120,20 @@ The vertex shader can be modified at runtime to use `a_batchId` to access indivi
...
 When a Batch Table is present or the `BATCH_LENGTH` property is greater than `0`, the `batchId` attribute (with the parameter semantic `_BATCHID`) is required; otherwise, it is not.
 
-Although not strictly required, clients may find the glTF [CESIUM_RTC](https://github.com/KhronosGroup/glTF/tree/master/extensions/1.0/Vendor/CESIUM_RTC) extension useful for high-precision rendering.
+### Coordinate Reference System (CRS)
 
-## File Extension
+Vertex positions are defined according to a right-handed coordinate system where Y-axis is up. [glTF Coordinate System and Units](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#coordinate-system-and-units)
 
-`.b3dm`
+Vertex positions may be defined relative-to-center for high-precision rendering [1]. If defined, `RTC_CENTER` specifies the center position and all vertex positions are treated as relative to this value.
```
lilleyse: Here it should be noted that `RTC_CENTER` is z-up whereas vertex positions are y-up.

In general I think our coordinate system messaging is still not _quite_ there... still need to think about the clearest way possible to describe it.

lilleyse:
Is https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/225 (padding) in here?

lilleyse:
I made some edits as I went along in https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/301/commits/192a6326145d9f3d98fb5ddd91f4483a98d8d886. More to come later.

lilleyse:
#295 and #280 will need a bit of work before they are done, maybe we can discuss offline
#184 - might be good to explicitly mention that the top level json is not required to have a specific name
#225 - doesn't seem to be included here. It would be good if each tile format has a easy to find link to the padding requirements

I think the rest of the issues look good or are covered by comments.

README.md
```diff
@@ -164,9 +164,35 @@ _TODO: include a screenshot of each bounding volume type._
 
 A tile references a _feature_ or set of _features_, such as 3D models representing buildings or trees, points in a point cloud, or polygons, polylines, and points in a vector dataset.  These features may be batched together into essentially a single feature to reduce client-side load time and WebGL draw call overhead.
 
-## Tile metadata
+A 3D tileset consists of at least one tileset `.json` file specifying the metadata and the tree of tiles, as well as any referenced tile content files which may be any valid tile format, described below.
```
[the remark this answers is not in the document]
ggetz: I will change `.json` to `JSON` throughout, but I do specify in the file extension and MIME type section that `Tileset files use the `.json` extension and the `application/json` MIME type.` Is that not true?

README.md
```diff
@@ -164,9 +164,35 @@ _TODO: include a screenshot of each bounding volume type._
...
+
+  1. JSON must use UTF-8 encoding without BOM.
+  2. All strings defined in this spec (properties names, enums) use only ASCII charset and must be written as plain text.
+  3. Names (keys) within JSON objects must be unique, i.e., duplicate keys aren't allowed.
+
+## URLs
+
+3D Tiles use URLs to reference tile content.
+
+All URLs must be valid resolvable URIs, and may be absolute or relative. When the url is relative, it's base is always referring to the referring tileset `.json` file.
```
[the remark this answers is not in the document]
ggetz: Do we want to specify URI schemes here as well (like `https://`)? glTF does so: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#uris

README.md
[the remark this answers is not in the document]
lilleyse: Yup that is still correct. But we should note that the extension is optional.

README.md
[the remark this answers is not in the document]
lilleyse: I don't see the harm in following glTF's lead here. If we follow glTF's same wording we should remove required support for absolute uris.

README.md
```diff
@@ -558,6 +599,10 @@ This colors features with a temperature above 90 as red and the others as white.
 
 For complete details, see the [Declarative Styling](Styling/README.md) spec.
 
+## Specifying Extensions and Application-Specific Extras
+
+3D Tiles defines extensions to allow the base specification to have extensibility for new features, as well as extras to allow for application specific metadata.
+
```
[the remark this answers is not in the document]
ggetz: Should we also include the `extensionsUsed` and `extensionsRequired` properties in the tileset schema?

For an extension, what about `VENDER_collision_volume` to define a collision volume for the tile?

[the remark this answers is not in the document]
lilleyse: > Should we also include the extensionsUsed and extensionsRequired properties in the tileset schema?

Sure

> For an extension, what about VENDER_collision_volume to define a collision volume for the tile?

Ah yeah, that's a good example.

ggetz:
@lilleyse Updated with the requested edits mentioned here and talked about offline.

TileFormats/Batched3DModel/README.md
```diff
@@ -49,86 +68,93 @@ in the Cesium implementation of 3D Tiles.
...
+{
+    "accessors": [
+        {
+            "bufferView": 1,
+            "byteOffset": 0,
+            "componentType": 5125,
+            "count": 4860,
+            "max": [2],
+            "min": [0],
+            "type": "SCALAR"
```
lilleyse: Below this example mention that `accessor.type` must be `"SCALAR"`, but there are no restrictions on the other properties (as long as they are in accordance with glTF).

TileFormats/Batched3DModel/README.md
```diff
@@ -49,86 +68,93 @@ in the Cesium implementation of 3D Tiles.
...
 
-#### Global Semantics
+#### Global semantics
 
 These semantics define global properties for all features.
 
 | Semantic | Data Type | Description | Required |
 | --- | --- | --- | --- |
 | `BATCH_LENGTH` | `uint32` | The number of distinguishable models, also called features, in the batch. If the Binary glTF does not have a `batchId` attribute, this field _must_ be `0`. | :white_check_mark: Yes. |
+| `RTC_CENTER` | `float32[3]` | A 3-component array of numbers defining the center position in _z_-up coordinates when positions are defined relative-to-center. | :red_circle: No. |
```
lilleyse: Drop `in z-up coordinates` and add a link to the CRS section below.

TileFormats/Instanced3DModel/README.md
```diff
@@ -6,24 +6,55 @@
...
 
 ![header layout](figures/header-layout.png)
 
+### Padding
+
+A tile's `byteLength` must be aligned to an 8-byte boundary.
+
+The [binary glTF](#gltf) (if present) must start and end on an 8-byte alignment so that glTF's byte-alignment guarantees are met. This can be done by padding the [Feature Table](../FeatureTable/README.md#padding) or [Batch Table](../BatchTable/README.md#padding) if they are present.
+
+Otherwise, if the glTF field is a UTF-8 string, it must be padded with trailing Space characters (`0x20`) to satisfy alignment requirements of the tile.
```
lilleyse: Mention that these trailing spaces should be removed at runtime before requesting the glTF.

TileFormats/Instanced3DModel/README.md
```diff
@@ -92,70 +123,64 @@ These semantics define global properties for all instances.
...
-This is suitable for instanced models like trees whose orientation is always facing up from their position on the ellipsoid's surface.
+If `NORMAL_UP` and `NORMAL_RIGHT` or `NORMAL_UP_OCT32P` and `NORMAL_RIGHT_OCT32P` are not present, the instance will not have a custom orientation. If `EAST_NORTH_UP` is `true`, the instance is assumed to be on the `WGS84` ellipsoid and its orientation will default to the `east/north/up` reference frame at its cartographic position.
+This is suitable for instanced models such as trees whose orientation is always facing up from 
 
-### Instance Position
+### Instance position
 
-`POSITION` defines the location for an instance before any tile transforms are applied. Positions may be defined relative-to-center for high-precision rendering [4]. `RTC_CENTER` defines the center position.
+`POSITION` defines the location for an instance before any tile transforms are applied. Positions may be defined relative to center for high-precision rendering, see [Precisions, Precisions](http://blogs.agi.com/insight3d/index.php/2008/09/03/precisions-precisions/). `RTC_CENTER` defines the center position.
```
lilleyse: The blog link doesn't seem to exist anymore. Does anyone know if this blog is still live anywhere? In any case this should just stay as a footnote. (Same comments for pnts)

TileFormats/Instanced3DModel/README.md
```diff
@@ -220,52 +245,43 @@ var featureTableBinary = Buffer.concat([positionQuantizedBinary, normalUpOct32PB
...
 
-* a UTF-8 string, which contains a url to a glTF model.
+`header.gltfFormat` determines the format of the glTF field
 
-When the value of `header.gltfFormat` is `1`, the glTF field is
+* When the value of `header.gltfFormat` is `0`, the glTF field is a UTF-8 string, which contains a uri of the glTF model content.
+* When the value of `header.gltfFormat` is `1`, the glTF field is a binary blob containing binary glTF.
 
-* a binary blob containing binary glTF.
+> Implementation note: Prefer a binary glTF blob to an glTF model data uri.
```
lilleyse: I know this note was my suggestion, but I think it can be removed since I doubt in practice many people will think to store the glTF as a data uri. The sentence as it is might just lead to confusion.

TileFormats/PointCloud/README.md
```diff
@@ -7,20 +7,47 @@
...
-A tile is composed of a header section immediately followed by a body section.
-
-**Figure 1**: Point Cloud layout (dashes indicate optional fields).
+A tile is composed of a header section immediately followed by a body section. The following figure shows the Point Cloud layout (dashes indicate optional fields):
 
 ![](figures/layout.png)
 
+### Padding
+
+A tile's `byteLength` must be aligned to an 8-byte boundary.
```
lilleyse: Link to the feature table and batch table padding requirements as well.

TileFormats/PointCloud/README.md
```diff
@@ -85,61 +112,61 @@ These semantics define global properties for all points.
 
 Examples using these semantics can be found in the [examples section](#examples) below.
 
-### Point Positions
-
-`POSITION` defines the position for a point before any tileset transforms are applied. Positions may be defined relative-to-center for high-precision rendering [4]. `RTC_CENTER` defines the center position.
+### Point positions
 
-#### Quantized Positions
+`POSITION` defines the position for a point before any tileset transforms are applied. Positions may be defined relative-to-center for high-precision rendering, see [Precisions, Precisions](http://blogs.agi.com/insight3d/index.php/2008/09/03/precisions-precisions/). `RTC_CENTER` defines the center position.
```
lilleyse: The last sentence feels incomplete. Maybe:

>When defined, `RTC_CENTER` specifies the center position that all positions are relative to.

TileFormats/PointCloud/README.md
```diff
@@ -85,61 +112,61 @@ These semantics define global properties for all points.
...
-If `POSITION` is not defined, positions may be stored in `POSITION_QUANTIZED` which defines point positions relative to the quantized volume.
-If neither `POSITION` or `POSITION_QUANTIZED` are defined, the tile does not need to be rendered.
+#### Quantized positions
 
-A quantized volume is defined by `offset` and `scale` to map quantized positions into model space.
+If `POSITION` is not defined, positions may be stored in `POSITION_QUANTIZED`, which defines point positions relative to the quantized volume.
+If neither `POSITION` nor `POSITION_QUANTIZED` is defined, the tile does not need to be rendered.
 
-**Figure 2**: A quantized volume based on `offset` and `scale`.
+A quantized volume is defined by `offset` and `scale` to map quantized positions into model space. The following figure shows a quantized volume based on `offset` and `scale`:
```
lilleyse: `model space` - while common in graphics terminology, might be misleading here.

TileFormats/Instanced3DModel/README.md
lilleyse: Actually I see that throughout foot notes were removed, so fine to keep inline once the link is fixed.

TileFormats/PointCloud/README.md
lilleyse: `local space` I think is fine.

ggetz:
Thanks @lilleyse, made the updates, updated the CRS wording, and tweaked the images.

lilleyse:
A few other miscellaneous notes not covered by any issues:

* All tile formats should specify that they are little endian. Only some do.
* Should we remove JavaScript terminology throughout? Except in example sections?
* In the [tree picture](https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/figures/tilesets.jpg) tiles.json should be renamed to tileset.json. Even though both are valid names tileset.json is more familiar. The picture should also be resized as it takes up a lot of space.
* This is an annoying breaking change, but `tile.content.url` should really be `tile.content.uri`. It would be relatively easy to add backwards compatibility in implementations. Thoughts on whether this change is worthwhile?
* Section on additive vs replacement refinement. I don't think we ever really explain the difference. Might be good to provide definitions and example use cases.
* `If featureTableJSONByteLength equals zero, the tile does not need to be rendered.` Should we instead just say the feature table is required for all tile formats? (And update figures accordingly).

ggetz:
> This is an annoying breaking change, but tile.content.url should really be tile.content.uri. It would be relatively easy to add backwards compatibility in implementations. Thoughts on whether this change is worthwhile?

Yeah, that's true. glTF uses `uri` throughout.

> Should we instead just say the feature table is required for all tile formats? (And update figures accordingly)

Yes, I think that's much more definitive than the current spec.

ggetz:
Added glTF `CESIUM_z_up` extension: https://github.com/KhronosGroup/glTF/pull/1338

ggetz:
Updated figures.

lilleyse:
Awesome work! Ready to merge to `1.0`.

PR #307, opened 2018-05-10
Remove CESIUM_z_up extension and add a detailed implementation note

lilleyse:
Specifically check out Coordinate Reference System and contained glTF section.

Thoughts @ggetz @pjcozzi @emackey?

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,16 +392,16 @@ The computed transform for each tile is:
 * `T4`: `[T0][T1][T4]`
 
 The positions and normals in a tile's content may also have tile-specific transformations applied to them _before_ the tile's `transform` (before indicates post-multiplying for affine transformations).  Some examples are:
-* `b3dm` and `i3dm` tiles embed glTF, which defines its own node hierarchy and coordinate system. `tile.transform` is applied after these transforms are resolved. See [coordinate reference system](#coordinate-reference-system-crs).
+* `b3dm` and `i3dm` tiles embed glTF, which defines its own node hierarchy and coordinate system. `tile.transform` is applied after these transforms are resolved. See [coordinate reference system](#gltf).
 * `i3dm`'s Feature Table defines per-instance position, normals, and scales.  These are used to create per-instance 4x4 affine transform matrices that are applied to each instance before `tile.transform`.
 * Compressed attributes, such as `POSITION_QUANTIZED` in the Feature Tables for `i3dm` and `pnts`, and `NORMAL_OCT16P` in `pnts` should be decompressed before any other transforms.
 
 Therefore, the full computed transforms for the above example are:
 * `TO`: `[T0]`
 * `T1`: `[T0][T1]`
-* `T2`: `[T0][T2][pnts-specific Feature Table properties-derived transform]`
-* `T3`: `[T0][T1][T3][b3dm-specific transform, including the glTF node hierarchy and coordinate system transform]`
-* `T4`: `[T0][T1][T4][i3dm-specific transform, including per-instance Feature Table properties-derived transform and the glTF node hierarchy]`
+* `T2`: `[T0][T2][pnts-specific transform, including RTC_CENTER (if defined)]`
+* `T3`: `[T0][T1][T3][b3dm-specific transform, including RTC_CENTER (if defined), coordinate system transform, and glTF node hierarchy]`
+* `T4`: `[T0][T1][T4][i3dm-specific transform, including per-instance Feature Table properties-derived transform, coordinate system transform, and glTF node hierarchy]`
 
 #### Implementation example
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -195,25 +196,38 @@ All angles are in radians.
...
 
-Some tile content types such as [Batched 3D Model](TileFormats/Batched3DModel/README.md) and [Instanced 3D Model](TileFormats/Instanced3DModel/README.md) embed glTF. The [glTF specification](https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#coordinate-system-and-units) defines a right-handed coordinate system with the _y_ axis as up. For consistency with the _z_-up coordinate system of 3D Tiles, embedded glTFs may instead use the [`CESIUM_z_up` glTF extension](https://github.com/ggetz/glTF/tree/cesium-z-up-ext/extensions/2.0/Vendor/CESIUM_z_up).
+An additional [tile transform](#tile-transform) may be applied to transform a tile's local coordinate system to the parent tile's coordinate system.
 
-If the `CESIUM_z_up` glTF extension is not used, the glTF model is considered to by _y_-up and must be transformed to a _z_-up coordinate system at runtime. This is done by rotating the model about the _x_-axis by &pi;/2 radians. Equivalently, apply the following matrix transform:
+The [region](#region) bounding volume type is defined in [EPSG 4326](http://nsidc.org/data/atlas/epsg_4326.html) coordinates.
```
ggetz: I think it's important to state `geographic coordinate system (latitude, longitude, height)` because everything else is in a right handed (x,y,z) coordinate system.

README.md
```diff
@@ -195,25 +196,38 @@ All angles are in radians.
...
+
+For consistency with the _z_-up coordinate system of 3D Tiles, glTFs must be transformed from _y_-up to _z_-up at runtime. This is done by rotating the model about the _x_-axis by &pi;/2 radians. Equivalently, apply the following column-major matrix transform:
 ```json
-[
-1.0, 0.0,  0.0, 0.0,
-0.0, 0.0, -1.0, 0.0,
-0.0, 1.0,  0.0, 0.0,
-0.0, 0.0,  0.0, 1.0
-]
+[1,0,0,0,0,0,1,0,0,-1,0,0,0,0,0,1]
```
ggetz: This formatting makes the matrix hard to see at a glance, and is inconsistant with how we format them elsewhere in the spec.

Both here and throughout the remaining changes.

README.md
```diff
@@ -195,25 +196,38 @@ All angles are in radians.
...
 Note that glTF defines its own node hierarchy, where each node has a transform. These transforms are applied before the coordinate system transform is applied.
 
-> Implementation Note: Using the `CESIUM_z_up` extension is preferred to transforming the model to a _z_-up coordinate system at runtime.
-
-Some [bounding volumes](#bounding-volumes) may specify bounds using a geographic coordinate system (latitude, longitude, height), in which case use the WGS 84 datum as defined in [EPSG 4326](http://nsidc.org/data/atlas/epsg_4326.html).
+> **Implementation note:** when working with source data that is inherently _z_-up, such as data in WGS 84 coordinates or a local _z_-up coordinate system, a common workflow is:
+> * Mesh data, including positions and normals, is not modified - it can remain _z_-up.
+> * The root node matrix specifies a _z_-up to _y_-up transform. This transforms the source data into a _y_-up coordinate system as required by glTF. For example:
+>
+>```
```
ggetz: Specify the language for syntax highlighting like `json`

README.md
lilleyse: Yeah I think I prefer a row-major, line separated matrix too.

The reason I changed it here was to be in the same form as the matrix in the gltf snippet so it was obvious how the two compare. But I'll change it back.

emackey:
I think letting go of the extension is the right move for both the glTF and the 3D Tiles ecosystems.

Such an extension can't be vendor-specific to Cesium if we expect other vendors to implement 3D Tiles as an open standard and interoperate with Cesium and ion-generated tiles.  For example, desktop applications that want to render ion-served tiles will need a desktop implementation of the single-vendor extension.

The idea was poorly received by several participants in the main glTF repo, as it runs counter to the spirit there: glTF asks everyone to agree to use the same conventions (axes selection, texture channel selection, etc) specifically to avoid the pain felt in older formats that allowed exporters to write whatever was convenient for the exporter, introducing too much variance into the resulting library of assets for any one reader to fully cope with.

Also, the extension was marked optional.  But authors of optional extensions are asked to include fallbacks for core readers to do the best they can do with the model.  The fallback here is incredibly trivial: Just add the above matrix to the root node, at which point the whole extension becomes moot.

I think letting go of this extension is well worth the effort, particularly to maintain the message that 3D Tiles uses core glTF without any vendor extensions, and the ecosystem is stable.

ggetz:
OK I'm good with this solution. Should we go ahead and close the z-up extension?

@lilleyse I made some tweaks to formatting and wording just for readability, feel free to switch it back don't like any of those changes.

lilleyse:
The formatting tweaks are good.

Yeah let's close the z-up extension.

PR #310, opened 2018-05-18
Draco compression for pnts

lilleyse:
This PR adds support for [Draco](https://google.github.io/draco/) compression to the Point Cloud tile format.

Do not merge yet - this is targeting `1.0-formatting` just to minimize the amount of formatting differences, but will eventually target `3d-tiles-next`.

Cesium implementation: https://github.com/AnalyticalGraphicsInc/cesium/pull/6559

TODO:
* [x] Schema updates


specification/TileFormats/PointCloud/README.md
```diff
@@ -171,6 +171,32 @@ This is useful for per-object picking and storing application-specific metadata
...
+### Draco
+
+Per-point semantics such as position, color, normal, and batch id may may be compressed with [Draco](https://google.github.io/draco/). Draco binary data is stored in the Feature Table binary and a list of the compressed per-point semantics is stored in the Feature Table JSON.
+
+The `DRACO` object in the Feature Table JSON header contains three properties:
+   * `byteOffset` specifies a zero-based offset relative to the start of the Feature Table binary body.
+   * `byteLength` specifies the length of the Draco binary data.
+   * `semantics` is an array containing the per-point semantics contained in the Draco binary data. Allowed semantics are `"POSITION"`, `"RGBA"`, `"RGB"`, `"NORMAL"`, and `"BATCH_ID"`.
+
+Point attribute data shall be encoded and decoded using the parameters in the following table.
```
ggetz: Should we specify that they must be encoded/decoded with the draco tools as well (and link to them)?

specification/TileFormats/PointCloud/README.md
```diff
@@ -171,6 +171,32 @@ This is useful for per-object picking and storing application-specific metadata
 The `BATCH_ID` semantic may have a `componentType` of `UNSIGNED_BYTE`, `UNSIGNED_SHORT`, or `UNSIGNED_INT`. When `componentType` is not present, `UNSIGNED_SHORT` is used.
 The global semantic `BATCH_LENGTH` defines the number of unique `batchId` values, similar to the `batchLength` field in the [Batched 3D Model](../Batched3DModel/README.md) header.
 
+### Draco
+
+Per-point semantics such as position, color, normal, and batch id may may be compressed with [Draco](https://google.github.io/draco/). Draco binary data is stored in the Feature Table binary and a list of the compressed per-point semantics is stored in the Feature Table JSON.
```
ggetz: `batch id` -> `batchId` or `batch ID`

specification/TileFormats/PointCloud/README.md
```diff
@@ -171,6 +171,32 @@ This is useful for per-object picking and storing application-specific metadata
...
+
+Semantic | Data Type | Number of Components | Draco Type
+---|---|---|---
+`POSITION` | `float32` | 3 | `POSITION`
+`NORMAL` | `float32` | 3 | `NORMAL`
+`RGBA` | `uint8` | 4 | `COLOR`
+`RGB` | `uint8` | 3 | `COLOR`
+`BATCH_ID` | `uint8`, `uint16`, or `uint32` | 1 | `GENERIC`
+
+If a semantic type is defined in both `DRACO.semantics` and in the Feature Table JSON header, the Draco attribute data should be used for rendering.
```
ggetz: Also state that `POSITION` or `POSITION_QUANTIZED` is no longer required.

specification/TileFormats/PointCloud/README.md
```diff
@@ -325,6 +351,22 @@ var batchTableJSON = {
...
+    POINTS_LENGTH : 4,
+    DRACO : {
+        byteOffset : 0
+        byteLength : 36,
+        semantics : ['POSITION', 'RGB']
+    }
+};
+
+// Use Draco encoder to build dracoBuffer
+var featureTableBinary = dracoBuffer;
```
ggetz: I don't think this line in the example adds much. Instead I would say that the binary much be encoded with the Draco encoder in the section above and link to the encoder tool(s).

specification/TileFormats/PointCloud/README.md
ggetz: Maybe add a quick sentence on why this would be used, eg. the glTF extension says

> This allows glTF to support streaming compressed geometry data instead of the raw data

specification/TileFormats/PointCloud/README.md
```diff
@@ -171,6 +171,32 @@ This is useful for per-object picking and storing application-specific metadata
...
+
+Point attribute data shall be encoded and decoded using the parameters in the following table.
+
+Semantic | Data Type | Number of Components | Draco Type
+---|---|---|---
+`POSITION` | `float32` | 3 | `POSITION`
+`NORMAL` | `float32` | 3 | `NORMAL`
+`RGBA` | `uint8` | 4 | `COLOR`
+`RGB` | `uint8` | 3 | `COLOR`
+`BATCH_ID` | `uint8`, `uint16`, or `uint32` | 1 | `GENERIC`
```
ggetz: How do we define the component here?

lilleyse:
Sorry for the churn, but I noticed a flaw with the current design yesterday. Since draco shuffles around the order of points we'll need to encode batch table properties in the draco buffer so that feature table properties and batch table properties are in sync.

This realization led to a few changes:

* We need to store Draco attribute id's now that multiple properties may use the `GENERIC` type. Previously we just stored names of the encoded semantics.
* https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/310#discussion_r189711728 - this comment brings up a good point, even though Draco encodes the type/component type in its buffer, we should still include the `componentType`/`type` in the json.
* Draco point compression might be better off as an extension rather than core spec, like in glTF.

The new design is modeled more after glTF and looks like:

**Feature Table**
```javascript
{
    POSITION: {
        byteOffset: 0 // ignored if draco encoded
    },
    RGB: {
        byteOffset: 0
    },
    BATCH_ID: {
        byteOffset: 0,
        componentType: 'UNSIGNED_BYTE'
    },
    extensions: {
        "3DTILES_draco_point_compression": {
            properties: {
                POSITION: 0,
                RGB: 1,
                BATCH_ID: 2
            },
            byteOffset: 0,
            byteLength: 100
        }
    }
}
```

**Batch Table**

```javascript
{
    Intensity: {
        byteOffset: 0,
        type: 'SCALAR',
        componentType: 'UNSIGNED_BYTE'
    },
    Classification: {
        byteOffset: 0,
        type: 'SCALAR',
        componentType: 'UNSIGNED_BYTE'
    }
    extensions: {
        "3DTILES_draco_point_compression": {
            properties: {
                Intensity: 3,
                Classiciation: 4
            },
            byteOffset: 0,
            byteLength: 100
        }
    }
}
```

If a property is encoded in Draco its `byteOffset` is ignored.

lilleyse:
One other thing, the draco buffer is still stored in the feature table binary but this is pretty awkward when batch table properties are stored there as well. Ideally draco would be its own buffer after the feature table and batch table sections but I wasn't sure if this was worth it since it complicates the overall tile design.

pjcozzi:
Why does this target the 1.0 branch?  Isn't that branch just for the OGC Community Standard?  Even as an extension, this should target a `3d-tiles-next` branch.

lilleyse:
I mentioned this in the PR description:

> Do not merge yet - this is targeting 1.0-formatting just to minimize the amount of formatting differences, but will eventually target 3d-tiles-next.

Due to all the formatting changes and moving around of files I thought it would be best to write this extension against the most recent spec. Once `3d-tiles-next` is updated to use the new formatting as well I can retarget to that branch.

pjcozzi:
Ah, gotcha.

lilleyse:
@ggetz updated. Rewritten to account for the changes in https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/310#issuecomment-391048414. I think I've addressed most of the initial feedback.

This is still targeting `1.0`. After `1.0` is merged into `master` this should be ok to merge into `master`.

ggetz:
Looks good, thanks @lilleyse ! Did we want to merge this into 1.0 or wait until 1.0 is in master? I think it should be fine to merge into 1.0 since it's an extension.

lilleyse:
I would wait to merge this into `master`.

lilleyse:
@ggetz I retargeted this to `master`. This should be ready to go now.

pjcozzi:
What is the plan with this PR?  If this goes into `master`, there should be a separate `ogc-community-standard-1.0` branch unless this was included in the Community Standard submission.  Perhaps that is the best approach compared to merging this into a `3d-tiles-next` branch.

lilleyse:
I think the plan was to merge into master still. We have a [1.0 tag](https://github.com/AnalyticalGraphicsInc/3d-tiles/releases/tag/1.0) that marks when the OGC community standard submission was released which can be converted to a branch if needed.

ggetz:
The 1.0 tag is also what we linked to from the OGC documents. So I think this should be fine to go into `master`. Otherwise changes here look good to me and this should be ready to go.

pjcozzi:
Please make sure there is a clear link to the OGC version from the main link.

Also, how will the tag work and bug fixes go into the OGC version?

ggetz:
I can add a link. We can branch off the tag (now or later), and we can update the tag as needed.

PR #325, opened 2018-07-06
EPSG Fix

ggetz:
Fixes https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/324

[EPSG:4979](http://spatialreference.org/ref/epsg/4979/) includes height above the ellipsoid, whereas 4326 does not. Will update accordingly.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -77,11 +77,11 @@ Acknowledgements:
 
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


lilleyse:
Also update `boundingVolume.schema.json` and `tile.schema.json`. Otherwise looks good.

ggetz:
Thanks @lilleyse, updated!

lilleyse:
👍

PR #326, opened 2018-07-08
Remove tile transform restriction on regions

lilleyse:
Removed the restriction that states a tile transform can't be used on the `region` bounding volume type. An implementation can get around the coordinate system problem by converting the regions to boxes at runtime, like in https://github.com/AnalyticalGraphicsInc/cesium/pull/6755. I wasn't sure if that strategy should be mentioned in the spec or left open.

This is targeting the 1.0 branch but can be retargeted if needed.


lilleyse:
Closing.

After some thinking I decided that this restriction is actually ok and makes the most sense. Engines can handle additional transforms however they want, whether that means moving the region or not.

PR #327, opened 2018-07-31
Merge 1.0 branch into master

lilleyse:
The updates here have already been reviewed, mostly in https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/308. The `1.0` changes should be ready to merge into `master` now.

Right after merging:
* [ ] Fix broken specification links. I did a search for `3d-tiles/tree/master/` and `3d-tiles/tree/1.0/` and these projects had links:
    * [ ] cesium
    * [ ] cesium.com
    * [ ] cesiumjs.org

@pjcozzi @ggetz is there anything left to add here?


ggetz:
List looks good.

Maybe a final copy edit if @slchow has the time, but I don't think it's necessary in order to merge this.

ggetz:
Only other thing I can think of is if we want to park current master in a branch because it has the start of some additional tile formats like VectorData, unless we want to carry over those changes here.

ggetz:
Disregard my above comment, I see the `next` branch and https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/328

lilleyse:
If not a branch, maybe a tag for the current master so we can reference it later as needed.

lilleyse:
I pushed a `0.0` tag.

slchow:
I could copyedit, but it would realistically be the end of next week at least before I could get to it. @lilleyse Do you want me to look it over?

lilleyse:
I think it's fine to merge without a copyedit.

ggetz:
@lilleyse I think we should also merge https://github.com/AnalyticalGraphicsInc/3d-tiles/pull/329 before merging this.

ggetz:
Fixed a few typos and published a new PDF. I think this should be ready to go.

lilleyse:
Otherwise looks great!

ggetz:
Thanks @lilleyse! Made the fixes in the PDF, and now reference the tag instead of the branch.

I don't see any `1.0` tag however, should we create once we merge this PR?

lilleyse:
Oh I was mixing up the tags...

But yeah, keep the `0.0` tag as-is and add a `1.0` rag right after merging this, and point to the `1.0` tag in the PDF (the link in the PDF should be `https://github.com/AnalyticalGraphicsInc/3d-tiles/releases/tag/1.0`).

It also looks like some files got added accidentally to the last commit.

ggetz:
Sounds good :+1: 

Whoops, removed the extra files, and updated.

lilleyse:
In addition to creating the tag, could you also fix the now-broken spec links mentioned in the PR description?

That looks like everything. Thanks @ggetz!

PR #345, opened 2018-10-12
Remove number array constraint from batch table schema

lilleyse:
Fixes https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/344

@ggetz can you review?


ggetz:
Looks good, thanks @lilleyse! This is already reflected in the property reference as well: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification/TileFormats/BatchTable#property

PR #354, opened 2018-11-14
Reorganize glTF transforms section

ggetz:
@lilleyse

Reorganize all glTF references to be after glTF is introduced.

### DIFF
--- a/Q-and-A.md
+++ b/Q-and-A.md
@@ -1,6 +1,6 @@
 # 3D Tiles Q&A
 
-## Contents 
+## Contents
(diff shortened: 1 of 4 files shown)


lilleyse:
Pushed a couple small tweaks.

Does glTF belong under the tile transforms section though? Maybe instead the table of contents should be organized like:

* Transforms
    * Tile transforms
    * glTF transforms
    * Example

ggetz:
Thanks @lilleyse, that makes sense. Updated!

PR #355, opened 2018-11-14
CRS clarifications

ggetz:
Specify EPSG and don't introduce `tile.transform` before its defined.

@lilleyse


specification/README.md
```diff
@@ -133,9 +133,9 @@ All angles are in radians.
...
+3D Tiles uses a right-handed Cartesian coordinate system; that is, the cross product of _x_ and _y_ yields _z_. 3D Tiles defines the _z_ axis as up for local Cartesian coordinate systems. A tileset's global coordinate system will often be in a [WGS 84](http://earth-info.nga.mil/GandG/publications/tr8350.2/wgs84fin.pdf) earth-centered, earth-fixed (ECEF) reference frame ([EPSG 4979](http://spatialreference.org/ref/epsg/4979/)), but it doesn't have to be, e.g., a power plant may be defined fully in its local coordinate system for use with a modeling tool without a geospatial context.
 
-An additional [tile transform](#tile-transform) may be applied to transform a tile's local coordinate system to the parent tile's coordinate system.
+It's possible to transform a tile's local coordinate system to the parent tile's coordinate system, see [tile transform](#tile-transform).
```
lilleyse: I don't know if this change is worth making. The original sentence reads more clearly to me.

ggetz: The issue is that we're referencing the tile transform property before it's introduced.

ggetz:
@lilleyse updated!

PR #356, opened 2018-11-15
Remove tiles3d_tileset_time

### LINKED ISSUE #321: Remove tiles3d_tileset_time
[tileset_3d_tileset_time](https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/Styling#tiles3d_tileset_time) is a built-in variable of the styling language that stores the current time, used to make time dynamic styles.

But I think it's actually better to remove this and use [mutables](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/252) instead.

Not urgent or anything, just a thought I had.

lilleyse:
Fixes #321

We want to go further with time-dynamic and analytics-based styling in [3D Tiles Next](https://github.com/AnalyticalGraphicsInc/3d-tiles/issues/247), and until we know what that looks like it's better to remove this property.

### DIFF
--- a/specification/Styling/README.md
+++ b/specification/Styling/README.md
@@ -23,7 +23,6 @@
    * [String conversion](#string-conversions)
    * [Constants](#constants)
    * [Variables](#variables)
-   * [Built-in variables](#built-in-variables)
    * [Built-in functions](#built-in-functions)
    * [Notes](#notes)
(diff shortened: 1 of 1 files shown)


ggetz:
Thanks @lilleyse! I agree this is probably best tackled when we have a better idea of what time-dynamic styling looks like holistically.

PR #376, opened 2019-05-28
fix example for Batched3DModel _BATCHID attribute

likangning93:
The gltf 2.0 specification doesn't allow `UNSIGNED_INT` accessors for general vertex attributes: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#accessorcomponenttype-white_check_mark

This PR updates the example to use `FLOAT` for the batch IDs accessor example instead.


lilleyse:
Good catch @likangning93

PR #395, opened 2019-11-08
Fix schema link

erikdahlstrom:


lilleyse:
Hey Erik! Thanks for the fix.

PR #400, opened 2020-01-17
Fixes incorrect batch table figure.

### LINKED ISSUE #398: 3DTILES_batch_table_hierarchy example inconsistencies
I'm struggling to understand the example provided at the end of the 3DTILES_batch_table_hierarchy spec. A few things are confusing me.

batch-table-hierarchy-block

1. Shouldn't batchId 3 map to the 4th element in the classIds array? In the car example batch Id 5 maps to the 5th element in the classIds array.
1. In the building example, the third classId is also the third occurrence of id `0`, however it maps to the 4th instance in wall class. In the "car" example the third occurrence of class ID 1, maps to the third instance. Which is correct, or am I missing a step in how these are mapped?
1. In the building example, how can batchId 2 and 3 reference the same building? They can't be the second occurrence of class 1.

Thanks in advance for your help clarifying these issues.

sanjeetsuhag:
Fixes #398


lilleyse:
Great, thanks @sanjeetsuhag!

PR #405, opened 2020-03-12
Add 3D Tiles Overview

shehzan10:
This PR adds:
* The 3D Tiles Overview PDF
* The source files for the overview
* The Windows batch files that convert the SVGs+Images to PDF
* 3D Tiles Overview README + Changelog

@lilleyse Can you please review?


chris-little:
The PDF is really helpful in understanding. However, perhaps a little clarification for an outsider: 
Can tilesets be nested hierarchically? I'm not sure from the text.
Can a tile can be "in" more than one tileset? I'm not sure how, but presumbly URIs allow this.

lilleyse:
@chris-little Yes tilesets can be nested hierarchically. This is mentioned at the bottom of page 2 but it could be more prominent. Also a tile can be referenced by more than one tileset.

We'll open an issue to address these points. And let us know if anything else is confusing. An outsider's perspective is always welcome!

lilleyse:
Thanks. The overview looks great!

lilleyse:
And big thank you to @javagl for creating the overview!

javagl:
Thanks to everybody who provided support, feedback, and reviews! (<- There's the Oxford comma ;-))

> Also a tile can be referenced by more than one tileset.

It may be a nitpick, but in one way, "a tile" is just a JSON block that is nested in a tileset JSON. Whether or not two structurally equal JSON blocks can be called "the same tile" (regardless of the context in which they appear) may be a matter of definitions...

(Of course, the tile *content* can be referred to multiple times - maybe this is what the question was about? One could also put a single tile into a tileset and use this as some sort of proxy to "refer to the same tile from multiple tilesets", but I'm not sure whether this would make sense in general...)

lilleyse:
> Of course, the tile *content* can be referred to multiple times - maybe this is what the question was about?

@javagl yeah that was my understanding of the question. It doesn't help that tile and content are often used interchangeably.

chris-little:
@javagl Thank you. I was trying to understand from the conceptual viewpoint rather than a hard implementation perspective. I am a Platonist rather than Aristotleian and tend to think Containers/Content and Classes/Instances ;-)

PR #410, opened 2020-04-08
Add first draft of `3DTILES_extent` spec

Samulus:
This is a WIP for a PR that adds 3DTILES_extent support to the 3dtiles spec to support various clipping uses cases.

Direct link to extension: https://github.com/CesiumGS/3d-tiles/tree/3DTILES_extent/extensions/3DTILES_extent


extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
+# 3DTILES_extent
+
+## Contributors
+
+* Sean Lilley, [@lilleyse](https://github.com/lilleyse)
+* Samuel Vargas [@Samulus](https://github.com/Samulus)
```
lilleyse, line 6: You can flip the order and make yourself the first author here

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+* [Overview](#overview)
+* [Defining Extents](#defining-extents)
+* [Holes in Extents](#holes-in-extents)
+* [Arc Types](#arc-types)
+* [Coordinate System](#coordinate-system)
+
+## Overview
+
+The extent 3dtiles extension allows the user to annotate the existence of a smaller 2D region(s) (an extent) in a given `tileset.json` via a 2D array of latitude, longitude pairs. This is useful for a variety of scenarios such as: Overlaying high resolution geometry ontop of low level geometry, insetting one tileset into another tileset, or clipping excess geometry inside of a provided tileset.
```
lilleyse, line 18: `The extent 3dtiles extension allows` - maybe just `This extension allows`

lilleyse, line 18: `the existence of a smaller 2D region(s)` - any region is allowed, it doesn't have to be smaller

lilleyse, line 18: Throughout, change `tileset.json` to `tileset JSON`. The rest of the 3D Tiles spec uses this form since the tileset JSON doesn't have to be called `tileset.json`.

lilleyse, line 18: From offline discussion, change latitude / longitude order to longitude / latitude throughout and update the examples.

Maybe borrow from [RFC 7946](https://tools.ietf.org/html/rfc7946):

> The first two elements are longitude and latitude, or
> easting and northing, precisely in that order and using decimal
> numbers. Altitude or elevation MAY be included as an optional third
> element.
>

lilleyse, line 18: > This is useful for a variety of scenarios such as: Overlaying high resolution geometry ontop of low level geometry, insetting one tileset into another tileset, or clipping excess geometry inside of a provided tileset.

The first two scenarios sound similar, perhaps combine into something like "Insetting a high resolution tileset into a lower resolution tileset". Maybe add another scenario like "Defining a 2D collision boundary". E.g:

> This is useful for a variety of scenarios such as: Insetting a high resolution tileset into a lower resolution tileset, defining a 2D collision boundary, or clipping excess geometry inside of a provided tileset.

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Defining Extents
+
+An extent is a collection of latitude and longitude coordinate pairs. Extents are two dimensional in nature, but an optional third component can be specified for **each** extent coordinate to specify its height (in meters). The coordinate pairs should be provided in **counterclockwise** winding order. Multiple extents can be specified. Convex and concave extents are both supported. At least three coordinates must be provided for an extent to be valid.
```
lilleyse, line 22: `its height (in meters)` - more formally `its height (in meters) from the WGS84 ellipsoid`

lilleyse, line 22: > Multiple extents can be specified

Mention that overlapping extents should be treated as a boolean union. (Fourth image [here](https://en.wikipedia.org/wiki/File:Boolean_operations_on_shapes.png))

lilleyse, line 22: Hm we should probably disallow self-intersecting polygons.

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+            "holes": [
+                [32.2032301338380, -116.61521629247]
+                [32.3482030238302, -116.323208320832]
+                [32.7382033280322, -116.4830243804380]
+            ]
+          }
+        ]
+      }
+    }
+```
```
lilleyse, line 126: This example is missing a closing curly brace

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+              [32.514456825587125, -116.61385130521423],
+              [32.515638616140173, -116.61829305963617],
+              [32.513154268152434, -116.61944030086889],
+              [32.511166344475825, -116.61512629247555]
+            ],
+            "holes": [
+                [32.2032301338380, -116.61521629247]
+                [32.3482030238302, -116.323208320832]
+                [32.7382033280322, -116.4830243804380]
+            ]
```
lilleyse, line 121: I think this would only let you defined a single hole.

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+    }
+  },
+  "geometricError": 100.0,
+  "root": {...}
+}
+```
+
+## Holes in Extents
+
+Holes are also supported, simpliy provide a `holes` object for a given polygon, e.g:
```
lilleyse, line 102: Should probably be stated here as well that overlapping holes are treated as a boolean union.

extensions/3DTILES_extent/README.md
```diff
@@ -0,0 +1,136 @@
...
+                [32.3482030238302, -116.323208320832]
+                [32.7382033280322, -116.4830243804380]
+            ]
+          }
+        ]
+      }
+    }
+```
+
+Holes should be provided in **counterclockwise** winding order, and at least three coordinates must be provided.
```
lilleyse, line 128: Like above, disallow self-intersection.

extensions/3DTILES_extent/README.md
Samulus: It's a 3D array now.

lilleyse:
Superseded by https://github.com/CesiumGS/3d-tiles/pull/424

Issue #411, opened 2020-05-06
Consider clarifying when ${feature['identifier']} should be used when styling


OmarShehata:
The styling section talks about two ways of referencing a feature by its name:

> Variables are used to retrieve the property values of individual features in a tileset. Variables are identified using the ES 6 (ECMAScript 2015) template literal syntax, i.e., ${feature.identifier} or ${feature['identifier']}, where the identifier is the case-sensitive property name. feature is implicit and can be omitted in most cases.

From: https://github.com/CesiumGS/3d-tiles/tree/master/specification/Styling#variables

One issue I recently ran into is that my identifier had spaces in the name, or characters like `:`, so I wasn't able to style it. It may be worth either clarifying that those are the cases when you cannot omit the "feature" part.

PR #412, opened 2020-05-06
Variable identifier clarifications

lilleyse:
Fixes https://github.com/CesiumGS/3d-tiles/issues/411

Adds a couple clarifications to the styling spec

* Strings are encoded in UTF-8
* Describes when to use the `${feature['identifier']}` syntax

CC @OmarShehata @mramato


OmarShehata:
This change looks good to me.

PR #413, opened 2020-06-04
Adds 3DTILES_tile_metadata extension

sanjeetsuhag:
This extension enables storage of per tile metadata. Specification is available [here](https://github.com/CesiumGS/3d-tiles/blob/3DTILES_tile_metadata/extensions/3DTILES_tile_metadata/README.md).

PR #414, opened 2020-06-04
Adds 3DTILES_implicit_tiling extension

sanjeetsuhag:
This extension enables 3D Tiles to better support implicit tiling schemes. Specification is available [here](https://github.com/CesiumGS/3d-tiles/tree/3DTILES_implicit_tiling/extensions/3DTILES_implicit_tiling).


pjcozzi:
@sanjeetsuhag congrats on this!  Please bump to me when it is ready for me to review. 😄 

CC @lilleyse

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**boundingVolume**|`object`|A bounding volume that encloses the tileset.|☑️ Yes|
+|**tilingScheme**|`string`|A string describing the tiling scheme used within the tileset|☑️ Yes|
+|**subdivision**|`object`|An object containing high level information about the subdivision buffer|☑️ Yes|
+|**content**|`object`|An object containing high level information about the content buffer. This may be omitted if no tiles in the tileset contain content.|No|
+|**metadata**|`object`|An object containing high level information about the metadata buffer. This may be omitted if no tiles in the tileset contain metadata.|No|
+|**bufferViews**|`array`|An array containing typed views into buffers|No|
+|**buffers**|`array`|An array of buffers.|No|
+|**tileExtension**|`string`|The extension applied to each tile in the tileset.|No|
```
lilleyse, line 202: Rename to `contentExtension`?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+|------|-------------------|
+| `0`  | No metadata       |
+| `1`  | Has metadata      |
+
+
+When the tile has content, the index of the tile in the subdivision buffer will be the index of corresponding metadata property in the associated buffer in the `3DTILES_tile_metadata` extension.
+
+#### Level Offsets
+
+In the content and metadata objects, a level offset and a default value can be specified to implicitly apply the default value to each tile in all levels before the specified level offset. Similar to
```
lilleyse, line 139: Similar to ... ?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+    },
+    "metadata": {
+        "levelOffset": 2,
+        "levelOffsetFill": 0,
+        "bufferView": 0
+    }
+}
+```
+
+Only the information with grey background is present in the bitstream.
```
lilleyse, line 172: Personal preference, but maybe swap the colors. I think of grayed out as default / not present.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+![States](figures/states.png)
+
+### Root Tiles
+
+Multiple tilesets using this extension can be combined by using parent tileset referring to one or more external tilesets using `3DTILES_implicit_tiling` as the children of the root tile.
```
lilleyse, line 178: I would introduce this first by describing the use case - some situations require there to be multiple root tilesets - e.g. double-headed quadtree like TMS, six-headed quadtree like S2

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+|**tilesetExtension**|`string`|The extension applied to each implicit external tileset in the tileset.|No|
+
+Additional properties are not allowed.
+
+---
+
+
+### boundingVolume 
+
+A bounding volume that encloses the tileset.  Exactly one `box`, `region` or `cell` property is required.
```
lilleyse, line 212: These sections should describe *how* the bounding volume gets subdivided for octree and quadtree alongside the diagrams.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+An array of 12 numbers that define an oriented bounding box. The first three elements define the x, y, and z values for the center of the box. The next three elements (with indices 3, 4, and 5) define the x axis direction and half-length.
+The next three elements (indices 6, 7, and 8) define the y axis direction and half-length. The last three elements (indices 9, 10, and 11) define the z axis direction and half-length.
+
+* **Type**: `number` `[12]`
+* **Required**: No
+
+| Box | Quadtree | Octree |
+|:---:|:--:|:--:|
+| ![](figures/box_one.png) | ![](figures/box_quadtree_one.png) | ![](figures/box_octree_one.png)  |
```
lilleyse, line 234: Throughout the diagrams are ok, but could use refinement in the next draft.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+![States](figures/states.png)
+
+### Root Tiles
+
+Multiple tilesets using this extension can be combined by using parent tileset referring to one or more external tilesets using `3DTILES_implicit_tiling` as the children of the root tile.
+
+### Jump Buffer
+
+The jump buffer is a buffer created at runtime to allow efficient traversal and tile lookup functions. The jump buffer is a bijection between a tile's Morton index and its index in the level.
```
lilleyse, line 182: Anything jump buffer related should be non-normative.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+            if (sBuffer[sBufferIndex++] === 1) {
+                jBuffer[currentLevel].push(jBuffer[currentLevel - 1][Math.floor(i / tilingScheme)] * tilingScheme + (i % tilingScheme));
+            }
+        }
+    }
+    return jBuffer;
+}
+```
+
+#### Tile Index Lookup
```
lilleyse, line 327: Is there a fast path for complete trees? Would be good to show off true random access.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+### 3DTILES_implicit_tiling Tileset JSON extension
+
+Specifies the Tileset JSON properties for the 3DTILES_implicit_tiling.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**boundingVolume**|`object`|A bounding volume that encloses the tileset.|☑️ Yes|
+|**tilingScheme**|`string`|A string describing the tiling scheme used within the tileset|☑️ Yes|
```
lilleyse, line 196: List all the possible tiling schemes here.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+### content
+
+Provides information about the content of the tileset.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**levelOffset**|`number`|An integer describing which level the content bitsream, if provided, starts at|No|
+|**levelOffsetFill**|`number`|An integer describing the state value for the content bit for all levels up to the `levelOffset`.|No|
```
lilleyse, line 283: What's the default value if `levelOffset` is provided but `levelOffsetFill` is not? I think `0` makes sense.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+### metadata
+
+Provides information about the metadata of the tileset.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**levelOffset**|`number`|An integer describing which level the metadata bitsream, if provided, starts at|No|
+|**levelOffsetFill**|`number`|An integer describing the state value for the metadata bit for all levels up to the `levelOffset`.|No|
```
lilleyse, line 297: Same comment as above. `0` makes sense as the default value.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+|**tilingScheme**|`string`|A string describing the tiling scheme used within the tileset|☑️ Yes|
+|**subdivision**|`object`|An object containing high level information about the subdivision buffer|☑️ Yes|
+|**content**|`object`|An object containing high level information about the content buffer. This may be omitted if no tiles in the tileset contain content.|No|
+|**metadata**|`object`|An object containing high level information about the metadata buffer. This may be omitted if no tiles in the tileset contain metadata.|No|
+|**bufferViews**|`array`|An array containing typed views into buffers|No|
+|**buffers**|`array`|An array of buffers.|No|
+|**tileExtension**|`string`|The extension applied to each tile in the tileset.|No|
+|**tilesetExtension**|`string`|The extension applied to each implicit external tileset in the tileset.|No|
+
+Additional properties are not allowed.
```
lilleyse, line 205: Might change if we allows `extras` or `extensions` on this object

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+A bounding volume that encloses the tileset.  Exactly one `box`, `region` or `cell` property is required.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**box**|`number` `[12]`|An array of 12 numbers that define an oriented bounding box. The first three elements define the x, y, and z values for the center of the box.  The next three elements (with indices 3, 4, and 5) define the x axis direction and half-length.  The next three elements (indices 6, 7, and 8) define the y axis direction and half-length.  The last three elements (indices 9, 10, and 11) define the z axis direction and half-length.|No|
+|**region**|`number` `[6]`|An array of six numbers that define a bounding geographic region in EPSG:4979 coordinates with the order [west, south, east, north, minimum height, maximum height]. Longitudes and latitudes are in radians, and heights are in meters above (or below) the WGS84 ellipsoid.|No|
+|**cell**|`number` `[3]`|An array of 10 numbers that defines a geodesic quadrilateral cell. The first 8 elements are 4 pairs of EPSG:4979 coordinates in radians, and the last 2 elements are heights in meters above the WGS84 ellipsoid.|No|
```
lilleyse, line 220: Is there a better name for `cell`?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+### Samples
+
+#### Complete Quadtree
+
+##### Base Tileset - tileset.json
+
+```json
+{
+    "asset": {
+        "version": "2.0.0-alpha.0"
```
lilleyse, line 373: For all the examples change this back to `1.0` since we haven't introduced any breaking changes

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
+# 3DTILES_implicit_tiling Extension
```
lilleyse, line 1: There should be some mention of how `3DTILES_implicit_tiling` and a tileset's `root` property interact.

lilleyse, line 1: From offline discussion, there should be a section about how this extension interacts with [`3DTILES_tile_metadata`](https://github.com/CesiumGS/3d-tiles/pull/413)

extensions/3DTILES_implicit_tiling/README.md
lilleyse, line 220: Offline we talked about `cell` being allowed in the general `boundingVolume` schema (as long as `3DTILES_implicit_tiling` is preset) so that base tilesets can also use `cell` bounding volumes. Is that still the intent?

extensions/3DTILES_implicit_tiling/README.md
lilleyse, line 1: Like https://github.com/CesiumGS/glTF/pull/1, all sample tilesets that we send from here on out should have an `extras.draftVersion` that follows semver and distinguishes it from future revisions.

```
"extensions": {
  "3DTILES_implicit_tiling": {
    "extras": {
      "draftVersion": "0.0.0"
    },
    ...
  }
}
```

lilleyse: Just to clarify, no need to edit the examples in the spec, just make sure to include in sample tilesets we send out.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+| Code | Description       |
+|------|-------------------|
+| `0`  | No content        |
+| `1`  | Has content       |
+
+When the tile has content, the content payload can be found implicitly by combining the tile location with the tile extension.
+
+#### Metadata
+
+This is a one bit representation of whether or not a tile has content associated with it.
```
lilleyse, line 127:
```suggestion
This is a one bit representation of whether or not a tile has metadata associated with it.
```

extensions/3DTILES_implicit_tiling/README.md
lilleyse: Never mind, I see the section talking about that now.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+| Code | Description       |
+|------|-------------------|
+| `0`  | No metadata       |
+| `1`  | Has metadata      |
+
+
+When the tile has content, the index of the tile in the subdivision buffer will be the index of corresponding metadata property in the associated buffer in the `3DTILES_tile_metadata` extension.
```
lilleyse, line 135: This doesn't leave room for empty tiles having metadata, which is something worth supporting. In that case the index would be the number of tiles with metadata before it.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+* **Type**: `number` `[6]`
+* **Required**: No
+
+| Region | Quadtree | Octree |
+|:---:|:--:|:--:|
+| ![](figures/region.png) | ![](figures/region_quadtree.png) | ![](figures/region_octree.png)  |
+
+#### boundingVolume.cell
+
+An array of 10 numbers that defines a geodesic quadrilateral cell. The first 8 elements are 4 pairs of EPSG:4979 coordinates in degrees in counterclockwise order, and the last 2 elements are heights in meters above the WGS84 ellipsoid.
```
lilleyse, line 249: Are certain configurations not allowed? Like concave shapes?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+### subdivision
+
+Provides information about the subdivision of the tileset.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**completeLevels**|`number`|An integer describing how many levels subdivide internally. We can assume the subdivision bitcode for all tiles on levels before this one to be `11`.|No|
+|**bufferView**|`number`|An index to the buffer view containing the subdivision buffer.|No|
```
lilleyse, line 270: What happens if `bufferView` is not defined and `completeLevels` is not `maximumLevels`?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+## Properties Reference
+
+---------------------------------------
+### 3DTILES_implicit_tiling Tileset JSON extension
+
+Specifies the Tileset JSON properties for the 3DTILES_implicit_tiling.
+
+**Properties**
+
+|   |Type|Description|Required|
```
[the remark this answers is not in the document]
lilleyse: Also... would need to be careful to distinguish this from the maximum level of the whole tree of tilesets. Do we need to store a separate property for that?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+Provides information about the content of the tileset.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**levelOffset**|`number`|An integer describing which level the content bitsream, if provided, starts at|No|
+|**levelOffsetFill**|`number`|An integer describing the state value for the content bit for all levels up to the `levelOffset`.|No|
+|**bufferView**|`number`|An index to the buffer view containing the content buffer.|No|
```
lilleyse, line 284: Same comment as above. What's the behavior if `bufferView` is not defined?

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+| Cell | Quadtree | Octree |
+|:---:|:--:|:--:|
+| ![](figures/cell.png) | ![](figures/cell_quadtree.png) | ![](figures/cell_octree.png)  |
+
+
+---
+
+### subdivision
+
+Provides information about the subdivision of the tileset.
```
lilleyse, line 263: The descriptions for `subdivision`, `content`, and `metadata` could be more concrete.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+```
+
+
+
+
+### Samples
+
+#### Complete Quadtree
+
+##### Base Tileset - tileset.json
```
lilleyse, line 368: As we talked about offline, let's present the examples like this:

1. Complete quadtree as a single tileset
2. Double-headed quadtree as two separate tilesets
3. Double-headed quadtree with a base explicit tileset
4. Six-headed S2-like example with base explicit tileset

lilleyse, line 368: The "Base Tileset" terminology should be reserved for base explicit tilesets just so it doesn't lead to any confusion

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+*Note: Directory structure is not complete. Only for demonstration purposes.*
+ 
+```
+.
+├── tileset.json
+└── R0/
+    ├── tileset.json
+    └── 0/
+        ├── tile.glb
+        ├── 0/
```
lilleyse, line 518: Something doesn't seem right about this... all `.glb` tiles should be four layers deep `R0/level/x/y.glb`.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Only the tilesets that use a base tileset now use that term. Combination strategies are mentioned in the title to help clarify the concept.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,605 @@
...
+
+#### Location in space
+
+![Morton](figures/morton.png)
+
+Level grids are indexed from the bottom left, using a right handed coordinate system, with +Z pointing in the up direction.
+
+Tiles in a level are indexed by applying the [Morton/Z-order](https://en.wikipedia.org/wiki/Z-order_curve) curve to the grid at that level. Using the Morton order serves three primary purposes:
+
+- increasing spatial locality of reference: Tiles that are close to each in other in space will be located close to each other on the file system.
```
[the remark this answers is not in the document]
sanjeetsuhag: All examples now use the Level structure to store tiles, instead of the Hierarchy structure.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+Tiles in a level are indexed by applying the [Morton/Z-order](https://en.wikipedia.org/wiki/Z-order_curve) curve to the grid at that level. Using the Morton order serves three primary purposes:
+
+- Increasing spatial locality of reference: Tiles that are close to each in other in space will be located close to each other on the file system.
+- Efficient tile location decomposition: The Morton order allows efficient encoding and decoding of locations of a tile in the level grid to it's location in the availability bitstream
+- Efficient traversal: The binary representation of tile locations in the grid allow for easy traversal of the tileset.
+
+#### Location on disk
+
+Tiles are located in the file system according to their position in the tileset hierarchy. For example, in a tileset using the quadtree tiling scheme, the 3rd tile at Level 2, which is a child of the 1st tile at Level 1, will be located at `0/1/3`.
```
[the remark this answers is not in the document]
sanjeetsuhag: This line has been removed.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+| Bitstream | Description       | Size (bits) |
+|------|-------------------|---|
+|  subdivision  | Encodes subdivision of each tile | 2 |
+|  content  | Encodes availability of content for tile | 1 |
+|  metadata  | Encodes availability of metadata for tile | 1 |
+
+#### Subdivision
+
+The subdivision state for each tile determines if and how it subdivides into children tiles, as per the `tilingScheme`. The subdivision state is encoded in 2 bits, and padded with 0s at the end to meet byte boundaries. The jump buffer generation will ignore the padding. Subdivision in a tile can have one of the following states:
```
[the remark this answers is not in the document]
sanjeetsuhag: - All buffers are tightly packed. Any padding will be added to the end.
 - Jump buffer description and algorithm have been moved to Appendix.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+An external tileset may exist within the file structure of its parent tileset, with the root of the external tileset being present at the implicit location of the tile with subdivision state `01`.
+
+##### External Tileset at Implicit Location
+
+An external tileset may exist outside the file structure of its parent tileset, with the root of the external tileset being present in the `tile.json` at the implicit location of the tile with subdivision state `10`.
+
+##### Complete Levels
+
+For tilesets that have uniform subdivision for each tile up to a certain level, it is redundant to store the tile subdivision state information. The `completeLevels` property is used to indicate how many levels within the tileset are complete. Complete levels are levels in which all tiles subdivide implicitly and internally (using the bitcode `11`).
```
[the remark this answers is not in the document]
sanjeetsuhag: This functionality will be different based on the nature of the buffer. The true binary nature of the availability buffers allow for a succinct way of knowing default value across levels with uniform state.

- `levelOffset` and `levelOffsetFill` can fill the tree with state `X`.
  - Specifying a `bufferView` implies explicit definition of state for all tiles on levels beyond the `levelOffset`.
  - Not specifying a `bufferView` implies the state `X'` will apply to all tiles on levels beyond the `levelOffset`.

The case is not so simple for the subdivision buffer, and for the current version of the spec, for levels with uniform state, compression should handle it well.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Updated spec.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Started a thought, never finished it. 😅

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Colors have been updated, with dashed boundaries for implicit values.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added mentions of TMS/quantized-mesh/S2 configuration in Root Tiles section.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added description of how split axes get defined.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Moved jump buffer to Appendix.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added description of `tilingScheme` in property reference.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: The default value is `0` and it is now mentioned in property reference.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: The default value is `0` and it is now mentioned in property reference.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Based on offline discussion, the current name is `geodesicQuad`.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Reverted explicit tileset version to `1.0`.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added rule in Root Tiles section.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: At the moment, it's only a part of the extension since we may decide to change the way this whole concept works in order to be a better fit for S2.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
sanjeetsuhag: No, that'll be added to the next version, primarily because it seems like we'll be changing the way bounding volumes get divided.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
sanjeetsuhag: Added refine property.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added examples along with directory structure.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Updated all examples to have directory structure.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added more description.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added this to property description.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added this to property description.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
sanjeetsuhag: Implicit tiling has strict bounding volume subdivision, the building would get cut in half.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
sanjeetsuhag: Added a `maximumLevel` property for highest level of **internal** subdivision.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: This was the result of a typo. Content and metadata are independent concepts.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Added a note about that, but next draft will have a more fleshed out algorithm.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+#### Content
+
+This is a one bit representation of whether or not a tile has content associated with it.
+
+| Code | Description       |
+|------|-------------------|
+| `0`  | No content        |
+| `1`  | Has content       |
```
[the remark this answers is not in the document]
sanjeetsuhag: The rule stated here says that it is ignored. I don't know if it needs to be "disallowed" or not. `https://github.com/CesiumGS/3d-tiles/blame/3DTILES_implicit_tiling/extensions/3DTILES_implicit_tiling/README.md#L106`

extensions/3DTILES_implicit_tiling/README.md
lilleyse: Let's make all examples `1.0` since we're going with the `1.0` + extensions approach rather than jumping to `2.0.0-alpha.0`

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
lilleyse: Ok, that's how I read it right now too. I think we should consider this case some more for the next draft so I'm going to mark it back to unresolved.

sanjeetsuhag:
@lilleyse Following is a list of to-do items for next draft:

 - [ ] [Equally sized tiles](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435486368) - The latest spec states that during subdivision, the total space occupied by the tile remains the same, but the number of tiles multiply (by 4 for quadtree, by 8 for octree). However, there is no guarantee of tiles of "equal size" since that is not the case for the `region` and `geodesicQuad` bounding volume types. This will change with the development of a better heuristic of where the split axes get defined for the bounding volumes that are based on ellipsoidal geometry.
- [ ] [Binary storage of external tileset URIs](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435496715) - The latest spec assumes the presence of a `tile.json` at the location of a tile that subdivides externally. Storing the URIs in a binary buffer will be added following the maturity of the `3DTILES_binary_buffers` extension.
- [ ] [Adding property to indicate uniform external subdivision](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435597753) - The latest spec does not include an additional property to store uniform state across because uniform values compress well and more research is needed to see the various state configurations this could lead to. Since it would save on storage and add a stage of indirection, such a property may be added in the next draft, however, it may include an enum for describing the tile state instead.
- [ ] [Better diagrams for tiling schemes across bounding volume types](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435600971) -  These diagrams can be improved quite a lot, and the next draft should provide more information about the split axes and a better color scheme.
- [ ] [Content/metadata overlap on tiles that subdivide externally](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r436966251) - A runtime implementation of the current spec will choose to ignore the content and metadata of a tile that subdivides externally, but it is not currently disallowed because it may be necessary in case of an explicit external tileset but an error in case of an implicit external tileset (due to overlap). The next version of the spec will have a better way of handling this case.
- [ ] [Improved description of explicit and implicit external tilesets](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435646312) -  A diagram along with an example may be needed to help illustrate this concept better than words. Algorithms need to also include how to handle these cases.
- [ ] [Order of description of geodesicQuad](https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r435652818) - This whole concept will receive an overhaul since we might have to place restriction on it's alignment in order to develop a better heuristic for determining split axes to yield tiles that are similar in size.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+
+Specifies the Tileset JSON properties for the 3DTILES_implicit_tiling.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**boundingVolume**|`object`|A bounding volume that encloses the tileset.|☑️ Yes|
+|**tilingScheme**|`string`|A string describing the tiling scheme used within the tileset.|☑️ Yes|
+|**refine**|`string`|A string to indicate if additive or replacement refinement is used when traversing the tileset for rendering. This property will apply to all available levels in the tileset.|☑️ Yes|
```
lilleyse, line 209: Update this description to apply to implicit tiling

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+### subdivision
+
+Provides information about the spatial hierarchy of the tileset. The structure of the tileset may be specified through the `completeLevels` property for tilesets that subdivide completely to a certain level. For all levels between the `completeLevels` and the `maximumLevel`, the subdivision bitstream is available in the `bufferView`. If a tileset is complete, the `bufferView` may be omitted.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**completeLevels**|`number`|An integer describing how many levels subdivide internally. The default value for the subdivision state for all tiles on complete levels is `11`.|No, the default value is 0.|
+|**maximumLevel**|`number`|An integer describing the highest level of internal subdivision.|No, default value is 1.|
```
lilleyse, line 311: I think `maximumLevel` should be a required property.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**boundingVolume**|`object`|A bounding volume that encloses the tileset.|☑️ Yes|
+|**tilingScheme**|`string`|A string describing the tiling scheme used within the tileset.|☑️ Yes|
+|**refine**|`string`|A string to indicate if additive or replacement refinement is used when traversing the tileset for rendering. This property will apply to all available levels in the tileset.|☑️ Yes|
+|**subdivision**|`object`|An object containing high level information about the subdivision buffer|☑️ Yes|
+|**content**|`object`|An object containing high level information about the content buffer. This may be omitted if no tiles in the tileset contain content.|No|
+|**metadata**|`object`|An object containing high level information about the metadata buffer. This may be omitted if no tiles in the tileset contain metadata.|No|
+|**contentExtension**|`string`|The extension applied to each tile in the tileset.|No|
+|**tilesetExtension**|`string`|The extension applied to each implicit external tileset in the tileset.|No|
```
lilleyse, line 214: `geometricError` should be in the extension's property reference as a way to store the root's geometric error. The root's geometric error differs from the top-level tileset geometric error. It should be a required property.

There should also be a note about how geometric error gets computed for child tiles. As written now the geometric error would divide by two for each level. Longer term geometric error could be encoded in a tile state buffer.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+                    0, 5, 0,
+                    0, 0, 5
+                ]
+            },
+            "refine": "REPLACE",
+            "tilingScheme": "octree",
+            "contentExtension": "glb",
+            "tilesetExtension": "json",
+            "content": {
+                "levelOffset": 4,
```
lilleyse, line 442: It feels awkward that in order to specify a default value for all tiles `levelOffset` needs to exceed the maximum level of the tileset. Makes sense with the current design, but feels off.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+                    0, 0, 0,
+                    5, 0, 0,
+                    0, 5, 0,
+                    0, 0, 5
+                ]
+            },
+            "tilingScheme": "quadtree",
+            "refine": "REPLACE",
+            "contentExtension": "glb",
+            "tilesetExtension": "json",
```
lilleyse, line 506: Should this be `"glb"` or `".glb"`? I think the latter, but not totally sure.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+│       │   │   └── 0/0.glb
+│       │   ├── 0/1/
+│       │   │   └── 0/1.glb
+│       │   ├── 1/0/
+│       │   │   └── 1/0.glb
+│       │   └── 1/1/
+│       │       └── 1/1.glb
+│       ├── L2/
+│       │   └── ...
+│       └── ...
```
lilleyse, line 571:
```
└── tileset.json/
    ├── 0/
    │   └── 0/
    │       └── 0.glb
    ├── 1/
    │   ├── 0/
    │   │   └── 0.glb
    │   │   └── 1.glb
    │   ├── 1/
    │   │   └── 0.glb
    │   │   └── 1.glb
    ├── 2/
    │   └── ...
    └── ...    
```

Feels like a more natural way to express this

extensions/3DTILES_implicit_tiling/README.md
lilleyse: I wrote this comment on the wrong line. See https://github.com/CesiumGS/3d-tiles/pull/414/files#diff-8e633d85b75fc0bf74e1677bd3c4d91dR292 instead:

> Specifies if additive or replacement refinement is used when traversing the tileset for rendering. This property is required for the root tile of a tileset; it is optional for all other tiles. The default is to inherit from the parent tile

lilleyse:
Made some smaller tweaks in https://github.com/CesiumGS/3d-tiles/pull/414/commits/06047d36df60619809347753c04209ed449dcbfe

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+
+This is a one bit representation of whether or not a tile has metadata associated with it.
+
+| Code | Description       |
+|------|-------------------|
+| `0`  | No metadata       |
+| `1`  | Has metadata      |
+
+
+When the tile has metadata, the index of the tile in the subdivision buffer will be the index of corresponding metadata property in the `3DTILES_tile_metadata` extension.
```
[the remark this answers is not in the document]
sanjeetsuhag: You are correct, my writing was so confusing that even I misread it. 😅

Here's the correct answer:

> When the tile has metadata, the index of the tile's properties in the `3DTILES_tile_metadata` buffers will be the index of the tile in the metadata bitstream added to the number of tiles in all levels before the level offset, if the level offset fill is `1`.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+
+#### Octree
+
+![Octree Image](figures/octree.png)
+
+When a tile subdivides into an [octree](https://en.wikipedia.org/wiki/Octree), it produces 8 equally size children tiles that occupy the same volume as the parent tile. The tile is split at the midpoint of the bounds along X, Y and Z axes.
+
+### Root Tiles
+
+Multiple tilesets using this extension can be combined by using a "base" tileset referring to one or more external tilesets using `3DTILES_implicit_tiling` as the children of the root tile. Such a configuration can be useful in representing structures like double-headed quadtrees in formats such as [TMS](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification) and [quantized-mesh](https://github.com/CesiumGS/quantized-mesh) and six-headed quadtrees in tiling schemes such as [S2geometry](https://s2geometry.io/).
```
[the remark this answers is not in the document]
sanjeetsuhag: Added a sentence linking to examples.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+
+#### Location in space
+
+![Morton](figures/morton.png)
+
+Level grids are indexed from the bottom left, using a right handed coordinate system, with +Z pointing in the up direction.
+
+Tiles in a level are indexed by applying the [Morton/Z-order](https://en.wikipedia.org/wiki/Z-order_curve) curve to the grid at that level. Using the Morton order serves three primary purposes:
+
+- Increasing spatial locality of reference: Tiles that are close to each in other in space will be located close to each other on the file system.
```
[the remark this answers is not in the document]
sanjeetsuhag: That's fair - but maybe this is good reason to store tiles as `Level/Morton` instead of `Level/X/Y/Z`.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
lilleyse: Even simpler variant:

> When the tile has metadata, the index of of the tile's properties in the `3DTILES_tile_metadata` buffers is the number of tiles containing metadata preceding this tile.

As long as it's clear that "preceding this tile" means all tiles with `1` in the bitstream and all tiles with an offset fill of `1`.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
lilleyse: Yeah it's one benefit of `Level/Morton`. Still `Level/X/Y/Z` is the familiar ordering. How much does the ordering matter once servers and edge caching come into play?

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
sanjeetsuhag: Updated with a description that will hopefully flow better:

> Existing formats that use implicit tiling schemes often use data structures such as double-headed quadtrees, in the case of [TMS](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification) and [quantized-mesh](https://github.com/CesiumGS/quantized-mesh), and six-headed quadtrees in tiling schemes such as [S2geometry](https://s2geometry.io/). To represent such tiling schemes, a "base" or parent tileset may be used to combine multiple tilesets that use this extension, where each tileset serves as the "head" or the root of subdivision.  The [examples](#examples) section includes some templates for representation of such data structures using this extension.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+- Efficient tile location decomposition: The Morton order allows efficient encoding and decoding of locations of a tile in the level grid to its location in the availability bitstream
```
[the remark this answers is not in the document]
sanjeetsuhag: Fixed.

extensions/3DTILES_implicit_tiling/README.md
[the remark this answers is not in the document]
lilleyse: 👍 I like the new wording

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Reading the examples with both the `maximumLevel` and `completeLevels` properties, their names seem a little inconsistent. We should pick one of the following:
- `completeLevels` and `levels`
- `maximumCompleteLevel` and `maximumLevel`
- `lastCompleteLevel` and `lastLevel`
- `completeDepth` and `depth`

sanjeetsuhag: Updated property to be required, updated samples to include this property.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: I agree that this is problematic. Perhaps, we can make it so that if `levelOffsetFill` is specified without `levelOffset`, then we can assume that the `levelOffsetFill` value applies to all tiles till `maximumLevel` (?)

extensions/3DTILES_implicit_tiling/README.md
lilleyse: True, one is a level and one is a depth, essentially.

I kind of like `completeDepth` and `depth`. CC https://github.com/CesiumGS/3d-tiles/pull/414#discussion_r437068930 too?

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: I think omitting the `.` is alright for now. Entwine, for examples, [does the same](https://entwine.io/entwine-point-tile.html#datatype).

lilleyse: The difference is that their `dataType` is more like an enum and ours is the actual extension.

As a data point for the other side, [`path.extname`](https://nodejs.org/api/path.html#path_path_extname_path) includes the dot.

sanjeetsuhag: Updated to include `.` in `contentExtension` and `tilesetExtension` properties.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Updated to include `geometricError` in property reference and in samples.

Regarding geometric error bitstream, would adding a custom multiplier suffice? Keep in mind that the geometric error would be reset in an external tileset.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Updated description to be applicable for implicit tiling.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,963 @@
...
+
+| Bitstream | Description       | Size (bits) |
+|------|-------------------|---|
+|  subdivision  | Encodes subdivision of each tile | 2 |
+|  content  | Encodes availability of content for tile | 1 |
+|  metadata  | Encodes availability of metadata for tile | 1 |
+
+#### Subdivision
+
+The subdivision state for each tile determines if and how it subdivides into children tiles, as per the `tilingScheme`. The subdivision state is encoded in 2 bits, and padded with 0s at the end to meet byte boundaries. At runtime, these padding bits will be ignored. A tile can have one of the following subdivision states:
```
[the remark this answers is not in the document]
sanjeetsuhag: Added a note in `3DTILES_binary_buffers` about padding for the `BIT` component type - and a link to it in the subdivision buffer description.

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: Updated directory structure in examples.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,985 @@
...
+---------------------------------------
+### 3DTILES_implicit_tiling Tileset JSON extension
+
+Specifies the properties for the 3DTILES_implicit_tiling object.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**boundingVolume**|`object`|A bounding volume that encloses the tileset.|☑️ Yes|
```
lilleyse, line 214: `transform` should also be included in this list

sanjeetsuhag:
Sandcastle for bounding volume diagrams:
- [Region](https://sandcastle.cesium.com/#c=7VZNTxsxEP0rVk4bNTgfLZcQUANFpWqhNKTlQDiY3UnWqtde2d6EFPHfO7Y3YZMmrUqREFJziOI3b+yZeeN4pkyTKYcZaLJPJMzIERheZPSbx6JRLfbrIyUt4xL0qNYgdyNJCJM8Y5Yr2SVjJgw0HGh5BgJpVWwCKlYJ6CqWqgwOC2tXvbkcq0N1W4VMDBJO0f2cx99X95Bsyic+ghMQ+cpuI3lf3xvJkQyZUQ1GicJRL2ImADPt0Nbe0uwPoROhboCaVM3ea1XIpG8zZfIUtOP7bdc9xreMoc3qAvxpsZLGkhng1z7ZaXda/pCAAivRdhU1qrApwq93K6BU2oNvAnPJzQW3/VswaLq6DumheBommBdiZWggLbccDGVJEt2FQmUoyKh2Xugck8+VmE/QozBcTohOi+yGOM0MmXE8FyNyK9TZ+ZbkbtAcheOgmY7TeXfRKEdMW/zF5Gs61ip7BxMNYPpas3l0FZyIr0kjJNZYYK4imzFflcaqr8cCdF0vbXCLpU8gOQE+SW2XdHZb+CmN2JygORMPgSqhNB0O+mcX5/3B8dmwJJYJd72Oq9glT2zaJW9WUb/R2raD43clCasznOewtPfDmg5Ovp4eOsp9Y9meS0lpXpg02qrgE6nge/AViXyH7vhlnTRJZ5MKyMP2a7Wfhr+q6PY4nkrly5MPw2Pq+rkv8pRFLdrZrT+l3oef+kcf/07xZ5G8cnlcxcNfy04AQs03M5da/sajclkf5bGB+V/6l/af23aatEtjWgVfrE7hXVW5mxf8Y+sIpRYW83dv6cC/uuUjiWFIAwJiNI0LGTtHEtUXPuTh6b5qXfsJozJRrDPaf2R0NjLuG4s8tgSLF+1LwRKLrfIvcZfTzvawtxAeH/Xn+PljrhDKkEcyzGAXTCYxXiEB7gYPlRI3TJ+CLKKyg1Ym0R9KZUO1fvfre6TWqPWMnQs4WITxlmc5XlRSaBFR2rSQ5QKvkmneFDgKWxobv7Wj9ppV117Cp4Qn+xtmdxILZgxaxoUQF/wHVvWg10T+L65CsQTnw89T0ILNHS1tH3wKIKW018TlZk8bSrC2808)
- [Box](https://sandcastle.cesium.com/index.html#c=7VZdT9swFP0rVp9SqbhpoS9tQSsFjWl8jVbjgfDgJreNNceObKctIP77bpwQ0gKatiGNh73k4/jc4+tz/bVkmiw5rECTfSJhRcZgeJbQ7w7zgkbo/sdKWsYl6KDRIg+BJIRJnjDLleyTORMGWjloeQICaXVsASpUEeg6FqsEDjNrN6O5nKtDta5DJgQJZxh+ycMfmxqSLfnCZXACIt1QC+RjcxDIQBYjoxqMEllOnYRMAI60S/1B1ew6oQuhZkBNrFaftcpkNLKJMmkMOuc72e2I+ZoxbLM6A9dbqKSxZAX42Cc7na7vOilQYCXaqaNGZTZGeLdXA6XSDtwrmBU3FdyO1mCw6ea2GB4Wb6bWCJR5gbTccjCURZHnypQqw4sqlYUdM23xi8ldOtcqOYKFBjDeTsfH7lqu0xbp9ny/6Vye5fVwSoREWF1pUMz061PlWdFDkUKm+ui5d6FFCM4Y0JyJ52yUUJpOr0bnk8vR1fH5tCSiL8U8ys3dxK55ZOM+2dtEndCW7NXxUc55rKZDZSFNMxN7H8W03zfs+uTL9JiuuI1HIo2Z59Nur/me1h2ejsZfK/M+tHvVlPvv3nss2M7HtK/Y7FSaW+N2wLy9HKaFte2ToIFnBx5Ppbw0ICBEfJ7JMI8iXvMpgDxvpjf+rdvza3v8NqPzS0b3VcajS6V4vsyU7JBvGYssFvNvki4Pn7dzfoPwhylfhP8+4RqhzDeQxXk4YTIK8aAVkK+nqVJixvQZyMwrJ87GreBeqWSqtldic0AarcbQ2DsBB09pfOJJiqcyybTwKG1bSFKBi8O0ZxleSywNjZPOqcN2PXQY8SXh0f4r9ygSCmYMtswzISb8Hl09GLaR/yJUKBZxubhYghbsLqfFnYPTAqSUDtv4+3qkLSzYUv4J)
- [Geodesic Quad]()

extensions/3DTILES_implicit_tiling/README.md
sanjeetsuhag: There's two new kinds of diagrams, we can select whichever one looks better:

Non-RGB version:
box_octree


RGB version:
box_octree_xyz

lilleyse:
Superseded by https://github.com/CesiumGS/3d-tiles/pull/444

PR #415, opened 2020-06-08
Adds 3DTILES_binary_buffers extension

sanjeetsuhag:
This extension enables storage of binary buffers. Specification is available [here](https://github.com/CesiumGS/3d-tiles/blob/3DTILES_binary_buffers/extensions/3DTILES_binary_buffers/README.md).

### DIFF
--- a/extensions/3DTILES_binary_buffers/README.md
+++ b/extensions/3DTILES_binary_buffers/README.md
@@ -41,3 +41,100 @@ Buffers are binary blobs of data.
 ### Buffer Views
 
 Buffer views offer typed views into buffers. They specify a subset of the data stored in a buffer through a `byteOffset` and a `byteLength`. The type of data inside the buffers can be derived from the `elementType` and the `componentType` properties.
+
+## Examples
+
+### 3DTILES_implicit_tiling and 3DTILES_tile_metadata
+
+```json
+{
+    "asset": {
+        "version": "2.0.0-alpha.0"
+    },
+    "extensions": {
+        "3DTILES_implicit_tiling": {
+            "tilingScheme": "quadtree",
+            "subdivision": {
+                "completeLevels": 2,
+                "bufferView": 0
+            },
+            "content": {
+                "levelOffset": 2,
+                "levelOffsetFill": 0,
+                "bufferView": 1
+            },
+            "metadata": {
+                "levelOffset": 2,
+                "levelOffsetFill": 0,
+                "bufferView": 1
+            }
+        },
+        "3DTILES_tile_metadata": {
+            "properties": {
+                "surfaceArea": {
+                    "semantic": "_SURFACE_AREA",
+                    "bufferView": 2
+                },
+                "tileId": {
+                    "semantic": "_TILE_ID",
+                    "bufferView": 3
+                }
+            }
+        },
+        "3DTILES_binary_buffers": {
+            "bufferViews": [
+                {
+                    "elementType": "VEC2",
+                    "componentType": "BIT",
+                    "byteOffset": 0,
+                    "byteLength": 2,
+                    "buffer": 0
+                },
+                {
+                    "elementType": "SCALAR",
+                    "componentType": "BIT",
+                    "byteOffset": 2,
+                    "byteLength": 4,
+                    "buffer": 0
+                },
+                {
+                    "elementType": "SCALAR",
+                    "componentType": "FLOAT",
+                    "byteOffset": 0,
+                    "byteLength": 32,
+                    "buffer": 1
+                },
+                {
+                    "elementType": "STRING",
+                    "byteOffset": 32,
+                    "byteLength": 64,
+                    "elementsOffsetBufferView": 5,
+                    "buffer": 2
+                },
+                {
+                    "elementType": "SCALAR",
+                    "componentType": "UNSIGNED_INT",
+                    "byteOffset": 0,
+                    "byteLength": 32,
+                    "buffer": 2
+                }
+            ],
+            "buffer": [
+                {
+                    "uri": "implicit.bin",
+                    "byteLength": 6
+                },
+                {
+                    "uri": "metadata_surface_area.bin",
+                    "byteLength": 32
+                },
+                {
+                    "uri": "metadata_tile_id.bin",
+                    "byteLength": 96
+                }
+            ]
+        }
+    }
+}
+
+```
--- a/extensions/3DTILES_binary_buffers/README.md
+++ b/extensions/3DTILES_binary_buffers/README.md
@@ -1,0 +1,43 @@
+# 3DTILES_binary_buffers
+
+## Overview
+
+This extension to 3D Tiles enables storage of binary data in external buffers.
+
+## Concepts
+
+### Element Types
+
+| Element Type | No. of components |
+|:------------:|:-----------------:|
+| SCALAR | 1 |
+| STRING | 1 |
+| VEC2 | 2 |
+| VEC3 | 3 | 
+| VEC4 | 4 |
+| MAT2 | 4 |
+| MAT3 | 9 |
+| MAT4 | 16 |
+
+### Component Types
+
+| Component | Size (bits) |
+|:---------:|:------------:|
+| BIT | 1 |
+| BYTE | 8 |
+| UNSIGNED_BYTE | 8 |
+| SHORT | 16 |
+| UNSIGNED_SHORT | 16 |
+| HALF_FLOAT | 16 |
+| INT | 32 |
+| UNSIGNED_INT | 32 |
+| FLOAT | 32 |
+| DOUBLE | 64 |
+
+### Buffers
+
+Buffers are binary blobs of data.
+
+### Buffer Views
+
+Buffer views offer typed views into buffers. They specify a subset of the data stored in a buffer through a `byteOffset` and a `byteLength`. The type of data inside the buffers can be derived from the `elementType` and the `componentType` properties.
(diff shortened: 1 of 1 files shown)


extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+                },
+                {
+                    "elementType": "SCALAR",
+                    "componentType": "UNSIGNED_INT",
+                    "byteOffset": 0,
+                    "byteLength": 32,
+                    "buffer": 2
+                }
+            ],
+            "buffer": [
```
lilleyse, line 146:
```suggestion
            "buffers": [
```

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+            },
+            "content": {
+                "levelOffset": 2,
+                "levelOffsetFill": 0,
+                "bufferView": 1
+            },
+            "metadata": {
+                "levelOffset": 2,
+                "levelOffsetFill": 0,
+                "bufferView": 1
```
lilleyse, line 93: Is it intentional that `content` and `metadata` reference the same `bufferView`?

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+                    "componentType": "FLOAT",
+                    "byteOffset": 0,
+                    "byteLength": 32,
+                    "buffer": 1
+                },
+                {
+                    "elementType": "STRING",
+                    "byteOffset": 32,
+                    "byteLength": 64,
+                    "elementByteOffsetsBufferView": 5,
```
lilleyse, line 135:
```suggestion
                    "elementByteOffsetsBufferView": 4,
```

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+|**byteLength**|`integer`|The total byte length of the buffer view.| ☑️ Yes|
+
+## Examples
+
+### 3DTILES_implicit_tiling and 3DTILES_tile_metadata
+
+```json
+{
+    "asset": {
+        "version": "2.0.0-alpha.0"
```
lilleyse, line 76: Change to `1.0` - see https://github.com/CesiumGS/3d-tiles/pull/414/files#r437013140

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**buffer**|`integer`|The index of the buffer.|☑️ Yes|
+|**byteOffset**|`integer`|The offset relative to the start of the buffer in bytes.|☑️ Yes|
+|**byteLength**|`integer`|The length of the bufferView in bytes.| ☑️ Yes|
+|**elementCount**|`integer`|The number of elements in the buffer view.| ☑️ Yes|
+|**elementByteOffsetsBufferView**|`integer`|The index of the bufferView containing byte offsets for each element. Must be defined for the STRING type.|No|
```
lilleyse: Can `elementByteOffsetBufferView` be used for properties containing arbitrary sized chunks of binary data? What would be the `elementType` in that case?

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**buffer**|`integer`|The index of the buffer.|☑️ Yes|
+|**byteOffset**|`integer`|The offset relative to the start of the buffer in bytes.|☑️ Yes|
+|**byteLength**|`integer`|The length of the bufferView in bytes.| ☑️ Yes|
+|**elementCount**|`integer`|The number of elements in the buffer view.| ☑️ Yes|
+|**elementByteOffsetsBufferView**|`integer`|The index of the bufferView containing byte offsets for each element. Must be defined for the STRING element type.|No|
```
lilleyse, line 22: Should there also be an element type for arbitrary sized chunks of binary data? That would look similar to `STRING` and would also use `elementByteOffsetsBufferView`.

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+|**elementType**|`string`|Specifies if the attribute is a scalar, vector, matrix or string.|No, default is `SCALAR`.|
+|**componentType**|`string`|The datatype of components in the attribute. Cannot be defined for the STRING element type.|No|
```
lilleyse, line 24: And must be defined for all other element types.

extensions/3DTILES_binary_buffers/README.md
lilleyse: Sorry, duplicate comment to https://github.com/CesiumGS/3d-tiles/pull/415#discussion_r437015903. I thought github lost this one.

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+#### Element Types
+
+| Element Type | No. of components |
+|:------------:|:-----------------:|
+| SCALAR | 1 |
```
lilleyse, line 31: Is `1` correct or should it be `NA`?

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+| STRING | 1 |
```
lilleyse, line 32: Is `1` correct or should it be `NA`?

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,164 @@
...
+---------------------------------------
+### 3DTILES_binary_buffers.buffers
+A buffer points to a blob data.
+
+**Properties**
+
+|   |Type|Description|Required|
+|---|----|-----------|--------|
+|**uri**|`string`|The uri of the buffer.| ☑️ Yes|
+|**byteLength**|`integer`|The total byte length of the buffer view.| ☑️ Yes|
```
lilleyse, line 67:
```suggestion
|**byteLength**|`integer`|The total byte length of the buffer.| ☑️ Yes|
```

extensions/3DTILES_binary_buffers/README.md
sanjeetsuhag: Yes, to show the case in which you only have metadata (rendering info, for example) for the tiles in which you have content.

extensions/3DTILES_binary_buffers/README.md
sanjeetsuhag: Yes, a `NONUNIFORM` element type will be introduced to meet the requirements for `EXT_3dtiles_feature_metadata`.

extensions/3DTILES_binary_buffers/README.md
sanjeetsuhag: Updated - now `componentType` will be ignored for `STRING` element type.

extensions/3DTILES_binary_buffers/README.md
sanjeetsuhag: `1` is correct. We're treating the element between offsets are one value.

extensions/3DTILES_binary_buffers/README.md
```diff
@@ -0,0 +1,166 @@
...
+| SHORT | 16 |
+| UNSIGNED_SHORT | 16 |
+| HALF_FLOAT | 16 |
+| INT | 32 |
+| UNSIGNED_INT | 32 |
+| FLOAT | 32 |
+| DOUBLE | 64 |
+
+
+*Note: The buffer views for the `BIT` component type must be padded with trailing `0`s to meet the nearest byte boundary.*
```
lilleyse, line 58: The padding requirements should be defined for the other component types. In general properties must start on a byte offset that's a multiple of the size of the component type.

See https://github.com/CesiumGS/3d-tiles/tree/master/specification/TileFormats/BatchTable#padding

sanjeetsuhag: What's the reason behind this? As long as the byte boundary is respected, the data can be read using byte offsets and byte lengths. What is gained by having the offset be a multiple of the size of the component type?

lilleyse: Talked offline about this, but it's for performance reasons. In JavaScript typed arrays need to be aligned to their component type within the underlying array buffer which means unaligned data needs to be cloned into a new typed array. Proper alignment also provides performance benefits in systems level languages.

Early on glTF had the same rationale: https://github.com/KhronosGroup/glTF/issues/167

https://github.com/CesiumGS/cesium/blob/2a9ca28dcbb38eee7013505987516446a8eacb4f/Source/Scene/Batched3DModel3DTileContent.js#L370-L382 - this is where CesiumJS clones the typed array if it isn't aligned. 3D Tiles didn't always have byte alignment requirements which lead to slow paths like this.

pjcozzi:
@sanjeetsuhag @lilleyse excited to see this.  Just a reminder to please bump to me when ready for review. 😄

lilleyse:
This extension may need a way to indicate whether a string property is stringified JSON

lilleyse:
A lot of the data alignment rules in this spec are now OBE since we're moving to Apache Arrow. I'll open a PR with the new, simplified version of the spec.

PR #418, opened 2020-08-03
added comment to vector data encoding example code

KeyboardSounds:
I don't know if this spec is still "alive", but I ran into a problem when writing code to create vector 3dtiles and thought that adding a hint to the spec might prevent people running into the same issue in the future.


lilleyse:
@KeyboardSounds the new vector data spec is in progress but I'm ok with merging this in the meantime

PR #429, opened 2020-09-22
Adds 3DTILES_tile_metadata extension

lilleyse:
Supersedes https://github.com/CesiumGS/3d-tiles/pull/413 which was not versioned.

Direct Link|Draft Version|Changes
--|--|--
[`3DTILES_tile_metadata`](https://github.com/CesiumGS/3d-tiles/blob/3DTILES_tile_metadata/extensions/3DTILES_tile_metadata) | 0.0.0 | Initial draft

This extension enables storage of per tile metadata. Eventually this will be the backbone of linearized trees for explicit tiling where bounding volumes, geometric error, child tile pointers, etc can all be stored in binary alongside application-specific properties. That's the next phase of this PR at which point it might not make sense to call it `tile_metadata`.

### DIFF
--- a/extensions/3DTILES_tile_metadata/README.md
+++ b/extensions/3DTILES_tile_metadata/README.md
@@ -1,0 +1,202 @@
+# 3DTILES_tile_metadata
+
+## Contributors
+
+* Sam Suhag, Cesium
+* Sean Lilley, Cesium
+
+## Status
+
+Draft
+
+## Dependencies
+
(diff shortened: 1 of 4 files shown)


lilleyse:
Tile metadata will be folded into [3DTILES_metadata](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_metadata)

PR #432, opened 2020-10-21
Adds draft for 3DTILES_layers

sanjeetsuhag:
Direct Link|Draft Version|Changes
--|--|--
[`3DTILES_layers`](https://github.com/CesiumGS/3d-tiles/blob/7c906811e4e6b0d43dc36687143df8f637aae194/extensions/3DTILES_layers) | 0.0.0 | Initial draft
[`3DTILES_layers`](https://github.com/CesiumGS/3d-tiles/tree/3DTILES_layers/extensions/3DTILES_layers) | Latest | Latest draft

This extension to 3D Tiles adds the concept of content layers to a tileset.


extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
+# 3DTILES_layers
+
+## Overview
+
+This extension to 3D Tiles enables grouping of different types of content into layers. A layer is a user-defined grouping of content - for example, a tileset of a city may have a buildings layer, a roads layer and a vegetation layer. Each layer follows an independent layout for its tiles and there are not dependencies between layers. A layer may also associate with itself some metadata by conforming to a `class` declared in `3DTILES_metadata`.
```
ptrgags, line 5: Maybe briefly mention that `class` in turn comes from the metadata spec?

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Contributors
+
+Cesium
```
ptrgags, line 9: @sanjeetsuhag @lilleyse I feel like we should talk about this briefly: since we have more extensions than we used to, how do we want to list contributors in each extension?

extensions/3DTILES_layers/README.md
lilleyse, line 5:
```suggestion
This extension to 3D Tiles enables assigning tile content to layers.

A layer is a user-defined grouping of content - for example, a tileset of a city may have a buildings layer, a roads layer and a vegetation layer. There are no dependencies between layers. A layer may contain application-specific metadata by conforming to a `class` defined in `3DTILES_metadata`.
```

> Each layer follows an independent layout for its tiles

I removed this part because it's not super clear what it means. Could it be reworded?

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+## Schema Changes
+
+### Layer Metadata
+
+Layers are declared inside the top-level `3DTILES_layers` extension dictionary. Each layer is an object with its key being the layer's ID. The layer may declare a `name` and a `description`. 
+
+```javascript
+{
+  "asset": {
+    "version": "2.0.0-alpha.0"
```
lilleyse, line 20:
```suggestion
    "version": "1.0"
```

Make sure to fix this for other examples too

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "extensions": {
+    "3DTILES_layers": {
+      "BUILDINGS": {
```
lilleyse, line 24: For consistency with the 3D metadata spec: `"buildings"` rather than `"BUILDINGS"`

Same with any other examples below, and same for properties, prefer `camelCase`.

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+        "name": "Buildings",
+        "description": "3D Buildings Layer"
+      }
+    }
+  }
+}
+```
+
+Additionally, if a layer needs to associate some metadata with itself, there needs to be a corresponding class declared in the `3DTILES_metadata` extension and the layer needs to conform to that class. To learn more about metadata classes, refer to the [Cesium 3D Metadata Specification](). For example,
```
lilleyse, line 33: > if a layer needs to associate some metadata with itself

Sounds a little awkward. Could be worded more like the suggestion above.

Explicitly say that a layer is an instance of class. Also say that it uses the single instance shorthand.

> The layer object must use the [single instance shorthand syntax](https://github.com/CesiumGS/3d-tiles/tree/3d-metadata-spec/specification/Metadata#single-intance-shorthand) to assign values to its properties.

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+  },
+  "extensions": {
+    "3DTILES_metadata": {
+      "classes": {
+        "CITY_LAYER": {
+          "properties": {
+            "LastModified": {
+              "type": "STRING",
+              "optional": false
+            }
```
lilleyse, line 48: For some variation it might be good to add a second property, maybe "color"

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+Cesium
+
+## Schema Changes
+
+### Layer Metadata
+
+Layers are declared inside the top-level `3DTILES_layers` extension dictionary. Each layer is an object with its key being the layer's ID. The layer may declare a `name` and a `description`. 
+
+```javascript
+{
```
lilleyse, line 18: It might be nice if all examples in this document have the three layers described in the overview: a buildings layer, a roads layer and a vegetation layer (or rename to "trees" layer). I don't think one layer is enough to convey the extension.

extensions/3DTILES_layers/README.md
lilleyse: It's worth having a paragraph or two talking about what layers enable.

* Enables a tile to have multiple contents that can be requested independently from each other. Avoids the drawback of the cmpt format which is all or nothing.
* Datasets that inherently follow the same tiling scheme don't need to be stored in separate tileset.json files
* Provides a way to pair regular tile content like b3dm, glTF with application specific content, e.g. some game engine specific payload.
* Would enable future capabilities like navigation meshes, physics meshes as separate layers
* Styling use cases: show/hide, change color. Enabled automatically by the styling language since layer properties are inherited by features.
* For some applications, z-order

extensions/3DTILES_layers/README.md
```diff
@@ -0,0 +1,98 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### Layer Content
+
+To associate content with a layer, it must be added as an object to the `contents` array of the extension declared at the `root` of the tileset.json. To pair the content to a layer, we must set the `layer` property of the layer content object to the ID of the layer declared at the top level extension. This layer object needs to declare the `layer`, the `uri` of its contents and the `mimeType` for the content linked in the `uri`. For example,
```
lilleyse, line 69: I think the key point that should be made clear from the start of this paragraph is that this extension allows a tile to have multiple contents. And _then_ talk about how to associate each content with a layer.

Other points

* The extension can be added to any tile, not just `root`
* Probably worth stating explicitly that content may be regular tile content (b3dm, etc) or external tileset content
* I kinda think `layer` should be optional. Some users of the extension might not need to associate content with layers
* Side note: we don't really talk much about runtime behavior. I guess default behavior would be to load all layers when a tile is selected?

lilleyse:
Hm I wonder if the 3DTILES_metadata diagram actually belongs in here instead. It would definitely clear up some concepts.

https://github.com/CesiumGS/3d-tiles/blob/0fab3ee872b6fb8bc7fa1db94f2b67f1306dca9a/extensions/3DTILES_metadata/README.md#overview

extensions/3DTILES_layers/README.md
sanjeetsuhag: Added `hightlightColor` property.

extensions/3DTILES_layers/README.md
sanjeetsuhag: Added `buildings`, `trees`, `roads` as instances of `cityLayer`.

lilleyse:
Thanks @sanjeetsuhag, merging into `3DTILES_metadata`.

PR #434, opened 2020-11-03
Cesium 3D Metadata Specification

ptrgags:
For 3D Tiles Next, we decided to move some of the parts of the spec that are common to multiple extensions into its own document. This is a draft of the main text of the document, along with a schema.

| Direct Link | Draft Version | Changes |
| ---------------| -------| ----|
| Cesium 3D Metadata Specification | 0.0.0 | Initial draft |
| [Cesium 3D Metadata Specification](https://github.com/CesiumGS/3d-tiles/tree/3d-metadata-spec/specification/Metadata) | Latest | latest changes|

This specification provides a method for describing metadata and storing it in JSON, in binary, or in textures. It will be referenced by a 3D Tiles extension and a glTF extension in time.

This is still a draft, and there's a couple details that need to be added:
- [x] I need to add diagrams to help explain some examples.
- [x] It looks like I missed a spot for some details about binary encoding (namely arrays)

@lilleyse and @sanjeetsuhag could you review this tomorrow morning?


specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+TODO: Diagram for binary string
+
+TODO: Diagram for array of fixed-length strings
+
+#### Arrays
+
+OUTLINE:
+- Vec3 example
+- longer example (2x2 matrix = 4 components at every point)
+- diagrams of the above
```
ptrgags, line 815: Whoops, I forgot to fill out this section for the binary encoding.

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+      - [Binary Alignment Rules](#binary-alignment-rules)
+    - [Metadata Texture Encoding](#metadata-texture-encoding-1)
+      - [Numeric Textures](#numeric-textures)
+      - [Array Textures](#array-textures)
+      - [Implementation Notes](#implementation-notes)
+  - [Glossary](#glossary)
+
+## Abstract
+
+This specification provides a standard format for adding metadata to 3D models and tilesets. It provides a method for declaring metadata, as well as methods for storing this metadata in JSON, binary, or texture format. This metadata format is shared by several Cesium specifications. This avoids repetition and enforces a consistent data layout. Specifications that reference this document must include at least one metadata encoding as described in this document, and must make clear which encodings are supported.
```
sanjeetsuhag, line 65: > adding metadata to 3D models and tilesets

I would explicitly refer to glTF and 3D Tiles here.

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+
+## Introduction
+
+```
+features = geometry + metadata
+```
+
+Using terminology from the geospatial industry, **features** are geometric entities in 3D space (**geometry**) and may have application specific properties associated with them (**metadata**). These properties may span a myriad of domains, such as the heights of buildings in a city, the names of different structures in a CAD model, the classification of different points in a point cloud, or the age of different trees in a forest of instanced tree models.
+
+3D models are useful for describing geometry, but often there is limited support for annotating the geometry with metadata. This specification adds a mechanism for storing an unlimited amount of metadata to existing 3D model formats such as Khronos Group's glTF or Cesium's 3D Tiles. Metadata can be applied at various granularities: per-vertex, per-texel, as well as larger structures from meshes to tilesets. This metadata can be used for analytics purposes, as well as styling in a 3D renderer.
```
sanjeetsuhag, line 75: > 3D models are useful for describing geometry, but often there is limited support for annotating the geometry with metadata.

I think we can remove this line.

> an unlimited amount

Is this necessary?

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+
+For a more detailed description of how classes are defined, see the [Class Definitions](#class-definitions) section below.
+
+### Instances
+
+A class definition is abstract, and just describes what metadata exists. Meanwhile, an **instance** is a concrete representation of the metadata for a single piece of geometry that can be stored in a file. This specification provides two main types of data storage for instances a columnar format (**instance tables**) and a texture-based format (**metadata textures**). The main distinction is that instance tables are designed for per-vertex metadata, while metadata textures are designed for per-texel metadata. Furthermore, instance tables can be stored in either JSON or binary formats, so there is a total of three storage encodings.
```
sanjeetsuhag, line 132:
```suggestion
A class definition is abstract, and just describes what metadata exists. Meanwhile, an **instance** is a concrete representation of the metadata for a single piece of geometry that can be stored in a file. This specification provides two main types of data storage for instances: a columnar format (**instance tables**) and a texture-based format (**metadata textures**). The primary distinction is that instance tables are designed for per-vertex metadata, while metadata textures are designed for per-texel metadata. Furthermore, instance tables can be stored in either JSON or binary formats, so there is a total of three storage encodings.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+
+A brief overview of each encoding follows to explain the concepts. The full details can be found further below in the [Storage Encodings](#storage-encodings) section.
+
+#### Instance Tables
+
+Instance tables are the typical method for storing metadata. An instance table is a mapping of (**instance IDs**) to metadata values which are stored in parallel vectors called **property arrays**. Instance IDs are simply integer indices into these arrays. The `i-th` value of every property array in an instance table together makes up the metadata for the `i-th` instance.
```
sanjeetsuhag, line 138:
```suggestion
An instance table is a mapping of (**instance IDs**) to metadata values which are stored in parallel vectors called **property arrays**. Instance IDs are simply integer indices into these arrays. The `i-th` value of every property array in an instance table together makes up the metadata for the `i-th` instance. 
```

The first line is unnecessary.

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+This specification keeps class definition separate from instantiation. While this adds a level of indirection, there are several reasons for this separation:
+
+* It allows the class definition to be stored seperately from the data itself. This is useful when one has a dataset that has many 3D models with the same shape of metadata. Then instead of re-declaring the class many times, a single definition can be shared across multiple models
+* It allows greater flexibility for storing properties. For example, take the `elevation` property from the example above. The natural representation is to use a metadata texture. However, if the corresponding 3D model is a high-resolution terrain mesh, storing the elevations per-vertex may be more appropriate.
+
+## Class Definitions
+
+### Basic Types
+
+A class is a collection of properties. Each class requires a unique class ID. This can be any UTF-8 string. The uniqueness is enforced by a JSON dictionary. Within the class, one or more properties is defined. Each property has a type associated with it. Some types have extra options; the sections below discuss such cases.
```
sanjeetsuhag, line 263: Both classes and properties need to be unique.

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+
+Classes are defined in a `classes` section of the JSON file
```
sanjeetsuhag, line 265: I'd remove this. It's on the implementing spec to specify where `classes` are declared.

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+  "classes": {
+    "basicTypes": {
+      "properties": {
+        "date": {
+          "name": "Date (YYYYMMDD)",
+          "type": "STRING",
+          "stringByteLength": 4
+        },
+        "fixedBlobProperty": {
+          "type": "BLOB"
```
sanjeetsuhag, line 539:
```suggestion
          "type": "BLOB",
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,990 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+#### Single Intance Shorthand
+
+In most situations, there are several instances of a class. Therefore, an array is a natural choice for the JSON representation. However, in cases where only a single instance is used, the syntax can feel cumbersome. This specification allows for a simpler property dictionary
```
sanjeetsuhag, line 631:
```suggestion
In most situations, there are several instances of a class. Therefore, an array is a natural choice for the JSON representation. However, in cases where only a single instance is used, this specification allows for a simpler property dictionary.
```

ptrgags:
@sanjeetsuhag @lilleyse I added the diagrams and responded to feedback. This is ready for review again.

@lilleyse did you ever write down your feedback for me?

specification/Metadata/schema/class.property.schema.json
```diff
@@ -0,0 +1,110 @@
...
+          },
+          "minItems": 1
+      },
+      "default": {
+          "type": ["boolean", "number", "string", "array"],
+          "description": "A default value to use when no value for this property is specified.  This happens only when using JSON encoding and `null` is used to indicate no value. If used, `optional` must be set to `true`. The data type here must match the property declaration. However, `BLOB` properties must be encoded in Base64 encoding since JSON does not support arbitrary binary sequences."
+      },
+      "optional": {
+          "type": "boolean",
+          "description": "If true, this property is optional and can take a null value. This is accomplished by not using a `bufferView` or by using a validity buffer. A validity bit set to 0 indicates the absence of data for that instance. In both cases, `optional` must be set to `true`.",
```
lilleyse, line 99: Remove mention of validity buffer. Should JSON encoding be mentioned at all?

lilleyse (commented):
This is a great start for the writeup. I thought the concepts were presented clearly and in the right order. Will do a more thorough pass over the schemas another time.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+    }
+  }
+}
+```
+
+![vec3 metadata texture](figures/vec3-texture.jpg)
+
+#### Implementation Notes
+
+Since the core glTF specification only supports PNG and JPEG images, in practice the data types available in textures is somewhat limited. Only data types based on `UINT8` or `INT8` (including normalized versions and fixed-length arrays of these) are straightforward to implement. Other data types would require additional glTF extensions to support other texture formats. Cases like these are left to the implementation to define.
```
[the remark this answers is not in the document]
ptrgags: Ah right, we talked about this a couple days ago but I guess I didn't update the text. I'll fix this today.

specification/Metadata/schema/class.property.schema.json
ptrgags: sounds good, and yes, we should mention that this is only usable with JSON encoding.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+For `ARRAY` of `BOOLEAN`, `componentsPerElement` refers to the number of _bits_ per element. These bits are counted out from the overall bit vector sequence.
+
+In the case of a `VARIABLE_SIZE_ARRAY` of `BOOLEAN`, the offset buffer represents
+the _bit_ offsets, rather than the usual byte offsets. Again, the offsets are counted relative to the overall bit vector sequence.
+
+#### Binary Alignment Rules
+
+Due to the possibility of 64-bit data types, this extension requires that each `bufferView` is aligned to a multiple of 8 bytes for efficient reads. This alignment requirement is always measured from the start of the binary file. This alignment requirement only applies to the start of each bufferView used for metadata.
+
+To meet this alignment requirement, padding is sometimes needed between `bufferView`s. When this happens, use `0x00` for padding.
```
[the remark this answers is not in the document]
ptrgags: Fair point, I'll reword it.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+![Numeric types binary layout](figures/numeric-types.png)
+
+(lengths measured in bytes)
```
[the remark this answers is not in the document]
ptrgags: okay I'll update that everywhere in the text (there's a few diagrams like this)

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+    }
+  }
+}
+```
+
+#### Metadata Texture Encoding
+
+The metadata texture encoding serves a different purpose than the other encodings. Instead of associating metadata with an instance, this encoding associates metadata with each individual texel in a texture. In this situation, values are indexed by texture coordinates rather than instance ID. This allows, among other things, interpolation of metadata values (at least for numeric types).
+
+This is useful for continuous properties such as elevation, depth, vector fields, and many other properties that vary with position. It also can be used for better compression of metadata values in some cases.
```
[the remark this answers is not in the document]
ptrgags: >What's the difference between elevation and depth?

The main difference is the sign of the values. Elevation is positive above the datum (say MSL) while depth is positive _below_ the datum (e.g. for underwater details).

Does this need to be clarified?

And I can add vegetation index.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+
+```json
+{
+  "classes": {
+    "basicTypes": {
+      "properties": {
+        "date": {
+          "name": "Date (YYYYMMDD)",
+          "type": "STRING",
+          "stringByteLength": 4
```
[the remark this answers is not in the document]
ptrgags: :man_facepalming:  wait I probably have one of my diagrams wrong as well, for some reason I had "YYYYMMDD is 6 characters" in my head when really it's 8.

Fortunately since these packet diagrams are generated using `packetdiag` should be easy to fix.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+  },
+  "metadataTextures": {
+    "vegetationTexture": {
+      "class": "vegetation",
+      "properties": {
+        "vegetationIndex": {
+          "texture": {
+            "index": 0,
+            "texCoord": 0
+          },
```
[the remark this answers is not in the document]
ptrgags: Yeah this seems out-of-scope for this week. I opened https://github.com/CesiumGS/3d-tiles-next/issues/89 for this instead.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+      "properties": {
+        "date": "2020-11-03",
+        "authors": ["Jane Doe", "John Smith"]
+      }
+    }
+  }
+}
+```
+
+### Binary Encoding
```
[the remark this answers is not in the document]
ptrgags: I'll do this but as a note under the "Storage Encodings" heading since this is referring to multiple encodings at once.

specification/Metadata/README.md
```diff
@@ -0,0 +1,1130 @@
...
+          "description": "A basic example of using `null` for unknown values",
+          "type": "FLOAT32",
+          "optional": true
+        },
+        "history": {
+          "name": "Previous 5 sensor readings",
+          "description": "This example demonstrates the use of `[]` for null arrays as well as how `null` can be used for individual array components.",
+          "type": "ARRAY",
+          "componentType": "STRING",
+          "optional": true,
```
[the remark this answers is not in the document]
ptrgags: I'd rather see `[null, null, null, null, null]` in this case. The only special case is that `[]` is used instead of `null` to indicate the absence of an array.

ptrgags:
@sanjeetsuhag @lilleyse okay this is ready for review again!

specification/Metadata/0.0.0/README.md
```diff
@@ -0,0 +1,1217 @@
...
+        }
+      } 
+    },
+    "shorthandTable": {
+      "name": "Shorthand property example",
+      "description": "This is equivalent to the table above, just more concise",
+      "class": "datasetInfo",
+      "properties": {
+        "date": "2020-11-03",
+        "authors": ["Jane Doe", "John Smith"]
```
lilleyse, line 753: For clarity the summary above should say that optional properties may be omitted in the short-hand version

lilleyse:
Merging into `3DTILES_metadata`! And soon `3d-tiles-next`. Thanks @ptrgags!!!

PR #436, opened 2020-11-16
Add Texture schema to Cesium 3D Metadata Spec

ptrgags:
This adds a couple missing schema files to the `schema` folder, and adds a few notes in the README to say that some examples are glTF specific (as `texture` is not actually part of this schema!)

@lilleyse could you review?

### DIFF
--- a/specification/Metadata/0.0.0/README.md
+++ b/specification/Metadata/0.0.0/README.md
@@ -273,8 +273,7 @@ This is useful for continuous properties such as elevation, vegetation index, ve
 }
 ```
 
-**Note**: The method of selecting a texture is implementation-defined. The above example is based on a glTF implementation where `index` selects a texture and `texCoord` selects the texture coordinates. Other implementations may do this in
-alternate ways.
+**Note**: The method of selecting a texture is implementation-defined. The above example is based on a glTF implementation where `index` selects a texture and `texCoord` selects the texture coordinates. Other implementations may do select a texture by other means.
 
 #### Comparison of Encodings
 
--- a/specification/Metadata/0.0.0/README.md
+++ b/specification/Metadata/0.0.0/README.md
@@ -244,7 +244,7 @@ The metadata texture encoding serves a different purpose than the other encoding
 
 This is useful for continuous properties such as elevation, vegetation index, vector fields, and many other properties that vary with position. It also can be used for better compression of metadata values in some cases.
 
-```json
+```jsonc
 {
   "classes": {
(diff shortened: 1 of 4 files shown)


specification/Metadata/0.0.0/schema/textureAccessor.schema.json
```diff
@@ -0,0 +1,14 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "A description of how to access data values from the color channels of a texture",
```
lilleyse, line 3: `"title"` should just be "Texture Accessor" and this summary should go in a `"description"` property. The description should also include the note that the method of selecting a texture is implementation-defined.

specification/Metadata/0.0.0/schema/metadataTexture.schema.json
```diff
@@ -0,0 +1,24 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema",
+    "title": "Metadata Texture",
+    "type": "object",
+    "description": "Instances whose properties are stored directly in the texture designated by a class and property values stored in texture channels. This requires a class but not a feature table.",
```
lilleyse, line 5: > This requires a class but not a feature table

I think this sentence can be removed

specification/Metadata/0.0.0/schema/metadataTexture.schema.json
```diff
@@ -0,0 +1,24 @@
...
+    "allOf": [ { "$ref": "glTFProperty.schema.json" } ],
```
lilleyse, line 6: This line should be removed

specification/Metadata/0.0.0/README.md
```diff
@@ -261,18 +261,21 @@ This is useful for continuous properties such as elevation, vegetation index, ve
...
+          }
         }
       }
     }
   }
 }
 ```
 
+**Note**: The method of selecting a texture is implementation-defined. The above example is based on a glTF implementation where `index` selects a texture and `texCoord` selects the texture coordinates. Other implementations may do this in
+alternate ways.
```
lilleyse, line 277:
```suggestion
**Note**: The method of selecting a texture is implementation-defined. The above example is based on a glTF implementation where `index` selects a texture and `texCoord` selects the texture coordinates. Other implementations may do this in alternate ways.
```

ptrgags:
@lilleyse updated!

lilleyse:
Thanks!

PR #437, opened 2020-11-19
Get rid of subtree availability buffer.

ErixenCruz:
From what we discussed in meeting https://github.com/CesiumGS/3d-tiles-next/issues/9#issuecomment-729845654. Still thinking about how to incorporate fully (un)availability.


lilleyse:
These changes are now OBE with the latest implicit tiling mockup that isn't yet on github.

PR #440, opened 2020-11-25
Metadata availability for implicit tiling

ErixenCruz:
Partially addresses https://github.com/CesiumGS/3d-tiles-next/issues/100

### DIFF
--- a/extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling.schema.json
+++ b/extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling.schema.json
@@ -1,0 +1,14 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "3DTILES_metadata extension for implicit tiling subtree metadata availability.",
+  "type": "object",
+  "properties": {
+    "class": {
+      "description": "The class referring back to a class definition in 3DTILES_metadata in tileset.json",
+      "type": "string"
+    },
+    "additionalProperties": {
+      "$ref": "3DTILES_metadata_implicit_tiling_object.schema.json"
+    }
+  }
+}
--- a/extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling_object.schema.json
+++ b/extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling_object.schema.json
@@ -1,0 +1,16 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "Implicit tiling metadata availability object.",
+  "type": "object",
+  "properties": {
+    "name": {
+      "type": "string",
+      "description": "The name of the metadata."
+    },
+    "bufferView": {
+      "type": "integer",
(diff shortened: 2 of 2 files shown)


extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling.schema.json
```diff
@@ -0,0 +1,14 @@
...
+  "title": "3DTILES_metadata extension for implicit tiling subtree metadata availability.",
+  "type": "object",
+  "properties": {
+    "class": {
+      "description": "The class referring back to a class definition in 3DTILES_metadata in tileset.json",
+      "type": "string"
+    },
+    "additionalProperties": {
+      "$ref": "3DTILES_metadata_implicit_tiling_object.schema.json"
+    }
```
ptrgags, line 12: @lilleyse I talked with @ErixenCruz about this. we clarified the syntax for `allOf`, however, we had a question, so we didn't commit any changes yet.

Does this extension directly depend on `3DTILES_metadata`? if so, can we reuse the `instanceTable.property.schema.json` and just mention in the spec that JSON encoding is forbidden? 

If not, should we copy in the schema file and just remove the parts we don't use? Not sure if we should use the same filename or not.

extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling_object.schema.json
```diff
@@ -0,0 +1,16 @@
+{
```
ptrgags, line 1: Depending on the answer to the other question, we might want to merge this file with `instanceTable.property.schema.json`, just not sure which is the best way to do that yet.

ptrgags (commented):
@ErixenCruz like we talked about, we need a little more clarification on this PR before I review again.

extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling.schema.json
lilleyse: Yeah it depends on `3DTILES_metadata`, and you can copy `instanceTable.property.schema.json` in the same way that was done for `EXT_feature_metadata` (I forget the specifics, but we would disallow JSON encoding here as well)

lilleyse:
Ultimately I think this should depend on [3DTILES_tile_metadata](https://github.com/CesiumGS/3d-tiles/pull/429) instead of 3DTILES_metadata. However it does make the extension nestedness even worse which is a downside to splitting metadata across multiple extensions.

Maybe it's time to revisit 3DTILES_tile_metadata and get that into shape. Currently it just talks about explicit tiling. This branch would be a branch off of that.

extensions/3DTILES_metadata/0.0.0/schema/3DTILES_metadata_implicit_tiling.schema.json
lilleyse: See https://github.com/CesiumGS/3d-tiles/pull/440#issuecomment-736117919. I think it actually depends on `3DTILES_tile_metadata`, which depends on `3DTILES_metadata`

ErixenCruz:
What does getting 3DTILES_tile_metadata into shape entail? I can work on it. Is there an issue or list somewhere about what should be done?

liuyingbin123:
If we can't get the actual size of tile in memory, we will never be able to accurately control the memory usage of tileset. Because the files on the disk are compressed.the browser will decompress  data in the process of using, which will lead to the increase of the amount of data.

lilleyse:
Will be folded into [3DTILES_metadata](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_metadata)

PR #441, opened 2020-11-25
Implicit tiling with layer availability

ErixenCruz:
Partially addresses https://github.com/CesiumGS/3d-tiles-next/issues/100

### DIFF
--- a/extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers.implicit_tiling_content_availability.schema.json
+++ b/extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers.implicit_tiling_content_availability.schema.json
@@ -9,7 +9,7 @@
     },
     "bufferView": {
       "type": "integer",
-      "description": "Index of a buffer view that indicates whether the layer for each tile in a subtree is available. The bitstream is a 1D boolean array that has an element for each tile. If the layer for a tile is available, its bit is 1, and if it is unavailable, the bit is 0. The number of elements in the array is (N ^ (subtreeDepth + 1) - 1) / (N - 1)."
+      "description": "Index of a buffer view that indicates whether the layer for each tile in a subtree is available. The bitstream is a 1D boolean array that has an element for each tile. If the layer for a tile is available, its bit is 1, and if it is unavailable, the bit is 0. The number of elements in the array is (N ^ (subtreeDepth + 1) - 1) / (N - 1) where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`."
     }
   },
   "required": ["name", "bufferView"]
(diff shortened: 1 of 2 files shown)


extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers.implicit_tiling_content_availability.schema.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema",
+  "title": "Implicit tiling layer availability object.",
+  "type": "object",
+  "properties": {
+    "bufferView": {
```
ptrgags, line 6: @lilleyse is the intent here that this extension depends on `3DTILES_layers` which in turn depends on `3DTILES_binary_buffers`? or is it that this extension depends on both?

extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers_implicit_tiling_subtree_EXT.schema.json
```diff
@@ -0,0 +1,16 @@
+{
```
ptrgags, line 1: @ErixenCruz @lilleyse I think we need to solidify our naming conventions. My understanding of this is that this extension is called `3DTILES_layers_implicit_tiling` which extends `3DTILES_layers` which extends core 3D Tiles,  is this correct?

If this is indeed our first extension of an extension, should the root file be `3DTILES_layers.3DTILES_layers_implicit_tiling.schema.json` or just `3DTILES_layers_implicit_tiling.schema.json`? I don't believe we have a precident

ptrgags (commented):
@ErixenCruz The schema itself looks good, but I have some higher-level questions we should discuss with @lilleyse

ptrgags:
oh yeah one other thing: do you have any mockups for this extension? if so, maybe include a simple example JSON file? Ditto for #440

extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers.implicit_tiling_content_availability.schema.json
lilleyse: Layer content availability, like tile availability and content availability, refers to the buffers/bufferViews that are in the subtree JSON, so no extension needed. 3DTILES_binary_buffers is just for tileset.json while this new subtree JSON is its own self contained thing.

extensions/3DTILES_layers/0.0.0/schema/3DTILES_layers_implicit_tiling_subtree_EXT.schema.json
lilleyse: The mental model I've been thinking is that this is still `3DTILES_layers` but as applied to tilesets using implicit tiling. It wouldn't be a separate `3DTILES_layers_implicit_tiling` extension.

ErixenCruz:
The mockup is in the other PR's branch here https://github.com/CesiumGS/3d-tiles/blob/implicit_tiling_4/extensions/3DTILES_implicit_tiling/0.0.0/example/subtree.json. Should I copy/pasta the file here as well?

lilleyse:
Will be folded into [3DTILES_layers](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_layers)

Issue #341, opened 2018-09-19
MIME type 3D tiles


volkercoors:
In OGC Testbed 14 we are implementing a first prototype that kind of connects 3DPS and WFS 3.0 (under development).

Basically, it is a defintion of requests to get access to specific 3d tiles data sets such as
https://services.thisserverisnotavailableyet.de/t14/wfs3/3D_CityModel_manhattan/collections/buildings/scene?FORMAT=application/json&bbox=-74.00635826977239,40.71778771238832,-73.97393297660074,40.75070138933127

The server shall deliver a 3D scene of Manhattan using 3D tiles in this example. we will use the Fraunhofer server implementation from Testbed 13. The main difference is the structure of the request.

The implementation will be open source as a public project on gitlab:
https://gitlab.com/volkercoors/ogc-testbed-14-wfs-and-3dps

Not much content yet, the entire discussion startet last week at  the OGC TC meeting in Stuttgart.

However, we should have registered (and more specific) MIME type for 3d tiles.

From the spec: https://github.com/AnalyticalGraphicsInc/3d-tiles/tree/master/specification#file-extensions-and-mime-types
3D Tiles uses the following file extensions and MIME types.
 
Tileset files use the .json extension and the application/json MIME type.
Tile content files use the file type and MIME format specific to their tile format specification.
Batched 3D Model tiles use the .b3dm extension and application/octet-stream MIME type.
Tileset style files use the .json extension and the application/json MIME type.

The "octet-stream" subtype is used to indicate that a body contains
arbitrary binary data. https://www.iana.org/assignments/media-types/application/octet-stream 

The problem is that we do not know what we get from the MIME type as it is too general. It basically says JSON and some arbitrary binary data. Do you have any plans to register a MIME type for 3d tiles?

ESRI’s I3S uses the mime type application/vnd.esri.i3s.json+gzip (not registered yet as far as I've seen).

This is helpful for us, as it would allow to specify the delivery format in the request by MIME type.
 
proposal:
application/vnd.3dtiles.json+octet-stream

pjcozzi:
CC previous discussion #60

CC glTF https://github.com/KhronosGroup/glTF/issues/412 and https://github.com/KhronosGroup/glTF/issues/943

lilleyse:
[...]

PR #444, opened 2020-12-07
Implicit tiling spec

ptrgags:
[Direct link to `3DTILES_implicit_tiling` spec](https://github.com/CesiumGS/3d-tiles/tree/implicit-tiling-spec/extensions/3DTILES_implicit_tiling)

Opening this early because I wanted to refer to some of the availability buffer formulas in another issue.

This is the specification to describe the implicit tiling schema we merged into `3d-tiles-next` last week.

I'll update this description once I have more of the text fleshed out.


extensions/3DTILES_implicit_tiling/0.0.0/schema/subtree.schema.json
```diff
@@ -19,17 +19,17 @@
       }
     },
     "tileAvailability": {
-      "description": "The availability of tiles in the subtree. If not all available or unavailable, the size of the buffer will be (N ^ (subtreeDepth + 1) - 1) / (N - 1) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
+      "description": "The availability of tiles in the subtree. If not all available or unavailable, the size of the buffer will be (N ^ (subtreeLevels + 1) - 1) / (N - 1) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
```
ptrgags, line 22: @ErixenCruz I think this formula has an off-by-one error due to how `subtreeLevels` is defined (note that the terminology changed recently) I think it should just be `subtreeLevels` without the +1

Here's my thinking:

| `subtreeLevels` | bits |
|----|---|
| 1 | 1|
| 2 | 1 + 4 = 5 |
| 3 | 1 + 4 + 16 = 21 |
| L | sum for k in `[0, L -1]` of N^k = (N^L - 1) / (N - 1) |

Could you double check this? Then I'll update the schema and tables in the spec.

extensions/3DTILES_implicit_tiling/0.0.0/schema/subtree.schema.json
```diff
@@ -19,17 +19,17 @@
...
     },
     "childSubtreeAvailability": {
-      "description": "The availability of the children of the subtree. If not all available or unavailable, the size of the buffer will be N ^(subtreeDepth) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
+      "description": "The availability of the children of the subtree. If not all available or unavailable, the size of the buffer will be N ^(subtreeLevels) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
       "type": "object",
       "allOf": [{"$ref": "availability.schema.json"}]
     },
     "contentAvailability": {
-      "description": "The availability of the content of the tiles of the subtree. If not all available or unavailable, the size of the buffer will be N ^(subtreeDepth) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
+      "description": "The availability of the content of the tiles of the subtree. If not all available or unavailable, the size of the buffer will be N ^(subtreeLevels) bits where N is 4 for tiling scheme `QUADTREE` and 8 for `OCTREE`.",
```
ptrgags, line 32: Once the other comment is addressed, I'll update this to match `tileAvailability` since these two buffers must match.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -19,19 +19,242 @@ Written against the 3D Tiles 1.0 specification.
...
+
+Tile availability describes what tiles exist at a specific position in space. Though template URIs can represent any possible `(level, x, y)` or `(level, x, y, z)` tile (depending on the tiling scheme), a tileset will typically need only a small subset of the possible tiles. Tile availability explicitly list which tiles are present. This allows querying for tiles while keeping network requests to a minimum.
+
+Tile availability is a bit vector with one bit per node in the [subtree](#subtrees).
+
+Useful formulas for the tile availability buffer:
+
+| Quantity | Formula | Description |
+| -------- | ------- | ----------- |
+| `lengthBits` | `N^subtreeLevels - 1` | Length of buffer is determined by subtree levels
```
ptrgags, line 160: Incorrect, this should be `(N^subtreeLevels - 1) / (N - 1)`

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -19,19 +19,242 @@ Written against the 3D Tiles 1.0 specification.
...
+Useful formulas for the tile availability buffer:
+
+| Quantity | Formula | Description |
+| -------- | ------- | ----------- |
+| `lengthBits` | `N^subtreeLevels - 1` | Length of buffer is determined by subtree levels 
+| `lengthBytes` | `ceil(lengthBits / 8)` | Bytes needed to store the buffer | 
+| `parent.index` | `floor((child.index - 1) / N)` | Index of the parent in the bit vector | 
+| `parent.indexOf(child)` | `(child.index - 1) % N` | Index of the child within the parent's `N` children |
+| `parent.children[k].index` | `N * index + k + 1` | Find the bit of the `k-th` child of a node |
+| `index` | `N^level - 1 + mortonIndex` | Find the index of a node from `(level, mortonIndex)`
```
ptrgags, line 165: I need to double check these formulae involving N^level.

extensions/3DTILES_implicit_tiling/0.0.0/example/backup.json
```diff
@@ -0,0 +1,68 @@
+{
```
ptrgags, line 1: Whoops I didn't mean to commit this.

extensions/3DTILES_implicit_tiling/0.0.0/example/everything.json
```diff
@@ -94,8 +94,8 @@
         "refine": "REPLACE",
         "tilingScheme": "QUADTREE",
         "geometricError": 5000,
-        "subtreeDepth": 7,
-        "maximumDepth": 21,
+        "subtreeLevels": 7,
+        "maximumLevel": 21,
```
ptrgags, line 98: @ErixenCruz these are the two renaming changes of depth -> level for clarity.

`subtreeLevels` is how many distinct levels in the tree there are. So for a quadtree with 2 levels (1 + 4 nodes), `subtreeLevels = 2`.

`maximumLevel` is the 0-indexed level of the last valid level. I know this sounds weird, as "depth" is _technically_ more accurate, but  Cesium 3D Tiles starts at level = 0 so we need to follow this convention.
image

ptrgags (commented):
Added some notes about problems with formulas

ptrgags:
@sanjeetsuhag I just pushed the commits @ErixenCruz and I have been working on. There's still a couple sections and diagrams in progress, but you can review what we have so far

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+| `y` | The 0-indexed number of tiles from the north end of the bounding region in the direction of increasing latitude. This is measured within a single level. |
+| `z` | The 0-indexed number of tiles from the bottom of the bounding region in the direction of increasing height. This is measured within a single level. |
+
+![Region Coordinates](figures/region-coordinates.jpg)
+
+## Template URIs
+
+>**Template URIs** are URI patterns used to refer to specific tiles by their tile coordinates.
+
+Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `${level}`, `${x}`, `${x}`, and for octrees also `${z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```
sanjeetsuhag, line 155:
```suggestion
Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `${level}`, `${x}`, `${y}`, and for octrees also `${z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+
+![Template URI](figures/template-uri.jpg)
+
+## Content
+
+>**Content** is an optional 3D model associated with a single tile, as described in the [Introduction section](https://github.com/CesiumGS/3d-tiles/tree/master/specification#introduction) of the Cesium 3D Tiles 1.0 specification, with the addition of an explicit `mimeType`. 
+
+This extension adds the `mimeType` property to identify the type of content. This is more reliable than a file extension for determining the file type.
+
+To use glTF models (`.gltf` files) or binary glTF models (`.glb` files) as content, the `3DTILES_content_gltf` extension must be listed as a required extension in the tileset.
```
sanjeetsuhag, line 185: This may not belong in this spec.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+- refering to nodes is confusing. review this!
+
+## Availability
+
+**Availability** is boolean data about which tiles, contents, or subtrees exist in a tileset. Availability serves two purposes:
+
+1. It provides an efficient method for checking what files are present
+2. Including ths information prevents extraneous HTTP requests that would result in 404 errors.
+
+Availablity takes the form of a bitstream with one bit per node in consideration. A 1 indicates that a tile/content/subtree is available at this node. Meanwhile, a 0 indicates that no data is available.
```
sanjeetsuhag, line 209:
```suggestion
Availablity takes the form of a bitstream with one bit per node in consideration. A 1 indicates that a tile/content/subtree is available at this node. Meanwhile, a 0 indicates that a tile/content/subtree is not available at this node.
```

I would be explicit about this here, even though it risks being too verbose. "No data" sounded vague to me.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+
+When every node is available or every node is unavailable, all the bits of the bitstream will be identical. That is, either all 1s or all 0s. Instead of storing a full bit stream, the `constant` property can be used instead. For example, `constant: 0` indicates that all bits are 0 and no bitstream must be stored.
```
sanjeetsuhag, line 211:
```suggestion
For cases where all node are available or all nodes are unavailable, the `constant` property should be used instead of storing a full bitstream. For example, `constant: 0` indicates that all bits are 0 and no bitstream must be stored.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+
+### Content Availability
+
+**Content availability** is a bitstream that determines which tiles have an associated content file. Like tile availability, there is one bit for each subtree node. A 1 indicates a content file exists for this tile, while a 0 indicates that no content file exists.
+
+![Content Availability](figures/content-availability.jpg)
+
+The purpose of content availability is to check if a file exists before making a network request. If content is marked as unavailable, the network request for that file must be skipped.
+
+A content availablity bit must only be set if the corresponding tile availability bit is set. Otherwise, it would be possible to specify content files that are not reachable by the tiles of the tileset. The content availability bitstream can be validiated by checking that the following equation holds true:
```
sanjeetsuhag, line 233:
```suggestion
A content availablity bit must only be set if the corresponding tile availability bit is set. Otherwise, it would be possible to specify content files that are not reachable by the tiles of the tileset. The content availability bitstream can be validated by checking that the following equation holds true:
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+
+In the extension object, the following information about the root tile is included:
+
+| Property | Description |
+| ------ | ----------- |
+| `tilingScheme` | Either `QUADTREE` or `OCTREE`|
+| `boundingVolume` | a bounding volume (either a `box` or `region`) describing the root tile |
+| `refine` | Either `ADD` or `REPLACE` as in the Cesium 3D Tiles 1.0 Specification. |
+| `geometricError` | Geometric error of the root tile as described in the Cesium 3D Tiles 1.0 Specification. |
+| `maximumLevel` | Maxium level of the entire tree |
```
sanjeetsuhag, line 298:
```suggestion
| `maximumLevel` | Maximum level of the entire tree |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+| `subtreeLevels` | How many distince levels in each subtree. |
```
sanjeetsuhag, line 299:
```suggestion
| `subtreeLevels` | How many distance levels in each subtree. |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+* **bitstream** - A boolean array stored as a sequence of bits rather than bytes.
+* **child subtree** - A subtree reachable from an available tile in the bottommost row of a subtree.
+* **child subtree availability** - Information about what child subtrees are available.
+* **content availability** - Information about which tiles have an associated content file within a single subtree.
+* **explicit tiling** - Describing a tileset by providing information about every tile.
+* **implicit tiling** - Describing a tileset by providing information about the root tile and a pattern for subdividing the tile.
+* **octree** - A 3D tiling scheme that divides each cuboid into 8 smaller cuboids.
+* **quadtree** - A 2D tiling scheme that divides each rectangle into 4 smaller rectangles.
+* **root tile** - The topmost tile in a tileset tree.
+* **subtree** - A fixed-depth section of the tileset tree used to break large tilesets into managable pieces.
```
sanjeetsuhag, line 360:
```suggestion
* **subtree** - A fixed-depth section of the tileset tree used to break large tilesets into manageable pieces.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+For `box`, the tile coordinates are listed along the same direction as the Cartesian coordinate system of the tileset. That is:
+
+| Coordinate | Description |
+| --- | --- |
+| `level` | The 0-indexed level number. The root tile is at level 0, the root tile's children are at level 1, and so on. |
+| `x` | The 0-indexed number of tiles from the side of the root bounding box in the `+x` direction. This is measured within a single level. |
+| `y` | The 0-indexed number of tiles from the side of the root bounding box in the `+y` direction. This is measured within a single level.|
+| `z` (only for octrees) | The 0-indexed number of tiles from the side of the root bounding box in the `+z` direction. This is measured within a single level. |
+
+![Box coordinates](figures/box-coordinates.jpg)
```
sanjeetsuhag, line 138: Template URIs are being used in this diagram, so we need to either use the tuple form, or move the template URI section up. Another solution could be to add a legend at the top of the diagram indicating how to interpret the tile labels.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+Since tilesets grow exponentially with depth, storing information about every tile in a single file is not always feasible or desirable. Even if RAM is not a direct limitation, streaming large files over the network can make loading times slower. To account for this, subtrees partition the tileset structure into pieces of bounded size.
+
+![exact cover](figures/union-of-subtrees.jpg)
+
+Each subtree is a tree-shaped container for tiles. A subtree has a fixed number of levels defined by the `subtreeLevels` property. This describes the number of distinct levels in the tree. The branching factor is also fixed due to the tiling scheme. For quadtrees, the branching factor is `4`, while octrees have a branching factor of `8`. Taken together, a subtree has exactly enough nodes to store a full quadtree or full octree with a limited depth. However, each node may or may not contain a tile, as a tileset only stores the tiles that are necessary.
+
+![subtree anatomy](figures/subtree-anatomy.jpg)
+
+OUTLINE:
+- refering to nodes is confusing. review this!
```
sanjeetsuhag, line 200:
```suggestion
- referring to nodes is confusing. review this!
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,522 @@
...
+
+When every node is available or every node is unavailable, all the bits of the bitstream will be identical. That is, either all 1s or all 0s. Instead of storing a full bit stream, the `constant` property can be used instead. For example, `constant: 0` indicates that all bits are 0 and no bitstream must be stored.
+
+Availability data is scoped to a subtree. This ensures that the size of each bitstream is bounded to a reasonable size.
+
+### Tile Availability
+
+**Tile availability** is a bitstream that determines which tiles exist within a subtree. There is one bit for each subtree node. A 1 indicates that a tile is available, while a 0 indicates that a tile is unavailable.
+
+![Tile Availability](figures/tile-availability.jpg)
```
sanjeetsuhag, line 219: This diagram may potentially mislead someone into believing that binary trees are supported. We should use quadtrees.

sanjeetsuhag (changes requested):

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: Yeah, my only hesitation is that a quadtree diagram could get quite a bit wider (since we need to show the entire buffer). 

As a group we decided to hold off for now and find a better solution later.

ptrgags:
@sanjeetsuhag updated!

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
 **Version 0.0.0**, December 1, 2020
 
+<!-- omit in toc -->
 ## Contributors
 
 * Sam Suhag, Cesium
 * Sean Lilley, Cesium
 * Peter Gagliardi, Cesium
 * Erixen Cruz, Cesium
+* Patrick Cozzi, Cesium
```
ptrgags, line 13: We should revisit the order of authors now that the text of this draft is written.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
+<!-- omit in toc -->
 # 3DTILES_implicit_tiling
 
 **Version 0.0.0**, December 1, 2020
```
ptrgags, line 4: What date should we put here? We technically finished the draft on the 11th, but we could also wait until the day we merge this PR. 

Just remember to update this on the landing page too.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+  - [Availability Formulas](#availability-formulas)
+    - [Tile and Content Availability Formulas](#tile-and-content-availability-formulas)
+    - [Child Subtree Availability Formulas](#child-subtree-availability-formulas)
 
-## Contents
-Contents go here.
 ## Overview
 
-This extension to 3D Tiles enables implicit tiling. 
+>**Implicit tiling** is a alternative method for describing a Cesium 3D Tileset that provides a more succinct representation for large tilesets. It encodes the hierarchical structure of the tileset in binary, as opposed to explicit tiling, which specifies the tile hierarchy in JSON. The Cesium 3D Tiles 1.0 specification only supports explicit tiling, as every tile is listed in the tileset JSON file.
```
ErixenCruz, line 58:
```suggestion
>**Implicit tiling** is an alternative method for describing a Cesium 3D Tileset that provides a more succinct representation for large tilesets. It encodes the hierarchical structure of the tileset in binary, as opposed to explicit tiling, which specifies the tile hierarchy in JSON. The Cesium 3D Tiles 1.0 specification only supports explicit tiling, as every tile is listed in the tileset JSON file.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+Implicit tiling keeps the tileset JSON file small, which makes loading large tilesets faster. While explicit tiling can be used to represent large datasets, the tileset JSON file grows linearly with the number of tiles. Implicit tiling keeps the tileset JSON file bounded in size.
+
+Implicit tiling also provides a method for accessing tiles by tile coordinates. This allows for abbreviated tree traversal algorithms.
+
+![Explicit vs Implicit Tiling](figures/implicit-vs-explicit.jpg)
```
ErixenCruz, line 64: Should clarify binary tree is for visualization only since we only support quad and oct.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+Implicit tiling also allows for better interoperability with existing GIS data formats with implicitly defined tiling schemes. Some examples are:
+
+* [CDB](https://docs.opengeospatial.org/is/15-113r5/15-113r5.html)
+* [S2](http://s2geometry.io/)
+* [WMTS](https://www.ogc.org/standards/wmts)
+* [TMS](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification)
+
+One new feature implicit tiling enables is procedurally-generated tilesets. Since implicit tiling encodes tile coordinates in URLs (such as `{level}/{x}/{y}/model.gltf`), consider the server that serves these files. Instead of serving static files, a server could extract the tile coordinates from the URL and generate tiles at runtime. This could be useful for making a large procedural terrain dataset without requiring much disk space.
```
ptrgags, line 83: This paragraph could use some rewording. Also we might want to be careful about talking about template URIs since they are not described until a later section.

ptrgags, line 83: Also, probably not wise to mention gltf content, that goes beyond the scope of this spec.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+| Property | Subdivision Rule | 
+| --- | --- |
+| `tilingScheme` | Constant for all tiles in tileset |
+| `refine` | Constant for all tiles in tileset |
+| `boundingVolume` | If `tilingScheme` is `QUADTREE`, the parent tile is divided into 4 child tiles. If `tilingScheme` is `OCTREE` the child tile is divided into 8 child tiles. |
+| `geometricError` | Each child's `geometricError` is half of the parent's `geometricError` |
+
+## Tile Coordinates
+
+>**Tile coordinates** are a tuple of integers that uniquely identify a tile. Tile coordinates are either `(level, x, y)` for quadtrees, and `(level, x, y, z)` for octrees.
```
ErixenCruz, line 126:
```suggestion
>**Tile coordinates** are a tuple of integers that uniquely identify a tile. Tile coordinates are either `(level, x, y)` for quadtrees or `(level, x, y, z)` for octrees.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+Tile coordinates are interpreted differently depending on the type of bounding volume (`box` or `region`).
+
+For `box`, the tile coordinates are listed along the same direction as the Cartesian coordinate system of the tileset. That is:
+
+| Coordinate | Description |
+| --- | --- |
+| `level` | The 0-indexed level number. The root tile is at level 0, the root tile's children are at level 1, and so on. |
+| `x` | The 0-indexed number of tiles from the side of the root bounding box in the `+x` direction. This is measured within a single level. |
+| `y` | The 0-indexed number of tiles from the side of the root bounding box in the `+y` direction. This is measured within a single level.|
+| `z` (only for octrees) | The 0-indexed number of tiles from the side of the root bounding box in the `+z` direction. This is measured within a single level. |
```
ptrgags, line 137: Reword these a little.  should be "number of tiles from the side of the root bounding box in order of increasing x coordinate"

However, if we can find a more concise way to express this, that would be even better.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+| `y` | The 0-indexed number of tiles from the north end of the bounding region in the direction of increasing latitude. This is measured within a single level. |
+| `z` | The 0-indexed number of tiles from the bottom of the bounding region in the direction of increasing height. This is measured within a single level. |
+
+![Region Coordinates](figures/region-coordinates.jpg)
+
+## Template URIs
+
+>**Template URIs** are URI patterns used to refer to specific tiles by their tile coordinates.
+
+Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `${level}`, `${x}`, `${y}`, and for octrees also `${z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```
ptrgags, line 156:
```suggestion
Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `${level}`, `${x}`, `${y}`. Template URIs for octrees must also include `${z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+        "mimeType": "model/gltf+json",
+        "uri": "terrain/${level}/${x}/${y}.gltf"
+      }
+    }
+  }
+}
+```
+## Glossary
+
+* **availability** - Data specifying which tiles are available within a single subtree. This helps prevent unnecessary network requests.
```
ErixenCruz, line 344:
```suggestion
* **availability** - Data specifying which tiles/subtrees/contents are available within a single subtree. This helps prevent unnecessary network requests.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+where `&` is the bitwise AND operation and `~` is the bitwise NOT operation.
+
+### Child Subtree Availability
+
+**Child subtree availability** is a bitstream that determines what subtrees can be reached from this subtree. There are `N` bits for every node in the bottom-most level of the subtree, where `N` is the branching factor of the tree. A 1 means there is a child subtree available at that position in the tree. Meanwhile, a 0 means there is no subtree available.
+
+![Child Subtree Availability](figures/subtree-availability.jpg)
+
+Child subtree availability is used to determine whether further subtree files exist before making network requests. If a child subtree availability bit is 0, any network request for that subtree must be skipped.
```
ptrgags, line 243: I find this wording somewhat ambiguous. "further" isn't the right word here.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+Each subtree JSON file contains the following information:
+
+* The location of a bitstream for tile availability (if not `constant`)
+* The location of a bitstream for content availability (if not `constant`)
+* The location of a bitstream for child subtree availability (if not `constant`)
+
+### Buffers and Buffer Views
+
+OUTLINE:
+- Which spec to reference? 3DTILES_binary_buffers or core metadata?
```
ptrgags, line 258: @sanjeetsuhag @ErixenCruz we should discuss this today.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+>**[Morton order](https://en.wikipedia.org/wiki/Z-order_curve)** assigns indices to nodes in the same level. The indices are used for lookup in availability buffers.
+
+Using the Morton order serves these purposes:
+
+- Efficient tile location decomposition: The Morton order allows efficient encoding and decoding of locations of a tile in the level to its location in the availability buffers.
+- Efficient traversal: The binary representation of tile locations in the tree level allow for easy traversal of the tileset (finding parent and child nodes).
+- Locality of reference: Adjacent indices are stored close to each other in memory and are close to each other spatially.
+
+*The following section is non-normative*
```
ptrgags, line 270: Put this in a subsection like `#### Morton Order Example`

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+The figure below shows the tile location decomposition of the tile `3/5/1` (Level: 3, X: 5, Y: 1). We first convert the tile location to its Morton index, which is `19`. At Level 3 of a Quadtree, we'll use 6 bits to represent the binary value of the Morton index: `010011`.
```
ptrgags, line 272: Use tuple notation when referring to coordinates, as template URIs can be configured by the user.

```suggestion
The figure below shows the tile location decomposition of the tile `(level, x, y) = (3, 5, 1)`. We first convert the tile location to its Morton index, which is `19`. At Level 3 of a Quadtree, we'll use 6 bits to represent the binary value of the Morton index: `010011`.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+
+![Morton Order](figures/morton-indexing.png)
+
+### Availability Encoding
```
ptrgags, line 276: Oh I just realized we never specified alignment rules. Can we inherit them from the metadata specification?

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+      "tilingScheme": "QUADTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 7,
+      "maximumLevel": 21,
+      "subtrees": {
+        "uri": "subtrees/${level}/${x}/${y}/subtree.json"
+      },
+      "content": {
+        "mimeType": "model/gltf+json",
+        "uri": "terrain/${level}/${x}/${y}.gltf"
```
ptrgags, line 336: Let's not refer to glTF, stick to core 3D tiles formats.

```suggestion
        "mimeType": "application/octet-stream",
        "uri": "terrain/${level}/${x}/${y}.b3dm"
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+          |__ model.pnts
+        |__ 1/
+          |__ model.pnts
+```
+
+OUTLINE:
+- double-headed quadtree as explicit root tileset + external implicit tilesets
+- sparse, shallow octree
+- deep, sparse quadtree
+- Full quadtree (medium depth)
```
ptrgags, line 676: We should remove this

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,739 @@
...
+* **subtree node** - A node in a subtree, regardless of whether there is an available tile there.
+* **template URI** - A URI pattern containing tile coordinates for directly addressing tiles.
+* **tile** - A tile as described in the Cesium 3D Tiles 1.0 specification.
+* **tile availability** - Information about which tiles exist within a single subtree.
+* **tileset** - A tileset as described in the Cesium 3D Tiles 1.0 specification.
+* **tileset JSON** - A JSON file describing a tileset, as described in the Cesium 3D Tiles 1.0 specification.
+* **tiling scheme** - A well-defined method for subdividing a bounding volume into a hierarchy of tiles.
+* **unavailable tile** - A region of space that does not contain a tile.
+
+## Examples
```
ptrgags, line 367: I feel like overall we could do better with examples if we had a more visual way of presenting at least one of them. Though as @ErixenCruz and I learned on Friday, putting together labeled examples can be time consuming (since determining morton order is necessary in some diagrams.

ptrgags (commented):
Found several small revisions, but nothing huge.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
sanjeetsuhag: Yes, unless we have a specific reason to deviate from them.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
sanjeetsuhag: An alternative phrasing could be:

```suggestion
Child subtree availability is used to determine whether files for child subtrees exist before making network requests. If a child subtree availability bit is 0, any network request for that subtree must be skipped.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
sanjeetsuhag:
```suggestion
One new feature implicit tiling enables is procedurally-generated tilesets. Since implicit tiling encodes tile coordinates in URLs using [Template URIs](#template-uris), consider the server that serves these files. Instead of serving static files, a server could extract the tile coordinates from the URL and generate tiles at runtime. This could be useful for making a large procedural terrain dataset without requiring much disk space.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
sanjeetsuhag: Instead of adding verbosity in the coordinate descriptions, consider adding some text immediately before or after this table noting that all coordinates are 0 indexed.

extensions/3DTILES_implicit_tiling/0.0.0/schema/3DTILES_implicit_tiling.schema.json
```diff
@@ -23,19 +23,19 @@
         "description": "The error, in meters, introduced if the root is rendered and its children are not. The geometric error is halved for each level of subdivision. At runtime, the geometric error is used to compute screen space error (SSE), i.e., the error measured in pixels.",
         "minimum" : 0
       },
-      "subtreeDepth": {
+      "subtreeLevels": {
         "type": "integer",
-        "description": "The depth of each subtree.",
+        "description": "The number of distinct levels in each subtree. For example, a quadtree with `subtreeLevels = 2` will have subtrees with up to 5 nodes (one root and 4 children)",
```
ErixenCruz, line 28:
```suggestion
        "description": "The number of distinct levels in each subtree. For example, a quadtree with `subtreeLevels = 2` will have subtrees with 5 nodes (one root and 4 children)",
```

ptrgags:
Note: we should be careful about subtree diagrams. subtree leaf nodes should point to the _subtree_ below them, not a node directly.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: This should reference 3DTILES_binary_buffers, though that [draft spec](https://github.com/CesiumGS/3d-tiles/tree/3DTILES_binary_buffers/extensions/3DTILES_binary_buffers) doesn't really explain what buffers/bufferViews are...

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+- content/3/2/1/0.pnts
+```
+
+Unless otherwise specified, template URIs are resolved relative to the tileset.json file.
+
+![Template URI](figures/template-uri.jpg)
+
+## Content
+
+**Content** is an optional 3D model associated with a single tile, as described in the [Introduction section](https://github.com/CesiumGS/3d-tiles/tree/master/specification#introduction) of the Cesium 3D Tiles 1.0 specification, with the addition of an explicit `mimeType` to reliably identify the type of content.
```
lilleyse, line 306: It's unfortunate that b3dm, i3dm, pnts, and cmpt don't have unique mime types yet which makes it hard to actually identify the type of content. CC https://github.com/CesiumGS/3d-tiles/issues/341

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+## Subtrees
+
+**Subtrees** are fixed-sized sections of the tileset tree used for breaking tilesets into manageable pieces.
+
+Since tilesets grow exponentially with depth, storing information about every tile in a single file is not always feasible or desirable. Even if RAM is not a direct limitation, streaming large files over the network can make loading times slower. To account for this, subtrees partition the tileset structure into pieces of bounded size.
+
+![exact cover](figures/union-of-subtrees.jpg)
+
+A subtree has a fixed number of levels defined by the `subtreeLevels` property. This describes the number of distinct levels in the tree. The number of children per tile is also fixed due to the subdivision scheme. For quadtrees, there are `4` children per tile, while octrees have `8` children per tile. Taken together, a subtree has exactly enough tiles to store a full quadtree or full octree with a limited number of levels. However, each tile may or may not exist or contain content, as a tileset only stores the tiles that are necessary.
```
lilleyse, line 316:
```suggestion
A subtree has a fixed number of levels defined by the `subtreeLevels` property. This describes the number of distinct levels in the subtree. The number of children per tile is also fixed due to the subdivision scheme. For quadtrees, there are `4` children per tile, while octrees have `8` children per tile. Taken together, a subtree has exactly enough tiles to store a full quadtree or full octree with a limited number of levels. However, each tile may or may not exist or contain content, as a tileset only stores the tiles that are necessary.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+![subtree anatomy](figures/subtree-anatomy.jpg)
```
lilleyse, line 318: This diagram might be slightly more readable if the arrows and the outer triangle are thinner

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+## Availability
+
+**Availability** is a boolean that defines whether a tile, content, or subtree exist in a tileset. Availability serves two purposes:
+
+1. It provides an efficient method for checking what files are present
+2. Including ths information prevents extraneous HTTP requests that would result in 404 errors.
```
lilleyse, line 325:
```suggestion
2. Including this information prevents extraneous HTTP requests that would result in 404 errors.
```

lilleyse, line 325: The second bullet seems more like an example of the first bullet rather than a use case by itself

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+Availability takes the form of a bitstream with one bit per node in consideration. A 1 indicates that a tile/content/subtree is available at this node. Meanwhile, a 0 indicates that no tile/content/subtree is available at this node.
```
lilleyse, line 327: Even though availability works the same way tile/content/subtree it might be confusing to describe them all together like this because it makes me think each node in the tree has subtree availability.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+For cases where all nodes are available or all nodes are unavailable, the `constant` property should be used instead of storing a full bitstream. For example, `constant: 0` indicates that all bits are 0 and no bitstream must be stored.
```
lilleyse, line 329: A JSON example might be helpful here. `constant: 1` might be a better example because it's a more common use case for the tree to be full rather than empty, though `constant: 0` does have its use cases too, like a tree where content only shows up deeper in the tree.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+Availability data is scoped to a subtree. This ensures that the size of each bitstream is bounded to a reasonable size.
+
+### Tile Availability
+
+**Tile availability** is a bitstream that determines which tiles exist within a subtree. There is one bit for each subtree node. A 1 indicates that a tile is available, while a 0 indicates that a tile is unavailable.
+
+![Tile Availability](figures/tile-availability.jpg)
```
lilleyse, line 337: The letters on the nodes are a little small and hard to read. Maybe instead of a gradient palette use a palette of different light colors.

lilleyse, line 337: Just to confirm, is the tile availability array in morton order here? For binary trees is morton order the same as being ordered left-to-right?

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+In the diagram above, colored nodes indicate available tiles, while nodes with dashed outlines are unavailable tiles.
+
+If a tile is marked as available, more information about the tile, such as its content or children can be queried. If a tile is marked as unavailable, the tile must be skipped.
```
lilleyse, line 341: There are some rules that are included in the schema that should also be mentioned in the spec text. One example is "If a non-root tile's availability is 1 its parent tile's availability must also be 1."

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+![Content Availability](figures/content-availability.jpg)
+
+The purpose of content availability is to check if a file exists before making a network request. If content is marked as unavailable, the network request for that file must be skipped.
+
+A content availability bit must only be set if the corresponding tile availability bit is set. Otherwise, it would be possible to specify content files that are not reachable by the tiles of the tileset. The content availability bitstream can be validated by checking that the following equation holds true:
+
+```
+contentAvailability & ~tileAvailability === 0
+```
```
lilleyse, line 355: Personally I don't think the the validation equation needs to be included since the explanation in the previous sentences is already pretty clear.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+where `&` is the bitwise AND operation and `~` is the bitwise NOT operation.
+
+### Child Subtree Availability
+
+**Child subtree availability** is a bitstream that determines what subtrees can be reached from this subtree. There are `N` bits for every node in the bottom-most level of the subtree, where `N` is the branching factor of the tree. A 1 means there is a child subtree available at that position in the tree. Meanwhile, a 0 means there is no subtree available.
```
lilleyse, line 361: "what" -> "which"?

```suggestion
**Child subtree availability** is a bitstream that determines which subtrees can be reached from this subtree. There are `N` bits for every node in the bottom-most level of the subtree, where `N` is the branching factor of the tree. A 1 means there is a child subtree available at that position in the tree. Meanwhile, a 0 means there is no subtree available.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+A subtree has a fixed number of levels defined by the `subtreeLevels` property. This describes the number of distinct levels in the tree. The number of children per tile is also fixed due to the subdivision scheme. For quadtrees, there are `4` children per tile, while octrees have `8` children per tile. Taken together, a subtree has exactly enough tiles to store a full quadtree or full octree with a limited number of levels. However, each tile may or may not exist or contain content, as a tileset only stores the tiles that are necessary.
+
+![subtree anatomy](figures/subtree-anatomy.jpg)
+
+## Availability
+
+**Availability** is a boolean that defines whether a tile, content, or subtree exist in a tileset. Availability serves two purposes:
+
+1. It provides an efficient method for checking what files are present
```
lilleyse, line 324: "what" -> "which"?

```suggestion
1. It provides an efficient method for checking which files are present
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+The figure below shows the tile location decomposition of the tile `(level, x, y) = (3, 5, 1)`. We first convert the tile location to its Morton index, which is `19`. At Level 3 of a Quadtree, we'll use 6 bits to represent the binary value of the Morton index: `010011`.
+
+![Morton Order](figures/morton-indexing.png)
+
+### Availability Encoding
+
+Availability bitstreams are packed in binary using the format described in the [Boolean Data section](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/specification/Metadata/0.0.0/README.md#boolean-data) of the Cesium 3D Metadata Specification. These bitstreams are subject to alignment requirements described in the [Binary Alignment Rules section](https://github.com/CesiumGS/3d-tiles/tree/3DTILES_binary_buffers/extensions/3DTILES_binary_buffers) of the same specification.
+
+Each availability bitstream must be stored as a separate `bufferView`, but multiple `bufferViews` may be stored in a single `buffer`.
```
lilleyse, line 404:
```suggestion
Each availability bitstream must be stored as a separate `bufferView`, but multiple `bufferViews` may refer to a single `buffer`.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+In the extension object, the following information about the root tile is included:
+
+| Property | Description |
+| ------ | ----------- |
+| `tilingScheme` | Either `QUADTREE` or `OCTREE`|
+| `boundingVolume` | a bounding volume (either a `box` or `region`) describing the root tile |
+| `refine` | Either `ADD` or `REPLACE` as in the Cesium 3D Tiles 1.0 Specification. |
+| `geometricError` | Geometric error of the root tile as described in the Cesium 3D Tiles 1.0 Specification. |
+| `maximumLevel` | Maximum level of the entire tree |
```
lilleyse, line 418: Is this actually the maximum level of the last available tile? Or last available content?

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+## Tileset JSON
+
+Using implicit tiling, the tileset JSON file describes the root tile, tiling scheme, and template URIs for locating resources.
+
+In the extension object, the following information about the root tile is included:
+
+| Property | Description |
+| ------ | ----------- |
+| `tilingScheme` | Either `QUADTREE` or `OCTREE`|
+| `boundingVolume` | a bounding volume (either a `box` or `region`) describing the root tile |
```
lilleyse, line 415:
```suggestion
| `boundingVolume` | A bounding volume (either a `box` or `region`) describing the root tile |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+| `refine` | Either `ADD` or `REPLACE` as in the Cesium 3D Tiles 1.0 Specification. |
+| `geometricError` | Geometric error of the root tile as described in the Cesium 3D Tiles 1.0 Specification. |
+| `maximumLevel` | Maximum level of the entire tree |
+| `subtreeLevels` | How many distance levels in each subtree. |
+
+Furthermore, template URIs are used for locating subtree JSON files as well as tile contents. The key properties are as follows:
+
+| Option | Description |
+| ------ | ----------- |
+| `subtrees.uri` | template URI for a subtree JSON file. see [Subtrees](#subtrees) for more info |
```
lilleyse, line 425:
```suggestion
| `subtrees.uri` | Template URI for a subtree JSON file. See [Subtrees](#subtrees) for more info |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+| `content.uri` | template URI for the content 3D Models |
```
lilleyse, line 426: Be careful not to use "3D Models" terminology too much because someday we may support other content types like rasters

```suggestion
| `content.uri` | Template URI for the content |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+        "uri": "subtrees/${level}/${x}/${y}/subtree.json"
+      },
+      "content": {
+        "mimeType": "application/octet-stream",
+        "uri": "terrain/${level}/${x}/${y}.b3dm"
+      }
+    }
+  }
+}
+```
```
lilleyse, line 465: It would be helpful to also see a subtree JSON alongside this

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+        "draftVersion": "0.0.0"
+      },
+      "boundingVolume": {
+        "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+      },
+      "refine": "REPLACE",
+      "tilingScheme": "QUADTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 7,
+      "maximumLevel": 21,
```
lilleyse, line 454: While it's totally valid for `maximumLevel` to be 20 or 21, I think the example should be 20 so that three subtrees neatly fit into the tileset JSON. With 21 there is spillover into a fourth layer of subtrees.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+        "draftVersion": "0.0.0"
+      },
+      "boundingVolume": {
+        "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+      },
+      "refine": "ADD",
+      "tilingScheme": "QUADTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 2,
+      "maximumLevel": 4,
```
lilleyse, line 515: Same as the comment above, consider changing this to 3

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+    "bufferView": 0
+  },
+  "contentAvailability": {
+    "constant": 0
+  },
+  "childSubtreeAvailability": {
+    "bufferView": 1
+  }
+}
+```
```
lilleyse, line 698: The formatting here is a little off

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+        |__ 0/
+          |__ model.pnts
+        |__ 1/
+          |__ model.pnts
+      |__ 1/
+        |__ 0/
+          |__ model.pnts
+        |__ 1/
+          |__ model.pnts
+```
```
lilleyse, line 786: While this is valid, a more common template uri would be `"models/${level}/${z}/${y}/${x}.pnts"`. It's a little weird to have so many `model.pnts` files in practice. It also makes this diagram bigger than it needs to be.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+  
+## Appendix A: Tree Indexing Formulas
+
+### Morton Indexing Formulas
+
+| Quantity | Formula | Description |
+| -------- | ------- | ----------- |
+| `N` | 4 or 8 | N is 4 for quadtrees, 8 for octrees |
+| `bits` | `log2(N)` | Quadtree address are a multiple of 2 bits, Octrees use a multiple of 3 bits | 
+| `mortonIndex` | `interleave(z, y, x)` or `interleave(y, x)` | The morton index is computed by interleaving bits. see below. |
```
lilleyse, line 801:
```suggestion
| `mortonIndex` | `interleave(z, y, x)` or `interleave(y, x)` | The morton index is computed by interleaving bits. See below. |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+| `length(mortonIndex)` | `level * bits` | length of morton index in bits
```
lilleyse, line 802:
```suggestion
| `length(mortonIndex)` | `level * bits` | Length of morton index in bits
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+| -------- | ------- | ----------- |
+| `lengthBits` | `(N^subtreeLevels - 1)/(N - 1)` | Length of buffer is determined by subtree levels |
+| `lengthBytes` | `ceil(lengthBits / 8)` | Bytes needed to store the buffer | 
+| `parent.index` | `floor((child.index - 1) / N)` | Index of the parent in the bitstream | 
+| `parent.indexOf(child)` | `(child.index - 1) % N` | Index of the child within the parent's `N` children |
+| `parent.children[k].index` | `N * index + k + 1` | Find the bit of the `k-th` child of a node |
+| `index` | `(N^level - 1)/(N - 1) + mortonIndex` | Find the index of a node from `(level, mortonIndex)`
+| `level` | `ceil(log(index + 1)/log(N))` | Find the level of a node relative to the subtree |
+| `globalLevel` | `level + subtreeRoot.globalLevel` | Find the level of a node relative to the entire tileset | 
+| `startOfLevel` | `(N^level - 1)/(N - 1)` | first index at a particular level (relative to the subtree root) |
```
lilleyse, line 836:
```suggestion
| `startOfLevel` | `(N^level - 1)/(N - 1)` | First index at a particular level (relative to the subtree root) |
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+#### Child Subtree Availability Formulas
 
-## Schema Changes
-Schema changes go here.
+| Quantity | Formula | Description |
+| -------- | ------- | ----------- |
+| `lengthBits` | `N^subtreeLevels` | Length of the buffer by subtree levels |
+| `lengthBytes` | `ceil(lengthBits / 8)` | Bytes needed to store the buffer |
+| `childSubtree.globalLevel` | `subtreeRoot.globalLevel + subtreeLevels` | Level of the child subtrees relative to the tileset root |
+| `leaf.children[k].index` | `N * leaf.mortonIndex + k` | index of the `k-th` child subtree |
```
lilleyse, line 847:
```suggestion
| `leaf.children[k].index` | `N * leaf.mortonIndex + k` | Index of the `k-th` child subtree |
```

lilleyse (commented):
I started looking at this last week and @IanLilleyT joined in since he was interested in how this relates to his work. Here's comments from both of us.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ErixenCruz: Yes I believe so.

IanLilleyT:
Are `subdivisonScheme` and `tilingScheme` the same thing? If so, which is the right one to use?

ErixenCruz:
Yes they are the same. we used `tilingScheme` at first but then decided to use subdivision instead since its clearer. havent changed all the mentions of `tilingScheme` yet

IanLilleyT:
We should consider doing `{...}` instead of `${...}` for URI variable names.

`content/${level}/${z}/${y}/${x}.pnts` ->  `content/{level}/{z}/{y}/{x}.pnts`

This is what [CesiumJS Resource](https://cesium.com/docs/cesiumjs-ref-doc/Resource.html) expects. It's also used by terrain's `layer.json`.

[mapserver does it like this way too](https://mapserver.org/mapcache/config.html#cache)

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: trivally, yes binary trees are in morton order when numbered from left-to-right. `mortonIndex = interleave(z, y, x)` but if you only have a single `x` coordinate, than `mortonIndex = interleave(x) = x`

That said, this gets back to the question of whether we should use binary trees at all in the spec. They're nice for illustration purposes since they fit better on the page, quadtrees tend to require wide diagrams. That said, in order to see the morton order, it really is non-obvious unless you use a quadtree...

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+      "maximumLevel": 21,
+      "subtrees": "subtrees/{level}/{x}/{y}/subtree.json",
+      "content": {
+        "mimeType": "application/octet-stream",
+        "uri": "terrain/{level}/{x}/{y}.b3dm"
+      }
+    }
+  }
+}
+{
```
IanLilleyT, line 434: copy paste mistake?

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+  "root": {
+    "boundingVolume": {
+      "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+    },
+    "refine": "REPLACE",
+    "geometricError": 5000,
+    "extensions": {
+      "3DTILES_implicit_tiling": {
+        ...
+      }
```
IanLilleyT, line 110: Missing `}` on line below

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+* [S2](http://s2geometry.io/)
+* [WMTS](https://www.ogc.org/standards/wmts)
+* [TMS](https://wiki.osgeo.org/wiki/Tile_Map_Service_Specification)
+
+Implicit tiling enables procedurally-generated tilesets. Instead of serving static files, a server could extract the tile coordinates from [Template URIs](#template-uris) and generate tiles at runtime while using little disk space.
+
+## Tileset JSON
+Like in [3D Tiles 1.0](https://github.com/CesiumGS/3d-tiles/tree/master/specification#tileset-json), one main tileset JSON file is the entry point for defining an implicit tileset. To use implicit tiling, the `3DTILES_implicit_tiling` extension must be defined in the root tile of the `tileset.json`.
+```json
+{
```
IanLilleyT, line 91: the tileset geometricError is missing

IanLilleyT (changes requested):

extensions/3DTILES_implicit_tiling/0.0.0/README.md
IanLilleyT: This problem shows up in a few other places within the document

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+## Examples
+Here are some complete examples for how to create commonly used data structures with implicit tiling.
+### Quadtree with four levels
+Consider a tileset with a quadtree tiling scheme and four levels of detail. Suppose that we want to use implicit tiling with subtrees with 2 levels.
+
+![Quadtree example](figures/quadtree-example.jpg)
+
+The root tileset JSON might look something this:
+```json
+{
```
IanLilleyT, line 490: This tileset is missing `"root"`

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+}
+```
+This subtree at the bottom of the tree is completely full. It uses constants for its availabilities instead of buffers. Because it is at the bottom of the tree, there are no child subtrees, so child subtree availability is a constant zero.
+
+## Octree Example
+
+`tileset.json`:
+
+```json
+{
```
IanLilleyT, line 621: also missing `"root"`

IanLilleyT (changes requested):

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+| `y` | From south to north (increasing latitude) |
+| `z` | From bottom to top (increasing height) |
+
+![Region Coordinates](figures/region-coordinates.jpg)
+
+## Template URIs
+
+A **Template URI** is a URI pattern used to refer to tiles by their tile coordinates.
+
+Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `{level}`, `{x}`, `{y}`. Template URIs for octrees must also include `${z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```
ptrgags, line 219: the `${z}` should be `{z}`

```suggestion
Template URIs are configured in the tileset.json. They may be any URI pattern, but must include the variables `{level}`, `{x}`, `{y}`. Template URIs for octrees must also include `{z}`. When referring to a specific tile, the tile's coordinates are substituted in for these variables.
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: On the other hand, if we convert our formats to glTF, they'd all have the same `mimeType` anyway... See https://github.com/CesiumGS/3d-tiles-next/issues/91#issuecomment-761122165. still good to have the mimeType though, but might be easier to enforce in 2.0.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+      "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+    },
+    "refine": "REPLACE",
+    "geometricError": 5000,
+    "extensions": {
+      "3DTILES_implicit_tiling": {
+        "subdivisionScheme": "QUADTREE",
+        "content": {
+          "mimeType": "application/octet-stream",
+          "uri": "terrain/${level}/${x}/${y}.b3dm"
```
ptrgags, line 268:
```suggestion
          "uri": "terrain/{level}/{x}/{y}.b3dm"
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
```
ptrgags, line 455:
```suggestion
          "uri": "terrain/{level}/{x}/{y}.b3dm"
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+      },
+      "boundingVolume": {
+        "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+      },
+      "refine": "ADD",
+      "subdivisionScheme": "QUADTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 2,
+      "maximumLevel": 4,
+      "subtrees": "subtrees/${level}/${x}/${y}/subtree.json",
```
ptrgags, line 514:
```suggestion
      "subtrees": "subtrees/{level}/{x}/{y}/subtree.json",
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+      "content": {
+        "mimeType": "application/octet-stream",
+        "uri": "models/${level}/${x}/${y}.b3dm"
```
ptrgags, line 517:
```suggestion
        "uri": "models/{level}/{x}/{y}.b3dm"
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+          0, 0, 100
+        ]
+      },
+      "refine": "ADD",
+      "subdivisionScheme": "OCTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 3,
+      "maximumLevel": 8,
+      "subtrees": {
+        "uri": "subtrees/${level}/${z}/${y}/${x}/subtree.json"
```
ptrgags, line 651:
```suggestion
        "uri": "subtrees/{level}/{z}/{y}/{x}/subtree.json"
```

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,845 @@
...
+      },
+      "content": {
+        "mimeType": "application/octet-stream",
+        "uri": "models/${level}/${z}/${y}/${x}/model.pnts"
```
ptrgags, line 655:
```suggestion
        "uri": "models/{level}/{z}/{y}/{x}/model.pnts"
```

extensions/3DTILES_implicit_tiling/0.0.0/example/sparseOctree/tileset.json
```diff
@@ -0,0 +1,42 @@
...
+          241.6
+        ]
+      },
+      "refine": "ADD",
+      "tilingScheme": "OCTREE",
+      "geometricError": 5000,
+      "subtreeLevels": 7,
+      "maximumLevel": 14,
+      "subtrees": {
+        "uri": "subtrees/${level}/${x}/${y}/${z}.json"
```
ptrgags, line 34: This goes for all the other tileset JSONs.

```suggestion
        "uri": "subtrees/{level}/{x}/{y}/{z}.json"
```

ptrgags: @ErixenCruz did you generate any of these tilesets via a script? if so, could you check the script in so we can update it to fix this change?

extensions/3DTILES_implicit_tiling/0.0.0/schema/3DTILES_implicit_tiling.schema.json
```diff
@@ -4,38 +4,24 @@
...
         "minimum" : 1
       },
-      "maximumDepth": {
+      "maximumLevel": {
         "type": "integer",
         "description": "The maximum depth of the tree.",
         "minimum" : 0
       },
       "subtrees": {
         "allOf": [{"$ref": "templateUri.schema.json"}],
```
ptrgags, line 23: not in this file, but in `templateUri.schema.json`: Update it to use `{x}` syntax.

ptrgags (commented):
It seems like most of us want `{x}` rather than `${x}` so pointing out some places where this needs to be updated. there were many, so check for things I missed.

extensions/3DTILES_implicit_tiling/0.0.0/example/sparseOctree/tileset.json
ErixenCruz: No, I did not use a script for this.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: tile. I updated the text.

IanLilleyT:
Should
```
  "tileAvailability": {
    "constant": 0
  }
```
be explicitly forbidden?

Because if it were 0, there would be no point in having a subtree in the first place.

Another benefit of this constraint is if you have tile metadata, it's guaranteed that a tile metadata buffer always exists.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: Resolving as there's nothing actionable until 2.0

extensions/3DTILES_implicit_tiling/0.0.0/README.md
ptrgags: No, we're just going to describe this inline. If possible we want only implicit tiling to need buffer views.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,849 @@
...
+
+For `box`, the tile coordinates are listed along the same direction as the Cartesian coordinate system of the tileset. That is:
+
+| Coordinate | Positive Direction |
+| --- | --- |
+| `x` | Along the `+x` axis of the bounding box |
+| `y` | Along the `+y` axis of the bounding box |
+| `y` | Along the `+z` axis of the bounding box |
+
+![Box coordinates](figures/box-coordinates.jpg)
```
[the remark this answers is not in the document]
ptrgags: so far addressed the double-labeling and curved arrows. Third image (region bounding volume) needs to be redone to be more curved. But that's lower on the priority list, the focus should be on schema/spec changes.

ptrgags:
@IanLilleyT @sanjeetsuhag @ErixenCruz @lilleyse 

Changelist from this week's round of implicit tiling schema/spec updates:

* subtree files are binary now! the format is similar to GLB, but a bit more slimmed down (no JSON/BIN chunk headers)
* the description of buffers/bufferViews is now included in this spec since we're getting rid of 3DTILES_binary_buffers. This description includes the 8-byte alignment requirement.
* clarified some wording about the size of `childSubtreeAvailability`
* clarification: Levels are measured starting at 0 at the root implicit tile, i.e. the tile with the `3DTILES_implicit_tiling` extension. This clarification is necessary since an implicit tileset may be part of a larger explicit tileset.
* `3DTILES_implicit_tiling.content` is optional, and when present, `subtree.contentAvailability` is required. If `content` is not specified, then `contentAvailability` is disallowed.
* remove `mimeType`, it won't be useful until 2.0, and even then it's not a complete solution. 
* Fixed some examples that still were putting the extension in the tileset rather than the root tile
* Added some details about availability that were present in the schema but not the spec.

What has NOT been changed (yet):
* the text of the spec still needs work. better diagrams and examples and a couple organization details. Full details in https://github.com/CesiumGS/3d-tiles-next/issues/119#issuecomment-746879326
* the example files need to be fixed, I need to transform the `subtree.json` files into binary files. I plan to do this early next week to avoid confusion.

Any questions/comments, please let me know.

extensions/3DTILES_implicit_tiling/0.0.0/README.md
```diff
@@ -1,37 +1,879 @@
...
+
+If content availability is 0 and its corresponding tile availability is 1 then the tile is considered to be an empty tile.
+
+When a subtree has at least one tile with content, content availability is required. If no tile in the subtree has content, then content availability is disallowed.
+
+### Child Subtree Availability
+
+**Child subtree availability** is a bitstream that determines which subtrees can be reached from the deepest level of this subtree.
+
+The child subtree availability bitstream has slightly different structure than tile or content availability. There is one bit for each node in the level of the tree immediately below the subtree. That is, if the deepest subtree is level `L`, then there is one bit for every node at level `L + 1`. Since each node at level `L` has `N` children (4 for `QUADTREE` or 8 for `OCTREE`), there are `N * (L + 1)` bits in the child subtree availability bitstream. A 1 bit means there is a child subtree available at that position in the tree. A 0 bit means there is no subtree available at that position.
```
IanLilleyT, line 334: * `That is, if the deepest subtree`. Should it be `deepest subtree node`?
* I assume `L` is a level local to the subtree, but it could be read as a global level.
* The equation `N * (L + 1)` should be `N ^ (L + 1)`.
* No mention of morton encoding
 
I feel like this paragraph would be more clear if defined in terms of `subtreeLevels`

lilleyse:
To keep the in-progress extensions in lockstep we're going to merge this PR into the [3d-tiles-next](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next) staging branch and work from there.

PR #121, opened 2016-08-23
i3dm fixes

lasalvavida:
Fixes for #120

PR #447, opened 2020-12-29
remove confusing header.gltfByteLength

renjianqiang:
I am confused with `header.gltfByteLength` in `i3dm`, and i find it was removed in here #121. So i removed the `header.gltfByteLength` in other place to avoid ambiguity.


lilleyse:
Great catch @renjianqiang

PR #450, opened 2021-02-02
Update osgjs link

dbuck:
Please verify whether this is the new correct link.
Previous link forwards... suspiciously to a russian adult site.


dbuck:
though on second glance, it appears that the project is archived, unless there is a more recent link.

lilleyse:
Thanks @dbuck, I pushed the change directly so we don't have to go through signing a CLA to get this merged. https://github.com/CesiumGS/3d-tiles/commit/afa1fa3815c3e3bb97fe26a5e5665186702743ba

PR #454, opened 2021-02-18
3DTILES_content_gltf spec revisions

ptrgags:
This updates the `3DTILES_content_gltf` spec to include an inline JSON example, JSON schema reference, and expand on the overview.

@lilleyse could you review when you get a chance?


lilleyse:
Looks good!

PR #455, opened 2021-02-28
Cesium 3D Metadata Specification Deep Dive

pjcozzi:
This is a pull request so that I can provide deep dive feedback on the draft **Cesium 3D Metadata Specification**.

This pull request is intended for comments only, not for merging.

It has been a few months since I've looked at this so I apologize in advance if some of my questions are repeats or beginner questions. 😄


pjcozzi (commented):
@ptrgags @lilleyse this is close.

Here's some general themes I extracted from the specific comments:

## Big Picture

* Is this an Abstract Language spec that we wrote because of our metadata needs, e.g., this isn't really a metadata spec?  It's kinda like a more well defined JSON that allows other encodings.  Would we use this for a future CZML, data-driven UI, app frameworks, etc.?  If yes or maybe then let's discuss renaming this.
* It's not clear to me how other spec writers know if they robustly conform to this spec; maybe it will become clear to me as we cross t's and dot i's in this spec, and as I get a refresher on the other specs that reference this.
   * Avoid **definition by example**
   * **Raster Format** encoding is confusing - see comments and let's discuss
   * Ditto - but less so - for the **JSON encoding**

## Spec Writing
* **Intro**.  Sweat the details. Concise and context.
* **Vocab**. Precise and consistent
* **Diagrams**. For context and to help the novice reader built their mental model of how the vocab comes together
* **Easing in the novice reader**.  Concise and precise, avoid distracting them with unimportant or untimely details.

Also, is this both minimal and complete?  Is there anything left that we could potentially cut?
 
For many of my comments, such as the many vocab questions, I don't necessarily care what you decide, just that it is intentional and aligned to the ecosystem.

specification/Metadata/1.0.0/README.md
```diff
@@ -0,0 +1,688 @@
...
+
+## Overview
+
+The Cesium 3D Metadata Specification defines a standard metadata format for 3D data. This spans a myriad of domains, such as heights of buildings in a city, names of different structures in a CAD model, material properties for textured surfaces, and classification codes for point clouds.
+
+This specification defines a set of core concepts to be used by multiple 3D formats. It is language and format independent. This document occasionally uses JSON examples for illustrating concepts but does not define a JSON schema.
+
+For usage see:
+
+* [`3DTILES_metadata`](../../../extensions/3DTILES_metadata/1.0.0) - 3D Tiles extension that assigns metadata to various components of 3D Tiles
```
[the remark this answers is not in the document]
ptrgags: To be consistent with other wording of these specs, we think that it would be good to talk about "at different levels of granularity" and then give examples (tileset, tile, feature)

specification/Metadata/1.0.0/README.md
```diff
@@ -0,0 +1,688 @@
...
+BOOLEAN | True or false
+STRING | A sequence of characters
+ENUM | An enumerated type
+ARRAY | An array of components
+
+### Arrays
+
+The `ARRAY` type is used to define fixed- or variable-length arrays. Each element in the array is called a **component**.
+
+An array property must declare a component type, which is the type of each component in the array. A component type may be any type from the table above with the exception of `ARRAY`; arrays of arrays are not supported.
```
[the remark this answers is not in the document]
ptrgags: Ah, it should be "define" everywhere.

specification/Metadata/1.0.0/README.md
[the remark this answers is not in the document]
pjcozzi: Sounds good.

specification/Metadata/README.md
```diff
@@ -0,0 +1,699 @@
...
+    - [Enums](#enums-3)
+    - [Arrays](#arrays-3)
+    - [Implementation Notes](#implementation-notes)
+- [Revision History](#revision-history)
+
+## Overview
+
+The Cesium 3D Metadata Specification defines a standard metadata format for 3D data. This spans a myriad of domains, such as heights of buildings in a city, names of different structures in a CAD model, material properties for textured surfaces, and classification codes for point clouds.
+
+This specification defines a set of core concepts to be used by multiple 3D formats. It is language and format independent. This document occasionally uses JSON examples for illustrating concepts but does not define a JSON schema.
```
donmccurdy, line 57: While the document "does not define a JSON schema", there is really quite a lot of schema here. 😅

If I understand correctly, the purpose of this document is to define the primitive data types and relationships that will be used in `3DTILES_metadata` and `EXT_feature_metadata`, in such a way that they can be (1) referenced in both, and (2) trivially portable and compatible between the two. Is that correct?

Assuming so, in the interest of clearer delineation between the three specifications, do you think it would be reasonable to replace all or most of the JSON schemas (in this document) with structured text equivalents? For example:

**Before:**

```json
{
  "schema": {
    "classes": {
      "car": {
        "properties": {
          "forwardDirection": {
            "description": "A forward direction vector",
            "type": "ARRAY",
            "componentType": "FLOAT64",
            "componentCount": 3
          },
          "passengers": {
            "name": "Passenger Names",
            "description": "There are a variable number of passengers because componentCount is undefined.",
            "type": "ARRAY",
            "componentType": "STRING",
          },
          "modelMatrix": {
            "description": "A 4x4 model matrix",
            "type": "ARRAY",
            "componentType": "FLOAT32",
            "componentCount": 16
          }
        }
      }
    }
  }
}
```


**After:**

#### Schema → Class → Car

| name | description | type | componentType | componentCount |
|---|---|---|---|---|
|  | "A forward direction vector" | 'ARRAY' | 'FLOAT64' | 3 |
| "Passenger Names" | "Variable number of passengers, because componentCount is undefined." | 'ARRAY' | 'STRING' | |
|  | "A 4x4 model matrix" | 'ARRAY' | 'FLOAT32' | 16 |

For me this is much easier to read, I'm hopping back and forth between the three specifications quite a bit as it is, and there are some discrepancies in the JSON schema shown now.

specification/Metadata/README.md
```diff
@@ -0,0 +1,699 @@
...
+A schema describes what properties are available. However, it does not describe how property values are stored. This allows the schema to be defined separately from the data itself. This has several benefits:
+
+* An application can know what metadata is present before requesting content.
+* A single schema can be shared across multiple assets. This is useful when one has many assets with the same types of metadata.
+* It allows greater flexibility for storing metadata. A property such as "elevation" may be stored per-vertex or per-texel depending on the use case, while conforming to the same property definition.
+
+Property values are formatted in specific ways depending on the use case. Two formats are defined for storing large collections of property values:
+
+* **Table format** - property values are stored in parallel 1D arrays
+* **Raster format** - property values are stored in channels of a 2D grid of pixels
```
donmccurdy, line 84: I think someone made a comment about the raster format in our call that I've forgotten now, so just to check that I have the right idea — is it still the current/latest intention that this document defines three encodings: JSON, Binary, and Raster?

It would be at least an option to defer definition of encodings to the `3DTILES_metadata` and `EXT_feature_metadata` documents, just mentioning here that different encodings are possible... This has some advantages: first, it avoids some ambiguous cases like the texture coordinate system in the raster format. Second, it leaves more room for additional encodings or compression mechanisms later on.

Happy to defer to you if this has already been decided though!

donmccurdy (commented):
Reading about metadata in EXT_feature_metadata led me here pretty quickly, so a few questions/comments on the intended scope of this.

**tl;dr** – I agree with the chosen "separation of concerns" among these three metadata specifications, but would suggest shuffling some sections between them to make that separation more crisp.

specification/Metadata/README.md
donmccurdy: Additional use case: this specification defines the raster format as a 2D grid of pixels, but you can easily imagine doing the same thing with a 3D volume – for example texture arrays or video, entities having 3D coordinates associated. Certainly not something that should be "in scope" at this stage, but which would benefit from some flexibility on the encoding later on.

specification/Metadata/README.md
```diff
@@ -0,0 +1,699 @@
...
+
+> **Implementation Note:** Depending on the implementation and the chosen integer type, there may be some loss of precision in the normalized values. For example, if the implementation uses 32-bit floating point variables to represent the normalized value, there are only 23 bits in the mantissa. In this case, if the value to normalize is 32- or 64-bit, a number of lower bits will be truncated when normalized. Therefore, it is recommended that implementations use the highest precision floats available for representing the result.
+
+### Minimum and Maximum Values
+
+Numeric properties may specify a minimum and maximum allowed value. For example a property storing GPS coordinates would define a range of `[-180, 180]` degrees for longitude values and `[-90, 90]` degrees for latitude values.
+
+### Optional Properties and Default Values
+
+Properties can be marked as optional. Optional properties do not need to have associated property values.
```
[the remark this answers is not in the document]
donmccurdy: Oops, maybe I misunderstood here – this is saying that for optional properties, the entire "column" may be omitted from the feature table? But that when the "column" is present, all entities must include valid property values?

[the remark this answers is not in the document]
lilleyse: Yes, the whole "column" would be omitted. That could definitely be made clearer.

specification/Metadata/README.md
lilleyse: :+1: for the structured texture version

specification/Metadata/README.md
lilleyse: I like the idea of moving the encoding definitions out of this spec and the intent is to allow for more encodings than described here.

The main concern is that `3DTILES_metadata` and `EXT_feature_metadata` use the same binary tabular encoding and we wanted to avoid defining it in two places.

@pjcozzi had similar feedback about the raster format encoding (although github has collapsed all his comments so it's hard to find them). I think it should be removed from this document.

donmccurdy: Ok, let's see how things go with the bufferview vs. accessor discussion goes? If the glTF-stored binary data does move into accessors (e.g. for better compression behavior), that may be enough of a "break" between the two that we don't want to share a common definition here.

specification/Metadata/README.md
donmccurdy: Will open a PR!

ptrgags: >If I understand correctly, the purpose of this document is to define the primitive data types and relationships that will be used in 3DTILES_metadata and EXT_feature_metadata, in such a way that they can be (1) referenced in both, and (2) trivially portable and compatible between the two. Is that correct?

Correct, the purpose of this document is to factor out the common data types and encodings used in both extensions so we're not repeating ourselves too much. As you say, portability and compatibility are goals here.

Note that a specific extension that references this document might restrict attention given the limitations of the format. For example, `3DTILES_metadata` focuses more on the JSON encodings, while `EXT_feature_metadata` focuses on the binary encodings.

>this specification defines the raster format as a 2D grid of pixels, but you can easily imagine doing the same thing with a 3D volume

Huh that's a great point. 3D volumes certainly have their applications (voxels for example), haven't thought much about video or other time-dynamic data. And you could go further, think n-dimensional NumPy arrays of data for example.

For this common document that's focused on concepts, I agree that we should try to be flexible here. Let the specs that reference this document handle restricting the number of dimensions.

lilleyse:

PR #456, opened 2021-02-28
Metadata Semantic Reference Deep Dive

pjcozzi:
@ptrgags @lilleyse this pull request is for comments only, not intended to be merged.

The new branch `metadata-semantic-reference` can also be deleted once this pull request is closed.  I had to create the branch because there is already a pull request open for the `3d-tiles-next` branch that targets `master`. #455


specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
+# Cesium Metadata Semantic Reference
+
+This document defines a general set of semantics for 3D Tiles and glTF. Tileset authors may define their own application- or domain-specific semantics separately.
```
pjcozzi, line 3: There's not enough context here.

Is this a spec?

Or a non-normative reference?

And "general set" just is not enough context to under the **why** and the **what** for this spec.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
+# Cesium Metadata Semantic Reference
```
pjcozzi, line 1: Is this the right name?  Should it be "Cesium?"

specification/Metadata/Semantics/README.md
pjcozzi: And diagram please showing how the components of the ecosystem come together.

Even placeholder Miro board screenshot would be fine then a graphics artist could improve things later.

pjcozzi: And these semantics are truly just for 3D Tiles and glTF?  Not anything else?  CZML in the future, etc.?

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+This document defines a general set of semantics for 3D Tiles and glTF. Tileset authors may define their own application- or domain-specific semantics separately.
+
+Semantics describe how properties should be interpreted. For example, an application that sees the `HORIZON_OCCLUSION_POINT` semantic would use the property for horizon occlusion culling as defined below.
```
pjcozzi, line 5: A good example, but how many readers will know about horizon culling.  Many readers will be from Mod Sim.  Why not use an example that speaks to them and to developers in general?

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+    },
+    "name": {
+      "type": "STRING",
+      "semantic": "NAME"
+    }
+  }
+}
+```
+
+For full usage see:
```
pjcozzi, line 24: I, the novice reader, don't yet understand why this spec (is it a spec?) is sending me to another spec "for full usage" - nor do I yet understand what usage is defined here or there.

I imagine this spec basically defines the vocab and maybe the runtime/application behavior.

Then these "full usage" specs define the encoding, how it attaches to concepts in those specs, and maybe the runtime/application behavior or at least the interaction of the behavior defined here in the context of the other spec.

The reader needs a roadmap to know what to expect where.

And a diagram.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+* [`3DTILES_metadata`](../../../extensions/3DTILES_metadata/1.0.0) - 3D Tiles extension that assigns metadata to various components of 3D Tiles
+* [`EXT_feature_metadata`](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0) - glTF extension that assigns metadata to features in a model on a per-vertex, per-texel, or per-instance basis
+
+<!-- omit in toc -->
+### **HORIZON_OCCLUSION_POINT**
```
pjcozzi, line 30: Add a table of contents; this is a reference, right?

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The horizon occlusion point expressed in an ellipsoid-scaled fixed frame. If this point is below the horizon, the entire entity is below the horizon. See [Horizon Culling](https://cesium.com/blog/2013/04/25/horizon-culling/) for more information. This semantic is often used with tile metadata.
```
pjcozzi, line 32: >This semantic is often used with tile metadata.

Does the reader know what this is?

And should there be a link?

specification/Metadata/Semantics/README.md
pjcozzi: I think this document is basically a dictionary then other specs can reference the specific vocab here instead of copy and pasting it.  Please state this explicitly and more precisely than I just did. 😄

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+* Type: `ARRAY`
+* Component type: `FLOAT32` or `FLOAT64`
+* Component count: `3`
+
+<!-- omit in toc -->
+### **MINIMUM_HEIGHT**
+
+The minimum height relative to some ellipsoid.
```
pjcozzi, line 41: Here and below, what is "some ellipsoid" - spec language can't be this imprecise.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+### **MAXIMUM_HEIGHT**
+
+The maximum height relative to some ellipsoid.
+
+* Type: `FLOAT32` or `FLOAT64`
+
+<!-- omit in toc -->
+### **BOUNDING_SPHERE**
+
+A bounding sphere as `[x, y, z, radius]`.
```
pjcozzi, line 55: This should be defined precisely and completely, not just "by example".

This is a bounding sphere (or just a sphere in this more generic context?)

specification/Metadata/Semantics/README.md
pjcozzi: Ah, I see that this doesn't really define any behavior but kinda tries to some time.

The behavior line needs to be crystal clear.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+* Type: `ARRAY`
+* Component type: `FLOAT32` or `FLOAT64`
+* Component count: `4`
+
+<!-- omit in toc -->
+### **NAME**
```
pjcozzi, line 62: Is there enough context around name vs. index?

And should this spec define that?

Here's how glTf motivates it: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#indices-and-names

pjcozzi (commented):
@ptrgags @lilleyse just a few comments from me on this one.  I'm sure this will grow a lot over time.  Two themes:

* **Intro** - needs to clearly and precisely state what this is and isn’t
* **Who defines runtime behavior? ** This spec or the spec referencing this?

pjcozzi:
@ptrgags @lilleyse also now wondering if this should be an appendix in the metadata spec, #455, we need to think through what is normative and what isn't - and what becomes normative once referenced elsewhere.

lilleyse:
See notes in https://github.com/CesiumGS/3d-tiles/pull/464. 

> Only `FLOAT64` should be allowed for regions.
>
>Still determining whether we want to keep `FLOAT32` for bounding boxes, bounding spheres, and min/max heights. For now it's ok but `FLOAT32` should have a caveat attached to it since it's not the most precise way to represent bounding volumes at global scale. That could be a TODO for later once we go through https://github.com/CesiumGS/3d-tiles/pull/456.

specification/Metadata/Semantics/README.md
lilleyse: This should be fixed in https://github.com/CesiumGS/3d-tiles/pull/464

specification/Metadata/Semantics/README.md
lilleyse: This should be fixed in https://github.com/CesiumGS/3d-tiles/pull/464 when `MINIMUM_HEIGHT` and `MAXIMUM_HEIGHT` are renamed to `TILE_MINIMUM_HEIGHT` and `TILE_MAXIMUM_HEIGHT` and defined according to `region`.

CC @sanjeetsuhag

IanLilleyT:
`FLOAT32` requires some extra carefulness on the tiler end to not shrink the bounding volume compared to the float64 version, but it can be done. And it has the benefit of being half the file size.

lilleyse:

PR #457, opened 2021-03-01
3DTILES_metadata Extension deep dive

pjcozzi:
@ptrgags @lilleyse this pull request is for deep dive comments only; it is not intended to be merge.

CC #455, #456


extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
+<!-- omit in toc -->
+# 3DTILES_metadata Extension
+
+**Version 1.0.0**, February 25, 2021
```
pjcozzi, line 4: Same comment on versioning as #455

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+<!-- omit in toc -->
+## Contributors
+
+* Sean Lilley, Cesium
+* Peter Gagliardi, Cesium
+* Sam Suhag, Cesium
+* Patrick Cozzi, Cesium
```
pjcozzi, line 12: Please put me last in any new extensions that I contributed to in a material way.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+## Status
+
+Draft
+
+<!-- omit in toc -->
+## Dependencies
+
+Written against the 3D Tiles 1.0 specification.
+
+Adds new functionality to the [`3DTILES_implicit_tiling` extension](../../3DTILES_implicit_tiling/0.0.0). See [Implicit Tile Metadata](#implicit-tile-metadata).
```
pjcozzi, line 25: When the novice reads top down, it seems like this extension may _only_ add new functionality to `3DTILES_implicit_tiling` since that is the first thing they read.  One more sentence may be helpful.

Let's also discuss/confirm how extensions can extend/require other extensions and make sure 3D Tiles is aligned with the Khronos and glTF approach.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+  - [Group Metadata](#group-metadata)
+  - [Feature Metadata](#feature-metadata)
+  - [Statistics](#statistics)
+  - [Semantics](#semantics)
+  - [Styling](#styling)
+- [JSON Schema Reference](#json-schema-reference)
+
+## Overview
+
+This extension defines a standard mechanism for adding metadata to 3D Tiles. This includes:
```
pjcozzi, line 52: Similar to the intro in #455, this should define a tad more what metadata is.

Readers will come into this with lots of ideas about what metadata means to them, e.g.,
* https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Khronos/KHR_xmp/README.md#overview
* https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/AGI_stk_metadata/README.md
* https://github.com/KhronosGroup/glTF/blob/master/extensions/2.0/Vendor/FB_geometry_metadata/README.md

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+* Tileset metadata - metadata about the tileset as a whole.
+* Tile metadata - metadata about individual tiles.
+* Group metadata - metadata about groups of content.
```
pjcozzi, line 56: A group can be content across multiple tiles, right?  I get the new to keep first-touch definitions concise but this is so concise that the novice reader won't get it.

Perhaps something like "meta about groups of content stored in one or more tiles."

Let's also discuss if group is the right term for this concept, could be, but I want to double check the intentionality.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+* Feature metadata - metadata about features. See the companion glTF extension [EXT_feature_metadata](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0).
+
+<img src="figures/metadata-granularity.png"  alt="Metadata Granularity" width="600">
```
pjcozzi, line 59: Nice diagram.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+A tileset defines a **schema**. A schema has a set of **classes** and **enums**. A class contains a set of **properties**, which may be numeric, boolean, string, enum, or array types.
```
pjcozzi, line 61: > A tileset defines a **schema**.

A tileset defines a lot more than that.

I think this means that a tileset may contain a schema to define metadata definitions.  Metadata definitions can be composed of classes and enums.

Or something along those lines.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+**Entities** (such as tiles, features, etc.) conform to classes and contain **property values**. Depending on the context, property values may be stored in JSON or binary.
```
pjcozzi, line 63: > Depending on the context

What value does this phrase add to the reader?

Probably remove it.

[Make every word tell](https://www.bartleby.com/141/)

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+**Statistics** provide aggregate information about the metadata. For example, statistics may include the min/max values of a numeric property for mapping property values to color ramps with the [declarative styling language](../../../specification/Styling) or the number of enum occurrences for creating histograms.
+
+By default, properties do not have any inherent meaning. A property may be assigned a **semantic**, an identifier that describes how the property should be interpreted. The full list of built-in semantics can be found in the [Cesium Metadata Semantic Reference](../../../specification/Metadata/Semantics). Tileset authors may define their own application- or domain-specific semantics separately.
+
+This extension implements the [Cesium 3D Metadata Specification](../../../specification/Metadata/1.0.0), which describes the metadata format in full detail.
+
+## Use Cases
```
pjcozzi, line 71: A spec should answer why, before what and how.

Can we put more of this motivation up front in the intro?  Then the reader knows why this spec defines entities, statistics, etc. when they read the current intro.

[Start with Why](https://www.ted.com/talks/simon_sinek_how_great_leaders_inspire_action?language=en)

extensions/3DTILES_metadata/1.0.0/README.md
pjcozzi: CC #455 to confirm vocab for "entity", etc.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+**Entities** (such as tiles, features, etc.) conform to classes and contain **property values**. Depending on the context, property values may be stored in JSON or binary.
+
+**Statistics** provide aggregate information about the metadata. For example, statistics may include the min/max values of a numeric property for mapping property values to color ramps with the [declarative styling language](../../../specification/Styling) or the number of enum occurrences for creating histograms.
+
+By default, properties do not have any inherent meaning. A property may be assigned a **semantic**, an identifier that describes how the property should be interpreted. The full list of built-in semantics can be found in the [Cesium Metadata Semantic Reference](../../../specification/Metadata/Semantics). Tileset authors may define their own application- or domain-specific semantics separately.
+
+This extension implements the [Cesium 3D Metadata Specification](../../../specification/Metadata/1.0.0), which describes the metadata format in full detail.
```
pjcozzi, line 69: Is it possible to "implement" the metadata spec?  Or just to reference sections of if, which then become normative in this spec?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+<img src="figures/metadata-granularity.png"  alt="Metadata Granularity" width="600">
+
+A tileset defines a **schema**. A schema has a set of **classes** and **enums**. A class contains a set of **properties**, which may be numeric, boolean, string, enum, or array types.
+
+**Entities** (such as tiles, features, etc.) conform to classes and contain **property values**. Depending on the context, property values may be stored in JSON or binary.
+
+**Statistics** provide aggregate information about the metadata. For example, statistics may include the min/max values of a numeric property for mapping property values to color ramps with the [declarative styling language](../../../specification/Styling) or the number of enum occurrences for creating histograms.
+
+By default, properties do not have any inherent meaning. A property may be assigned a **semantic**, an identifier that describes how the property should be interpreted. The full list of built-in semantics can be found in the [Cesium Metadata Semantic Reference](../../../specification/Metadata/Semantics). Tileset authors may define their own application- or domain-specific semantics separately.
```
pjcozzi, line 67: To implement `3DTILES_metadata`, do you need to support the semantics in the semantic reference?  It sounds that way but is that the intent?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+This extension implements the [Cesium 3D Metadata Specification](../../../specification/Metadata/1.0.0), which describes the metadata format in full detail.
+
+## Use Cases
+
+_This section is non-normative_
+
+This extension is designed with several new use cases in mind.
```
pjcozzi, line 75: These are all good but this extension is such a huge leap forward in the ability to efficiently store and transmit metadata at a variety of granularity levels (and then efficiently interact with it at runtime) that I think the Intro and Use Cases section under sell this a bit and do not put enough emphasis on this.  There should be a ton of material that I wrote in GitHub issues or slides dating back up to a few years or more recent work that @lilleyse did that properly motivates this.

Just looking for a few sentences and a diagram.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+* An application can style tilesets using metadata stored at per-tileset, per-tile, per-group, and per-feature granularity
+* An application can see what metadata is available before requesting content, e.g. for populating a UI
+* Tileset authors can extend metadata with their own semantic data model
+* Metadata schemas can be shared within a tileset and across different tilesets
+* A runtime engine can optimize traversal algorithms by using per-tile metadata
+* An application can create a layering system by adding show/hide or color styles for groups, as well as selectively loading groups on demand
+
+## Compatibility Notes
```
pjcozzi, line 84: Seems like this should be combined with the **Dependencies** section.

> Instead use glTF content directly with [`3DTILES_content_gltf`](../../3DTILES_content_gltf/0.0.0).

The reader might think that b3dm and i3dm should use 3DTILES_content_gltf; instead, I think this means that you need to use 3DTILES_content_gltf to use this extension for content, is that right?

Please make things super explicit.

pjcozzi (commented):
@ptrgags @lilleyse I only read down to just before **Concepts** so far.  I'll follow up with the rest and some top level themes tomorrow if I am lucky. 😄

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+    }
+  }
+}
+```
+
+### Group Metadata
+
+Metadata may be assigned to groups. Groups represent collections of contents. Contents are assigned to groups with the `3DTILES_metadata` content extension.
+
+Group metadata can be paired with [`3DTILES_multiple_contents`](../../3DTILES_multiple_contents/0.0.0) to assign multiple contents in a tile to different groups.
```
pjcozzi, line 325: Not sure that the novice reader is ready for this detail yet, but it would be welcomed in an Appendix on how other extensions interact with metadata.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+        }
+      }
+    ]
+  }
+}
+```
+
+### Feature Metadata
+
+A feature is an entity that has both geometry and metadata, such as a 3D building in a city. Features are the finest level of granularity within the tileset.
```
pjcozzi, line 407: Does this redefine **feature** compared to the 3D Tiles core spec?  There my intent was that a feature did not need to have metadata; in practice, it might have had metadata to allow for batching within a tile.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+Metadata may be assigned to features using the glTF extension [`EXT_feature_metadata`](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0).
+
+Feature metadata classes may be included in the `3DTILES_metadata` schema. This lets an application know what classes exist before requesting content.
```
pjcozzi, line 411: Example please.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+* A runtime engine can optimize traversal algorithms by using per-tile metadata
+* An application can create a layering system by adding show/hide or color styles for groups, as well as selectively loading groups on demand
+
+## Compatibility Notes
+
+This extension is independent of the [Batch Table](../../../specification/TileFormats/BatchTable) used in the Batched 3D Model, Instanced 3D Model, and Point Cloud formats. Similarly, this extension is independent of the [`properties`](../../../specification/schema/properties.schema.json) object in tileset JSON.
+
+glTF models in Batched 3D Model or Instanced 3D Model content must not use the [EXT_feature_metadata](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0) extension. Instead use glTF content directly with [`3DTILES_content_gltf`](../../3DTILES_content_gltf/0.0.0).
+
+## Concepts
```
pjcozzi, line 90: The concepts section introduces things in the right order: schemas, followed coarse-to-fine-grained: tileset, tile (JSON and buffer backed), groups, then features.

This is a lot for the novice reader.

A diagram would be helpful.

That diagram or a separate diagram should also show the interaction with other extensions as needed, at least glTF's EXT_feature_metadata.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+A feature is an entity that has both geometry and metadata, such as a 3D building in a city. Features are the finest level of granularity within the tileset.
+
+Metadata may be assigned to features using the glTF extension [`EXT_feature_metadata`](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0).
+
+Feature metadata classes may be included in the `3DTILES_metadata` schema. This lets an application know what classes exist before requesting content.
+
+### Statistics
+
+Statistics provide aggregate information about entities in a tileset. Statistics are provided on a per-class basis.
```
pjcozzi, line 415: > Statistics provide aggregate information about entities in a tileset

Reading just this far, I don't yet know if this is only at the tileset granularity or at any subtree; we'll want the flexibility for the later unless there is a good reason to avoid the complexity.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+* `count` is the number of entities that conform to the class
```
pjcozzi, line 417: I'm still not sure I understand the vocab decision for Entity as an instance of a Class, but is "conform" the right word for instance here?  I'm not a Programming Languages researcher, but I'm pretty sure this would read something like "count is the number of **objects** that are **instances** of the **class**" or something like that using generic programming language type system terms

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+* `properties` contains statistics about property values
+
+Properties have the following built-in statistics:
+
+Name|Description|Type
+--|--|--
+`min`|The minimum property value|Numeric types or fixed-length arrays of numeric types
```
pjcozzi, line 424: Are we sure we want to use abbreviations?

3D Tiles and Cesium in general doesn't: https://github.com/CesiumGS/3d-tiles/blob/master/specification/schema/properties.schema.json

However, I think glTF does for `min` and `max`.

But seems like 3D Tiles should stay consistent with 3D Tiles.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+### Statistics
+
+Statistics provide aggregate information about entities in a tileset. Statistics are provided on a per-class basis.
+
+* `count` is the number of entities that conform to the class
+* `properties` contains statistics about property values
+
+Properties have the following built-in statistics:
+
+Name|Description|Type
```
pjcozzi, line 422: Do we have use cases for each row in this table?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+`mean`|The arithmetic mean of the property values|...
+`median`|The median of the property values|...
+`standardDeviation`|The standard deviation of the property values|...
+`variance`|The variance of the property values|...
+`sum`|The sum of the property values|...
+`occurrences`|Number of enum occurrences|Enums or fixed-length arrays of enums
+
+Tileset authors may define their own additional semantics, like `mode` in the example below.
+
+```jsonc
```
pjcozzi, line 435: This is an extremely long example to show one line of JSON:

`"mode": 5.0`

In general, don't force the reader to sort through unimportant detail to get to the essence of your message.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+      },
+      "statistics": {
+        "classes": {
+          "building": {
+            "count": 100000,
+            "properties": {
+              "height": {
+                "min": 3.9,
+                "max": 341.7,
+                "mode": 5.0
```
pjcozzi, line 489: Are we sure that we want to encourage/allow users to add properties like this?  It is not forward compatibility, e.g., 3D Tiles could introduce `mode` in the future, which could have a different meaning.  This is why 3D Tiles and glTF do not allow adding properties like this and instead require that they be added via `extras`.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+### Semantics
+
+By default properties do not have any inherent meaning. A property may be assigned a **semantic**, an identifier that describes how the property should be interpreted. The full list of built-in semantics can be found in the [Cesium Metadata Semantic Reference](../../../specification/Metadata/Semantics).
```
pjcozzi, line 509: Semantics should have a consistent definition here, #455, and #456 - and we should define more than "interpreted", e.g., is it runtime behavior or just more rich type information?  I think it will vary based on the semantic itself and that specific runtime behavior should be limited.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+<!-- omit in toc -->
+## Tileset extension
+
+3D Tiles extension that defines metadata for a tileset.
+
+**`Tileset extension` Properties**
+
+|   |Type|Description|Required|
+|---|---|---|---|
+|**schema**|`schema`|An object defining classes and enums.|No|
```
pjcozzi, line 582: Can both `schema` and `schemaUri` be defined?  Maybe not because then duplicates would have to be resolved?  But one of the two must be defined?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**schemaUri**|`string`|A uri to an external schema file.|No|
```
pjcozzi, line 583: Where is the JSON schema for the external root file?  And a complete example?  Perhaps in the appendix?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**statistics**|`statistics`|An object containing statistics about entities.|No|
+|**groups**|`object`|A dictionary, where each key is a group ID and each value is an object defining the group.|No|
```
pjcozzi, line 585: For symmetry of other properties in this table and precision in using JSON (which I believe doesn't formally define "dictionary"), this description should be written as "An object that defines a dictionary where each key..."

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**tileset**|`tileset`|An object containing metadata about the tileset.|No|
+|**extensions**|`any`||No|
+|**extras**|`any`||No|
+
+Additional properties are allowed.
```
pjcozzi, line 590: > Additional properties are allowed

Is this intentional?  See the forward compatibility comment above.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+An object defining classes and enums.
+
+**`Schema ` Properties**
+
+|   |Type|Description|Required|
+|---|---|---|---|
+|**name**|`string`|The name of the schema.|No|
+|**description**|`string`|The description of the schema.|No|
+|**version**|`string`|Application-specific version of the schema.|No|
```
pjcozzi, line 659: Let's talk through the use cases for schema versioning.

extensions/3DTILES_metadata/1.0.0/README.md
pjcozzi: Same comment throughout.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+Additional properties are allowed.
+
+<!-- omit in toc -->
+#### schema.name
+
+The name of the schema.
+
+* **Type**: `string`
+* **Required**: No
+* **Minimum Length**`: >= 1`
```
pjcozzi, line 674: Minor, but why do the strings throughout have a minimum length of 1?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+**`Class property` Properties**
+
+|   |Type|Description|Required|
+|---|---|---|---|
+|**name**|`string`|The name of the property, e.g. for display purposes.|No|
+|**description**|`string`|The description of the property.|No|
+|**type**|`string`|The property type. If `ENUM` is used, then `enumType` must also be specified. If `ARRAY` is used, then `componentType` must also be specified. `ARRAY` is a fixed-length array when `componentCount` is defined, and variable-length otherwise.| &#10003; Yes|
+|**enumType**|`string`|An enum ID as declared in the `enums` dictionary. This value must be specified when `type` or `componentType` is `ENUM`.|No|
+|**componentType**|`any`|When `type` is `ARRAY` this indicates the type of each component of the array. If `ENUM` is used, then `enumType` must also be specified.|No|
+|**componentCount**|`integer`|The number of components per element for `ARRAY` elements.|No|
```
pjcozzi, line 798: I'm still confused on component vs. element vocab (CC #455 or #456), and now I am more confused. 😄 

An array element can have more than one component just like a vertex attribute can?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**normalized**|`boolean`|Specifies whether integer values are normalized. This applies both when `type` is an integer type, or when `type` is `ARRAY` with a `componentType` that is an integer type. For unsigned integer types, values are normalized between `[0.0, 1.0]`. For signed integer types, values are normalized between [-1.0, 1.0]. For all other types, this property is ignored.|No, default: `false`|
```
pjcozzi, line 799: > [-1.0, 1.0]

Put in code: `[-1.0, 1.0]`

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**min**|`number,array`|Minimum allowed values for property values. Only applicable for numeric types and fixed-length arrays of numeric types. For numeric types this is a single number. For fixed-length arrays this is an array with `componentCount` number of elements. `normalized` property has no effect on these values: they always correspond to the integer values.|No|
+|**default**|`boolean,number,string,array`|A default value to use when the property value is not defined. If used, `optional` must be set to true. The type of the default value must match the property definition: For `BOOLEAN` use `true` or `false`. For `STRING` use a JSON string. For numeric types use a JSON number. For `ENUM` use the enum `name`, not the integer value. For `ARRAY` use a JSON array containing values matching the `componentType`.|No|
+|**optional**|`boolean`|If true, this property is optional.|No, default: `false`|
```
pjcozzi, line 803: If `optional` is true, is `default` required?  Or is the property value `undefined`?  OK to follow whatever glTF & 3D Tiles do, just confirming.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+|**semantic**|`string`|An identifier that describes how this property should be interpreted. The semantic cannot be used by other properties in the class.|No|
```
pjcozzi, line 804: > The semantic cannot be used by other properties in the class

Case sensitive?  It should be since it is JSON.  Not sure that it needs to be explicit.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+<!-- omit in toc -->
+#### class.property.componentCount
+
+The number of components per element for `ARRAY` elements.
+
+* **Type**: `integer`
+* **Required**: No
+* **Minimum**: ` >= 2`
+
+<!-- omit in toc -->
```
pjcozzi, line 890: Some readers will read the spec top down / browse for learning.

Other readers will come directly to the reference for a specific piece of information.

The reference is sometimes meaty and other times extremely light.

It would be nice for the reference to link to the corresponding parts of the **Concepts** section for more context but I know that is a bit painful at the moment with the wetzel workflow - so maybe write an issue to do this in the future?

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+* **Type**: `any`
+* **Required**: No
+
+---------------------------------------
+<a name="reference-enum"></a>
+<!-- omit in toc -->
+### Enum
+
+An object defining the values of an enum.
```
pjcozzi, line 956: Not just the values, but the enum itself, which includes its values.  This creates the type.

pjcozzi (commented):
@ptrgags @lilleyse I made it down to the Tileset Metadata reference doc section so far. I will follow up with more specific comments as well as the extracted general themes.  Most comments so far are typical/minor around vocab, precision in language, consistency, easing in the novice reader, etc.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+* **Type**: `any`
+* **Required**: No
+
+
+
+
+---------------------------------------
+<a name="reference-subtree-property"></a>
+<!-- omit in toc -->
+### Subtree property
```
pjcozzi, line 1556: I'm a bit confused; when I read the **Concepts** section, I assumed that properties could be backed by inline JSON or buffers - now it looks like buffers are just for subtrees.  Whatever the design is is probably fine, but the delivery can be more explicit.

extensions/3DTILES_metadata/1.0.0/README.md
```diff
@@ -0,0 +1,1623 @@
...
+
+An array of binary property values.
+
+**`Subtree property` Properties**
+
+|   |Type|Description|Required|
+|---|---|---|---|
+|**bufferView**|`integer`|The index of the buffer view containing property values. The data type of property values is determined by the property definition: When `type` is `BOOLEAN` values are packed into a bitfield. When `type` is `STRING` values are stored as byte sequences and decoded as UTF-8 strings. When `type` is a numeric type values are stored as the provided `type`. When `type` is `ENUM` values are stored as the enum's `valueType`. When `type` is `ARRAY` elements are packed tightly together and the data type is based on the `componentType` following the same rules as above. `arrayOffsetBufferView` is required for variable-size arrays and `stringOffsetBufferView` is required for strings (for variable-length arrays of strings, both are required). The buffer view `byteOffset` must be aligned to a multiple of 8 bytes.| &#10003; Yes|
```
pjcozzi, line 1564: Somewhere a diagram for these would be valuable even if it needs to be in more than one spec document so the reader isn't expected to have everything memorized or know when to cross reference.

pjcozzi (commented):
@ptrgags @lilleyse this is my third and final set of comments (just 2!) for `3DTILES_metadata`.

Here's the general themes / a few details for discussion:
* All these Metadata specs really define a Type System, what should it be called (CC #455 where I said Abstract Language, which may or may not be right)
* Interaction with Cesium 3D Metadata Specification and Cesium Metadata Semantic Reference
* Extensions extending other extensions.  Confirm 3D Tiles aligns with Khronos and glTF.
* Vocab
   * Group vs. layer
   * Metadata vs. Entity.  Metadata vs/contains schema
* Introducing concepts (properties, and properties backed by buffers) vs. how other extensions use these (implicit tiling, multiple contents)
* "A feature is an entity that has both geometry and metadata"
* Use cases for all the statistics properties: min, max, mean, median, etc.?
* `schema.version` - needed?
* `groups` / `properties` / `tileset` at the same level as `schema`?
* Should there be MAT4, Color, etc. types? CC #455

lilleyse:

PR #463, opened 2021-05-04
Update README.md

tanghaojie:
add globalpropertyboolean reference


javagl:
Resolving the merge conflicts was not trivially possible (and the instructions that GitHub gives do not work _at all_, at least, not for someone who is not a Git Guru). So this is now updated via https://github.com/CesiumGS/3d-tiles/pull/597

PR #464, opened 2021-05-05
Adds semantics for tile and content bounding volumes

sanjeetsuhag:
The following semantics allow `3DTILES_metadata` to specify information about the bounding volume of a tile or a tile's content:

- `TILE_BOUNDING_BOX`
- `TILE_BOUNDING_REGION`
- `TILE_BOUNDING_SPHERE`
- `CONTENT_BOUNDING_BOX`
- `CONTENT_BOUNDING_REGION`
- `CONTENT_BOUNDING_SPHERE`

### DIFF
--- a/specification/Metadata/Semantics/README.md
+++ b/specification/Metadata/Semantics/README.md
@@ -56,7 +56,7 @@ The bounding volume of the tile, expressed as a [sphere (as defined by 3D Tiles 
 <!-- omit in toc -->
 ### **CONTENT_BOUNDING_BOX**
 
-The bounding volume of the content of a tile, expressed as a [box (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#box). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). The content bounding volume must be completely contained within the bounding volume of a tile.
+The bounding volume of the content of a tile, expressed as a [box (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#box). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). This property is equivalent to `tile.content.boundingVolume.box`.
 
 * Type: `ARRAY`
 * Component type: `FLOAT32` or `FLOAT64`
@@ -65,7 +65,7 @@ The bounding volume of the content of a tile, expressed as a [box (as defined by
 <!-- omit in toc -->
 ### **CONTENT_BOUNDING_REGION**
 
-The bounding volume of the content of a tile, expressed as a [region (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#region). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). The content bounding volume must be completely contained within the bounding volume of a tile.
+The bounding volume of the content of a tile, expressed as a [region (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#region). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). This property is equivalent to `tile.content.boundingVolume.region`.
 
 * Type: `ARRAY`
 * Component type: `FLOAT64`
@@ -74,7 +74,7 @@ The bounding volume of the content of a tile, expressed as a [region (as defined
 <!-- omit in toc -->
 ### **CONTENT_BOUNDING_SPHERE**
 
-The bounding volume of the content of  tile, expressed as a [sphere (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#sphere). The content bounding volume must be completely contained within the bounding volume of a tile.
+The bounding volume of the content of  tile, expressed as a [sphere (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#sphere). This property is equivalent to `tile.content.boundingVolume.sphere`.
 
 * Type: `ARRAY`
 * Component type: `FLOAT32` or `FLOAT64`
--- a/specification/Metadata/Semantics/README.md
+++ b/specification/Metadata/Semantics/README.md
@@ -27,6 +27,42 @@ For full usage see:
 * [`EXT_feature_metadata`](https://github.com/CesiumGS/glTF/tree/master/extensions/2.0/Vendor/EXT_feature_metadata/1.0.0) - glTF extension that assigns metadata to features in a model on a per-vertex, per-texel, or per-instance basis
 
 <!-- omit in toc -->
+### **TILE_BOUNDING_BOX**
+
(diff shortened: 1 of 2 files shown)


specification/Metadata/Semantics/README.md
```diff
@@ -26,6 +26,60 @@ For full usage see:
...
+* Component type: `FLOAT32` or `FLOAT64`
+* Component count: `12`
+
+<!-- omit in toc -->
+### **TILE_BOUNDING_REGION**
+
+The bounding volume of the tile, expressed as a [region (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#region). This property may be used to describe a tighter bounding volume for a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md).
+
+* Type: `ARRAY`
+* Component type: `FLOAT32` or `FLOAT64`
```
lilleyse, line 44: Only `FLOAT64` should be allowed for regions.

Still determining whether we want to keep `FLOAT32` for bounding boxes, bounding spheres, and min/max heights. For now it's ok but `FLOAT32` should have a caveat attached to it since it's not the most precise way to represent bounding volumes at global scale. That could be a TODO for later once we go through https://github.com/CesiumGS/3d-tiles/pull/456.

specification/Metadata/Semantics/README.md
```diff
@@ -26,6 +26,60 @@ For full usage see:
...
+### **CONTENT_BOUNDING_SPHERE**
+
+The bounding volume of the content of  tile, expressed as a [sphere (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#sphere). The content bounding volume must be completely contained within the bounding volume of a tile.
+
+* Type: `ARRAY`
+* Component type: `FLOAT32` or `FLOAT64`
+* Component count: `4`
+
 <!-- omit in toc -->
 ### **HORIZON_OCCLUSION_POINT**
```
lilleyse, line 84: For consistency let's make the following renames:

* `HORIZON_OCCLUSION_POINT` -> `TILE_HORIZON_OCCLUSION_POINT`
* `MINIMUM_HEIGHT` -> `TILE_MINIMUM_HEIGHT`
* `MINIMUM_HEIGHT` -> `TILE_MINIMUM_HEIGHT`

lilleyse, line 84: `BOUNDING_SPHERE` below can be removed. The original idea was that it was generic and could be used by anything that needs a bounding sphere, but I don't think that's useful any more.

lilleyse: Along with the rename `TILE_MINIMUM_HEIGHT` and `TILE_MAXIMUM_HEIGHT` need more precise definitions now that they are specific to tile bounding regions.

specification/Metadata/Semantics/README.md
```diff
@@ -26,6 +26,60 @@ For full usage see:
...
+The bounding volume of the tile, expressed as a [sphere (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#sphere).
+
+* Type: `ARRAY`
+* Component type: `FLOAT32` or `FLOAT64`
+* Component count: `4`
+
+<!-- omit in toc -->
+### **CONTENT_BOUNDING_BOX**
+
+The bounding volume of the content of a tile, expressed as a [box (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#box). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). The content bounding volume must be completely contained within the bounding volume of a tile.
```
lilleyse, line 59: > The content bounding volume must be completely contained within the bounding volume of a tile.

This is not necessarily true - see https://github.com/CesiumGS/3d-tiles/tree/master/specification#bounding-volume-spatial-coherence. I think the sentence can be removed and instead link to a relevant section in the 3D Tiles spec or schema that describes what content bounding volumes are.

specification/Metadata/Semantics/README.md
lilleyse: Make sure to do the rename everywhere in 3d-tiles-next branch. I think we reference `HORIZON_OCCLUSION_POINT` in multiple places.

specification/Metadata/Semantics/README.md
lilleyse: Added a note here https://github.com/CesiumGS/3d-tiles/pull/456#issuecomment-832742634

specification/Metadata/Semantics/README.md
lilleyse: > Make sure to do the rename everywhere in 3d-tiles-next branch. I think we reference `HORIZON_OCCLUSION_POINT` in multiple places.

CC @IanLilleyT @ptrgags for updating tilers / CesiumJS

lilleyse:
What happens if someone defines all three bounding volume types in the metadata?

`TILE_BOUNDING_BOX`
`TILE_BOUNDING_REGION`
`TILE_BOUNDING_SPHERE`

Does the implementation have a precedence order?

sanjeetsuhag:
> What happens if someone defines all three bounding volume types in the metadata?
> 
> `TILE_BOUNDING_BOX`
> `TILE_BOUNDING_REGION`
> `TILE_BOUNDING_SPHERE`
> 
> Does the implementation have a precedence order?

This should be disallowed by the spec, as is done by 3D Tiles 1.0. Given that these are semantics, I think it's reasonable to enforce  a restriction like that.

specification/Metadata/Semantics/README.md
sanjeetsuhag: >  [The bounding volumes defined by this property must maintain spatial coherence](https://github.com/CesiumGS/3d-tiles/tree/master/specification#bounding-volume-spatial-coherence).

Would the line above (or something like it) be needed for all relevant properties? This includes the `HEIGHT` properties too.

lilleyse: Maybe...

> The bounding volume of the content of a tile, expressed as a [box (as defined by 3D Tiles 1.0)](https://github.com/CesiumGS/3d-tiles/tree/master/specification#box). This property may be used to describe a tighter bounding volume for the content of a tile than is implicitly calculated by [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_implicit_tiling/0.0.0/README.md). This property is equivalent to `tile.content.boundingVolume.box`.

sanjeetsuhag:
@lilleyse Your feedback has been addressed.

IanLilleyT:
I'm a little bit confused about when you would use `CONTENT` vs `TILE`.

For something like horizon occlusion point, it could either be computed from the corners of the tile or from the content geometry itself. Is it important to be able to distinguish these?

And what does `TILE_MINIMUM_HEIGHT` imply? If you're setting a minimum height for the tile, wouldn't it typically be derived from the content within? If that's not always the case, would we need `CONTENT_MINIMUM_HEIGHT` too?

sanjeetsuhag:
For the horizon occlusion point, I agree that it makes more sense to do it by `content` than `tile`.

> And what does TILE_MINIMUM_HEIGHT imply? If you're setting a minimum height for the tile, wouldn't it typically be derived from the content within? If that's not always the case, would we need CONTENT_MINIMUM_HEIGHT too?

The primary use case here is when implicit tiling plays well with subdivision over the Earth's surface, but not so much in the vertical dimension. `QUADTREE` subdivision doesn't address height at all and `OCTREE` subdivision might not be correct (in most cases). 

This diagram below shows what would happen when explicit heights are not defined:

quadtrees

So, you are correct that these need to be content driven. @lilleyse Any thoughts on whether tile specific use cases?

> And what do you set when the tile and content have the "same" minimum height?

You'd have to put data for both, but `content` bounding volumes would be used for culling over `tile`.

lilleyse:
> And what does `TILE_MINIMUM_HEIGHT` imply? If you're setting a minimum height for the tile, wouldn't it typically be derived from the content within? If that's not always the case, would we need `CONTENT_MINIMUM_HEIGHT` too?

For terrain the tile bounding volume usually wraps the contents but that's not always the case for additive refinement tilesets:

image

We should probably add `CONTENT_MINIMUM_HEIGHT` and `CONTENT_MAXIMUM_HEIGHT` for completeness, though I think `CONTENT_BOUNDING_BOX/SPHERE/REGION` is more useful since tilesets that use content bounding volumes usually want to wrap the content in all directions, not just Z.

lilleyse:
I added some more implementation notes about what to do if multiple bounding volumes types are specified. Also added `CONTENT_HORIZON_OCCLUSION_POINT` for completeness plus an implementation note and when this would be applicable.

PR #470, opened 2021-06-21
Clarify enum binary properties in 3DTILES_implicit_tiling and 3DTILES_metadata

lilleyse:
This is targeting the `3d-tiles-next` branch.

For implicit tile metadata, clarifies that each enum value in the buffer must match one of the values in the enum definition. This prevents extraneous values from being added to the buffer.

See https://github.com/CesiumGS/glTF/pull/7 for the companion PR for `EXT_feature_metadata`


ptrgags:
Feels like déjà vu. Changes look good, thanks @lilleyse!

PR #473, opened 2021-07-08
Explicitly set type of tileset.properties

krupkad:
`tileset.properties` is implicitly an `object` due to `additionalProperties`, making it explicit


lilleyse:
Thanks @krupkad. One step closer to the 3D Tiles serializer in Cesium Native! https://github.com/CesiumGS/cesium-native/issues/278

PR #476, opened 2021-08-25
Change feature table byteOffset to integer

krupkad:
`byteOffset` should be integral, as this version of JSON Schema allows it (https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-04#section-3.5) and we do use `integer` in `3DTILES_batch_table_hierarchy`

### DIFF
--- a/specification/schema/featureTable.schema.json
+++ b/specification/schema/featureTable.schema.json
@@ -11,7 +11,7 @@
             "description" : "An object defining the reference to a section of the binary body of the features table where the property values are stored if not defined directly in the JSON.",
             "properties" : {
                 "byteOffset" : {
-                    "type" : "number",
+                    "type" : "integer",
                     "description": "The offset into the buffer in bytes.",
                     "minimum" : 0
                 },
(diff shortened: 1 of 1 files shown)


ptrgags (approved):
@krupkad Nice catch, I wasn't aware that "integer" was an available JSON Schema type. Changes look good to me.

@sanjeetsuhag we should take note of this when we get back to 3D Tiles Next specifications.

lilleyse:
I pushed some other fixes I noticed: https://github.com/CesiumGS/3d-tiles/pull/477


There was only one change for 3d-tiles-next so I pushed directly: https://github.com/CesiumGS/3d-tiles/commit/236e1c03fb61090890b2be054dbbbdf2fd417dc8. However there are a few areas in 3d-tiles-next that you could argue should have both "number" and "integer" but I'm not sure if that really made sense.

https://github.com/CesiumGS/3d-tiles/blob/236e1c03fb61090890b2be054dbbbdf2fd417dc8/extensions/3DTILES_metadata/1.0.0/schema/class.property.schema.json#L83-L91

lilleyse:
Pushed one change to `EXT_feature_metadata` as well: https://github.com/CesiumGS/glTF/commit/b3f48c2a594ac1c6319dd0087b4f5fef2218a832

PR #479, opened 2021-09-01
add and use GlobalPropertyInteger for feature table lengths

krupkad:
- Add and use `GlobalPropertyInteger` to properly serialize/deserialize feature table lengths
- Rename `GlobalPropertyScalar` -> `GlobalPropertyNumber` to be consistent with JSON primitive names
- Remove the one-element array variant for both of these

### DIFF
--- a/specification/TileFormats/Batched3DModel/README.md
+++ b/specification/TileFormats/Batched3DModel/README.md
@@ -160,7 +160,7 @@ in the CesiumJS implementation of 3D Tiles.
 * [`Batched 3D Model Feature Table`](#reference-batched-3d-model-feature-table)
     * [`BinaryBodyReference`](#reference-binarybodyreference)
     * [`GlobalPropertyCartesian3`](#reference-globalpropertycartesian3)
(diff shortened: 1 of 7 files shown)


lilleyse:
Could you also update the property reference sections in for `TileFormats/Batched3DModel`, `TileFormats/Instanced3DModel`, etc? If you do a global search for `globalPropertyScalar` you'll notice there are still some occurrences.


Normally we use https://github.com/CesiumGS/wetzel to generate Markdown from JSON schema, but since it can be finicky you should just do a simple find/replace.

lilleyse:
Does anything actually use `globalPropertyNumber`?

krupkad:
@lilleyse no but neither is `globalPropertyCartesian4`, and there's no real "cost" to leaving it in

lilleyse:
Fixed up the grammar in a few places: https://github.com/CesiumGS/3d-tiles/commit/65a873c8ffcef7b335f3a9524864c9db5bbc58b6

Thanks @krupkad

PR #480, opened 2021-09-15
typo

pasu:
I think it should be URL instead of URI?


ptrgags:
@pasu correct, thanks for pointing this out!

PR #487, opened 2021-09-29
Add `TILE_GEOMETRIC_ERROR` semantic

lilleyse:
In [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_implicit_tiling) each child's geometric error is half of its parent's geometric error, but this imposes constraints on how the data can be tiled; in some cases you may want child tiles to be half as detailed as parent tiles but in other cases you may not, and it may vary a lot depending on the decimation algorithm. This PR adds a new semantic for implicit tile metadata that let's you specify geometric error for each tile in a subtree.

We discussed an alternate approach of specifying a geometric error factor for each subsequent level in the tree (to override the default 0.5), but it seemed better to just give full control like this.

This also gets us closer to being able to represent the whole JSON tree in binary form, which isn't really an immediate effort but is always worth thinking about.


ptrgags:
This looks good to me, thanks @lilleyse!

PR #490, opened 2021-10-05
Core Metadata: Remove raster format.

donmccurdy:
Removes raster format. Storing metadata in texture channels remains a valid implementation of the Core Metadata specification, but is not within the scope of the document itself.

An implementation of raster storage is already described in similar depth by `EXT_mesh_features`, see https://github.com/CesiumGS/glTF/pull/21.


donmccurdy:
@ptrgags rebase is done ✅

PR #493, opened 2021-10-06
Core Metadata: Language pass for binary table format

donmccurdy:
This PR is branched from https://github.com/CesiumGS/3d-tiles/pull/490 (which removes the raster format); it may be better to review that PR first.

Changes:
- Remove explicit definition of "encodings", referring only to "binary table format" and "JSON table format". I don't have a strong preference between the word "encoding" and "format", but didn't want to include both terms if we aren't defining a raster format here.
- Small restructuring of header hierarchy
- Language and clarity pass on binary data format, particularly language related to arrays.

Next PR will add language changes for JSON table format.

### DIFF
--- a/specification/Metadata/README.md
+++ b/specification/Metadata/README.md
@@ -22,10 +22,12 @@ Draft
 - [Overview](#overview)
 - [Concepts](#concepts)
 - [Schemas](#schemas)
+  - [Overview](#overview-1)
   - [Version](#version)
   - [Enums](#enums)
   - [Classes](#classes)
   - [Properties](#properties)
+    - [Overview](#overview-2)
     - [ID](#id)
     - [Name](#name)
     - [Description](#description)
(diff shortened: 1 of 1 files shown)


ptrgags:
Oh and make sure to resolve the merge conflict

ptrgags:
oh to answer the questions above:

>I suggest that we rephrase this in terms of appropriate precision instead of highest precision. There is no reason to use Float64 to store the result of a denormalized uint8.

>Not sure what to do here for 64-bit integers, but it goes beyond this spec (thinking back to noData/min/max in 3DTILES_metadata, exactness is important). I'm kind of ok with strings for values outside the safe range [-(2**53)+1, (2**53)-1] but yeah, that's kind of messy.

yeah that's tricky... I would still allow 64-bit integer types anywhere, just maybe put an implementation note that JSON doesn't support values outside the safe range.

e.g. we wouldn't want to rule out a `noData: -1` for a 64-bit type just because extremely large values aren't exactly representable.

Yeah "appropriate" precision makes sense here.

>(2c) Existing: Floating point numbers must be representable as IEEE-754 floating point.

keep the existing wording by @lilleyse's reasoning. Though do mention that it's `IEEE-754` to be more specific.

donmccurdy:
> But yes, I think we should leave [issues with values outside double precision IEEE-754 safe range] as an implementation note for now.

Sounds good to me, OK with something like this?

> **Implementation note:** Many JSON serialization libraries support only numeric values that can be represented as IEEE-754 double precision floating point numbers. Authoring tools should use double (or lower) precision IEEE-754 types for best compatibility.

***

> The thing with "should" is does that imply that the number could be encoded in some non-IEEE-754 format?

I think that's the situation, though — "should" might be the best we can do here. We're supporting int64 and uint64, and those values in JSON encoding (or their min/max in schema for binary encoding) could be outside of the IEEE-754 double precision safe range. Similarly, the glTF specification only uses "should", perhaps partly to require JSON encoders to do things that the JSON spec does not itself require.

ptrgags:
The updates look good, thanks @donmccurdy!

Issue #495, opened 2021-10-10
Extensions properties should point to extensions schema


javagl:
The `extensions` property in the schema files of extensions is currently defined as 

    "extensions": {},

(for example, [here](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/extensions/3DTILES_content_gltf/schema/tileset.3DTILES_content_gltf.schema.json#L23), but it appears to be done consistently like that for the extensions).

In the core schema, the extensions are pointing to the [`extension.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/specification/schema/extension.schema.json), via

        "extensions" : {
            "$ref": "extension.schema.json"
        },

(for example, [here](https://github.com/CesiumGS/3d-tiles/blob/3d-tiles-next/specification/schema/content.schema.json#L16), but it appears to be done consistently like that for the core)

---

I *think* that the extensions should also point to the `extensions.schema.json` (despite the difficulty that *might* arise from "nested extensions" - they are currently *required*, e.g for combining `3DTILES_multiple_contents` and `3DTILES_metadata`).

---

The same is currently true for the [glTF `gltf.EXT_mesh_features.schema.json`](https://github.com/CesiumGS/glTF/blob/8e4fde4d06590f365d1ca9a729f80db33f6472e0/extensions/2.0/Vendor/EXT_mesh_features/schema/gltf.EXT_mesh_features.schema.json#L32), which should similarly point  to the [glTF `extension.schema.json`](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/extension.schema.json).

lilleyse:
@javagl good timing - I already did this in https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next-json-schema-2020-12 and plan to open a PR soon.

lilleyse:
Fixed in https://github.com/CesiumGS/3d-tiles/pull/497

javagl:
I think they are still explicitly listed, e.g. https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/boundingVolume.schema.json#L40 - I hope it's OK to reopen

lilleyse:
3D Tiles schemas are mostly followed what glTF is doing. If glTF changes we should change here as well, but until then maybe not worry about it?

javagl:
[...]

PR #497, opened 2021-10-10
JSON schema cleanup

lilleyse:
Part of https://github.com/CesiumGS/3d-tiles/issues/495

While working on a 3D Tiles schema to C++ code generator in cesium-native and I started to clean up the 3D Tiles schema files so that they are more similar to glTF and produced better generated code. The changes include:

* Switched from JSON schema draft-4 to JSON schema 2020-12
* Formatted the schema files with the built-in VSCode JSON formatter (see below)
* Removed `"additionalProperties" : false` from several objects since extensions are allowed to add new properties to objects. This is a remnant of the glTF 1.0 schema which 3D Tiles was originally modeled after.
* Removed enum arrays in favor of glTF-style enums ([example](https://github.com/KhronosGroup/glTF/blob/f8c5f84c974508bcf4eadbc4943f4f7dcb1c3d89/specification/2.0/schema/accessor.schema.json#L26-L60))
* Copied glTF's way of handling extensible objects. Every object now extends `tilesetProperty.schema.json`, which is similar to [`glTFProperty.schema.json`](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/glTFProperty.schema.json)
* Renamed `3DTILES_batch_table_hierarchy.json` to `3DTILES_batch_table_hierarchy.schema.json`
* Added description to `tileset.asset`
* Renamed `tile.content.schema.json`'s title from Tile Content to Content
* Remove outdated requirement in `tile.schema.json` that content is required for leaf tiles. This note was removed in the property reference in the main spec but was leftover in the schema.

#

VSCode `settings.json`
```json
    "[json]": {
        "editor.formatOnSave": true,
        "editor.defaultFormatter": "vscode.json-language-features"
    },
```

CC @javagl @ptrgags @donmccurdy


javagl:
> The internal definitions should be inheriting from `tilesetProperty.schema.json`

That's right but they also seem to (re-) declare the `extras` and `extensions`. To my understanding, these could be removed now.

ptrgags:
Looks good to me too, thanks @lilleyse!

PR #506, opened 2021-10-15
Update 3D Tiles Next landing page

ptrgags:
Today I noticed that the overview landing page is still pointing to `EXT_feature_metadata`, not `EXT_mesh_features`. Updated the links and the text to match.

@lilleyse could you review?

CC @donmccurdy @javagl


lilleyse:
I pushed a commit to replace other occurrences of EXT_feature_metadata with EXT_mesh_features

PR #513, opened 2021-10-19
Core Metadata: Rewrite 'JSON Table Format' example with less JSON

donmccurdy:
*Continued from https://github.com/CesiumGS/3d-tiles/pull/511.*

### DIFF
--- a/specification/Metadata/README.md
+++ b/specification/Metadata/README.md
@@ -369,7 +369,7 @@ Property values are encoded as their corresponding JSON types: numeric types are
 >
 > | id                  | type       | componentType | componentCount | enumType      | required |
 > |---------------------|------------|---------------|----------------|---------------|----------|
-> | floatProperty       | `"SINGLE"` | `"FLOAT64"`   |                |               | ✓        |
+> | floatProperty       | `"SINGLE"` (default) | `"FLOAT64"`   |                |               | ✓        |
 > | integerProperty     | `"SINGLE"` | `"INT32"`     |                |               | ✓        |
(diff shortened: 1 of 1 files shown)


javagl:
But... (sorry) ... : The `type` is `"SINGLE"` for most cases. I think this is (or should be) the default. The fact that it is the default is not mentioned. And one could argue that (if it is the default), it could be omitted in the table. 

(I don't have a strong opinion here, but wanted to mention it...)

More generally: Some of the constraints that had originally be imposed by the schema _might_ have to be mentioned at some point. Browsing back to an arbitrary example: [An old schema version](https://github.com/CesiumGS/3d-tiles/blob/3db74d71728072073d61cd1745b1b27ac46fe810/specification/Metadata/1.0.0/schema/class.property.schema.json#L36) contained statements like

> If `ARRAY` is used, then `componentType` must also be specified.

and so on. Maybe not _all_ of them have to be added (because some of them may be specific for the JSON encoding). But leaving that open could raise questions...

donmccurdy:
> If ARRAY is used, then componentType must also be specified.

I think this might be too prescriptive for a spec that is serialization-agnostic. In both 3DTILES_metadata and EXT_mesh_features the `componentType` property is always required. But an implementation could just as easily give it a default value, and I don't think we intend to say that a default value must be redundantly included with `"ARRAY"`? If there are other constraints that are missing from places that they ought to be mentioned let's start an issue or PR to gather them.

I've added @ptrgags's suggestion with the clarifying `(default)` on the first occurrence of `"SINGLE"` in this table. 👍

PR #514, opened 2021-10-19
Split offsetType into arrayOffsetType and stringOffsetType

lilleyse:
Same change as https://github.com/CesiumGS/glTF/pull/35 for `EXT_mesh_features`, now for `3DTILES_metadata`


ptrgags:
Looks good, thanks @lilleyse!

Issue #512, opened 2021-10-18
Update and extend the implicit tiling examples


javagl:
There are currently three examples for implicit tiling data sets. One of them is https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_implicit_tiling/examples/sparseOctree 

The tileset JSON says `"subtreeLevels": 2`.

The section https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_implicit_tiling#availability-bitstream-lengths says that the length of the tile availability bitstream is

`(N^subtreeLevels - 1)/(N - 1)`

For `N=8` and `subtreeLevels=2`, this should yield 9.

The section also says that

`lengthBytes = ceil(lengthBits / 8)`, which would be 2, in this case.

But the tileAvailability of the JSON of subtree `0/0/0/0.subtree` refers to `bufferView: 0`, which is said to have `"byteLength": 1`:
```
{
  "buffers" : [ {
    "byteLength" : 9
  } ],
  "bufferViews" : [ {
    "buffer" : 0,
    "byteOffset" : 0,
    "byteLength" : 1
  }, {
    "buffer" : 0,
    "byteOffset" : 1,
    "byteLength" : 8
  } ],
  "tileAvailability" : {
    "bufferView" : 0
  },
  "childSubtreeAvailability" : {
    "bufferView" : 1
  },
  "contentAvailability" : {
    "bufferView" : 0
  }
}
```` 

The same applies to the `basicExample`.

---

The examples should be updated to be _valid_, and extended to show different configurations, like sparse- and dense quadtrees and octrees with different heights and subtree-heights. 

Preferably, the examples should be documented extensively, including information that can otherwise only be derived from actually _reading_ the subtree files and _interpreting the data correctly_. This means that the JSON part that is contained in the subtree files should be part of the documentation, and the binary parts (i.e. the availability info) should be documented in a form that might say, for example,

> The availability bytes are `0xCAFEBEEF`, which is the bit pattern `01101010111101...`

so that implementors can easily verify that their interpretation of the data is correct.

---

(Rumours say that such examples could go into https://github.com/CesiumGS/3d-tiles-samples at some point, but making sure that the ones that are currently _here_ are _valid_ would be the first step, **if** these new examples are not created with a completely new mechanism...)

ptrgags:
[...]

PR #525, opened 2021-10-26
Add implicit tiling extension specification

javagl:
This adds the implicit tiling extension specification that was developed in the 3d-tiles-next branch, based on the latest state after https://github.com/CesiumGS/3d-tiles/pull/492.

Direct link to README.md for preview: https://github.com/javagl/3d-tiles/blob/implicit-tiling-spec-to-main/extensions/3DTILES_implicit_tiling/README.md


extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,946 @@
...
+}
+```
+
+In the example above, every tile in the subtree exists, but not every tile has content. 
+
+When all tiles exist, then their availability can be encoded by setting `tileAvailability.constant` to `1`, without needing an explicit bitstream.
+
+Only some tiles have content, and `contentAvailability.bufferView` indicates where the bitstream for the content availability is stored: The `bufferView` with index 0 refers to the `buffer` with index 0. This buffer does not have a `uri` property, and therefore refers to the _internal_ buffer. The `byteOffset` and `byteLength` indicate that the content availability bitstream is stored in the bytes `[0...11)` of the internal buffer.
+
+Some child subtrees exist, so `childSubtreeAvailability.bufferView` refers to another bitstream. The `bufferView` with index 1 refers to the buffer with index `1`. This buffer has a `uri` property, indicating that this second bitstream is stored in an external binary file.
```
lilleyse, line 394: This example should be quoted (like the examples in [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features))

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,946 @@
...
+* **implicit root tile** - A tile with the `3DTILES_implicit_tiling` extension, which denotes the root of an implicit tileset.
+* **octree** - A 3D subdivision scheme that divides each bounding volume into 8 smaller bounding volumes along the midpoint of the `x`, `y`, and `z` axes.
+* **quadtree** - A 2D subdivision scheme that divides each bounding volume into 4 smaller bounding volumes along the midpoint of the `x` and `y` axes.
+* **subtree** - A fixed-size section of the tree that contains availability information.
+* **subtree file** - A binary file storing information about a specific subtree.
+* **subdivision scheme** - A recursive pattern of dividing a parent tile into smaller child tiles occupying the same area. This is done by uniformly dividing the bounding volume of the parent tile.
+* **template URI** - A URI pattern containing tile coordinates for directly addressing tiles.
+* **tile** - A division of space that may contain content.
+* **tileset** - A hierarchical collection of tiles.
+* **tileset JSON** - A JSON file describing a tileset, as defined in the [3D Tiles specification](../../specification#tileset-json).
```
lilleyse, line 417: I'm kind of iffy on whether we need a glossary...

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,946 @@
...
+
+## Extension JSON Schema Reference
```
lilleyse, line 419: We decided to remove the JSON schema reference from other extensions, so it can be removed from here.

extensions/3DTILES_implicit_tiling/schema/tile.3DTILES_implicit_tiling.templateUri.schema.json
```diff
@@ -0,0 +1,7 @@
+{
```
lilleyse, line 1: Since the template uri is really used in two places - `subtrees` and `content` - I think it makes sense to rename this to `templateUri.schema.json`.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,946 @@
...
+  ],
+  "root": {
+    "boundingVolume": {
+      "region": [-1.318, 0.697, -1.319, 0.698, 0, 20]
+    },
+    "refine": "REPLACE",
+    "geometricError": 5000,
+    "content": {
+      "uri": "content/{level}/{x}/{y}.b3dm"
+    },
```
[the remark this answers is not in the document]
javagl: (I'll integrate the comments (also for the content-gltf one) soon, but a quick question on that one:) 

I'll have to re-scan the section and think more deeply about whether one could apply _any_ reasonable bounding volume here. Something like "a bounding volume of the highest level of the content" _might_ make sense for culling, to not send out the request.

So the question is: **SHOULD** it be omitted (under certain conditions?), or **MUST** it be omitted?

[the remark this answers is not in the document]
lilleyse: Since the content uri refers to the collective set of contents I don't know if we want the content bounding volume to apply to only the root tile's content, though that is a use case I was thinking about as well.

I'm leaning towards **MUST** be omitted.

Similar to my comment above we can say that content bounding volumes may be defined with tile metadata and point to the appropriate semantics in the Cesium Metadata Semantic Reference.

javagl:
@lilleyse The inlined comments have been addressed. 

Whether or not the glossary should be removed still has to be decided. I think it does not harm - how useful it is (considering that it is located at the _end_ of the document) is hard to say...

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,540 @@
...
+When all tiles exist, then their availability can be encoded by setting `tileAvailability.constant` to `1`, without needing an explicit bitstream.
+
+Only some tiles have content, and `contentAvailability.bufferView` indicates where the bitstream for the content availability is stored: The `bufferView` with index 0 refers to the `buffer` with index 0. This buffer does not have a `uri` property, and therefore refers to the _internal_ buffer. The `byteOffset` and `byteLength` indicate that the content availability bitstream is stored in the bytes `[0...11)` of the internal buffer.
+
+Some child subtrees exist, so `childSubtreeAvailability.bufferView` refers to another bitstream. The `bufferView` with index 1 refers to the buffer with index `1`. This buffer has a `uri` property, indicating that this second bitstream is stored in an external binary file.
+
+### Availability Packing
+
+Availability bitstreams are packed in binary using the format described in the [Booleans](../../specification/Metadata#booleans) section of the Cesium 3D Metadata Specification.
+
```
ptrgags, line 401: Not sure where to put this (Maybe make a new examples section here?) but I got feedback from Patrick that we want to show another use case: making a global tileset. Typically we do this with a "double-headed quadtree" (one quadtree for the western hemisphere and one for the eastern hemisphere). That would look like this:

```
{
  "asset": {
    "version": "1.0"
  },
  "geometricError": 8000,
  "extensionsUsed": [
    "3DTILES_implicit_tiling"
  ],
  "extensionsRequired": [
    "3DTILES_implicit_tiling"
  ],
  "root": {
    "boundingVolume": {
      "region": [-3.14159265359, -1.57079632679, 3.14159265359, 1.57079632679, 0, 1000]
    },
    "refine": "REPLACE",
    "geometricError": 4000,
    "children": [
      {
        "boundingVolume": {
          "region": [-3.14159265359, -1.57079632679, 0, 1.57079632679, 0, 1000]
        },
        "content": {
          "uri": "content/western-hemisphere/{level}/{x}/{y}.b3dm"
        },
        "geometricError": 2000,
        "extensions": {
          "3DTILES_implicit_tiling": {
            "subdivisionScheme": "QUADTREE",
            "maximumLevel": 20,
            "subtrees": {
              "uri": "subtrees/eastern-hemisphere/{level}/{x}/{y}.subtree"
            },
            "subtreeLevels": 7
          }
        }
      },
      {
        "boundingVolume": {
          "region": [0, -1.57079632679, 3.14159265359, 1.57079632679, 0, 1000]
        },
        "content": {
          "uri": "content/eastern-hemisphere/{level}/{x}/{y}.b3dm"
        },
        "geometricError": 2000,
        "extensions": {
          "3DTILES_implicit_tiling": {
            "subdivisionScheme": "QUADTREE",
            "maximumLevel": 20,
            "subtrees": {
              "uri": "subtrees/eastern-hemisphere/{level}/{x}/{y}.subtree"
            },
            "subtreeLevels": 7
          }
        }
      }
    ]
  }
}

```

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -0,0 +1,540 @@
...
+        "subtrees": {
+          "uri": "subtrees/{level}/{x}/{y}.subtree"
+        },
+        "subtreeLevels": 7
+      }
+    }
+  }
+}
+```
+The `content` of an implicit tile must not have an associated `boundingVolume` property, but the [`3DTILES_metadata`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/extensions/3DTILES_metadata#implicit-tile-metadata) extension still allows defining bounding volumes for the content of implicit tiles. The possible [Semantics](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/specification/Metadata/Semantics) of the metadata include semantics like [`CONTENT_BOUNDING_BOX`](https://github.com/CesiumGS/3d-tiles/tree/3d-tiles-next/specification/Metadata/Semantics#content_bounding_box) that can be used to associate bounding volumes with the content of implicit tiles, for all tiles that are available in the implicit tree.
```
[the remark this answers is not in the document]
javagl: Sure, it's a dilemma: We can either have the *PR state* being "broken" (in that the links point to places that are not available in the PR), or the _merged_ state can be "broken" (in that it points to `3d-tiles-next`, which may be deleted soon after merging). 

I'm a bit on the fence when it comes to absolute vs. relative links in general, but AFAICT, `3d-tiles` mainly uses relative ones, so I think committing this is fine.

extensions/3DTILES_implicit_tiling/README.md
javagl: Note entirely sure what to do with this one. In `content-gltf`, there was an example, and Sean suggested to remove it, because the examples should be in `3d-tiles-samples` in the future (and then, we could add a link here). 

Maybe we'd defer this until we actually _have_ examples? 

(I mean, real examples. The ones that you linked to in https://github.com/CesiumGS/3d-tiles/issues/512#issuecomment-952284192 are  a bit "trivial" - _none_ of them has more than one subtree layer...)

ptrgags: okay fair, I'll open an issue for this.

ptrgags: I left a comment here: https://github.com/CesiumGS/3d-tiles/issues/512#issuecomment-956636721

ptrgags:
@javagl thank you, the updates look good. I agree, let's keep the glossary for now. I'm merging this now.

PR #530, opened 2021-10-28
Add Core Metadata specification

donmccurdy:
Changes:

- Add `specification/Metadata/` section defining metadata schema, encodings, and semantics.

Preview:

- [specification/Metadata/README.md](https://github.com/donmccurdy/3d-tiles/blob/core-metadata-to-main/specification/Metadata/README.md)
- [specification/Metadata/Semantics/README.md](https://github.com/donmccurdy/3d-tiles/blob/core-metadata-to-main/specification/Metadata/Semantics/README.md)

To sync from `3d-tiles-next`:

```shell
git checkout 3d-tiles-next -- specification/Metadata
```


lilleyse:
Wondering about naming...

Right now there's some asymmetry in the names of these two documents

**Cesium 3D Metadata Specification**
**Cesium Metadata Semantic Reference**

... and sometimes we also call it **Core Metadata**.


* Should we keep "Cesium" in the title?
* Should we keep "3D" in the title?
* Should we have a more unique name?

I kind of like 3D because there's a nice parallel between 3D Tiles and 3D Metadata and this spec is designed for 3D formats  (even though the metadata itself is usually columnar).

What does everyone think about these names?

**3D Metadata Specification**
**3D Metadata Semantic Reference**

CC https://github.com/CesiumGS/3d-tiles/pull/456#discussion_r584380769

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,190 @@
...
+Each semantic is defined in terms of its meaning, and the datatypes it may assume. Datatype specifications include "type", "component type", and "component count" attributes as defined by the [Cesium 3D Metadata Specification](../).
+
+For use of semantics in extensions of specific standards, see:
+
+* [`3DTILES_metadata`](../../../extensions/3DTILES_metadata) (3D Tiles 1.0) — Assigns metadata to tilesets, tiles, or tile contents
+* [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) (glTF 2.0) —  Assigns metadata to subcomponents ("features") of geometry or textures
+
+## Contents
+
+- [Cesium Metadata Semantic Reference](#cesium-metadata-semantic-reference)
```
lilleyse, line 18: `<!-- omit in toc -->` should be added above certain sections so that they don't get included in the table of contents, like in Cesium 3D Metadata.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,190 @@
...
+This document provides common definitions of meaning ("semantics") used by metadata properties in 3D Tiles 1.0 and glTF 2.0. Tileset authors may define their own application- or domain-specific semantics separately.
+
+Semantics describe how properties should be interpreted. For example, an application that encounters the `ID` or `NAME` semantics while parsing a dataset may use these values as unique identifiers or human-readable labels, respectively.
+
+Each semantic is defined in terms of its meaning, and the datatypes it may assume. Datatype specifications include "type", "component type", and "component count" attributes as defined by the [Cesium 3D Metadata Specification](../).
+
+For use of semantics in extensions of specific standards, see:
+
+* [`3DTILES_metadata`](../../../extensions/3DTILES_metadata) (3D Tiles 1.0) — Assigns metadata to tilesets, tiles, or tile contents
+* [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) (glTF 2.0) —  Assigns metadata to subcomponents ("features") of geometry or textures
```
lilleyse, line 14: Cesium 3D Metadata has a similar list of extensions in its overview but with slightly different wording. We should make them consistent.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+Units for all linear distances are meters, and all angles are radians.
+
+### Tile
+
+#### Overview
+
+`TILE_*` semantics provide meaning for properties associated with a particular tile, and should take precedence over equivalent metadata on parent objects, as well as over values derived from subdivision schemes like [3DTILES_implicit_tiling](../../../extensions/3DTILES_implicit_tiling).
+
+In particular, `TILE_BOUNDING_BOX`, `TILE_BOUNDING_REGION`, and `TILE_BOUNDING_SPHERE` semantics each define a tighter bounding volume for a tile than is implicitly calculated from [3DTILES_implicit_tiling](../../../extensions/3DTILES_implicit_tiling). If more than one of these semantics are available for a tile, clients may select the most appropriate option based on use case and performance requirements.
```
lilleyse, line 78: > tighter bounding volume

Not sure if we should say tighter since there are some cases where it may be larger, like using `TILE_BOUNDING_SPHERE` when the implicit root tile is a region. Though it's still the most common use case so it should be mentioned, just in a different way.

specification/Metadata/Semantics/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+The geometric error of the tile. Equivalent to `tile.geometricError`.
+
+* Type: `FLOAT32` or `FLOAT64`
+
+### Content
+
+#### Overview
+
+`CONTENT_*` semantics provide meaning for properties associated with a particular tile content, and should take precedence over equivalent metadata on the containing tile.
```
lilleyse, line 135: > and should take precedence over equivalent metadata on the containing tile

I see them more as complementary than taking precedence over the other. The tile's bounding volume is used for spatial coherence while the content bounding volume is used for finer grained culling, both serving different purposes.

I guess the main point is that `3DTILES_implicit_tiling` doesn't implicitly define content bounding volumes so this is how to define them.

lilleyse: We should probably mention that if `3DTILES_multiple_contents` is used this defines the content bounding volume for all the contents in the tile.

donmccurdy:
> What does everyone think about these names?
> 
> - 3D Metadata Specification
> - 3D Metadata Semantic Reference

That feels like the right name structure to me, dropping the term "Core", as well as the term "Cesium" if it doesn't normally appear in 3D Tiles specification names. There are probably other terms similar to "3D" that could be substituted for it, like "Scene", "Spatial", "GPU-ready", etc. I'm happy enough with 3D, but since that's the only word in the title that uniquely identifies this from other metadata we should probably discuss it briefly.

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+The specification does not enumerate or define the semantic meanings of metadata, and assumes that separate specifications will define semantics for their particular application or domain. One example is the [Cesium Metadata Semantic Reference](./Semantics/) which defines built-in semantics for 3D Tiles and glTF. Identifiers for externally-defined semantics can be stored within the Cesium 3D Metadata Specification.
+
+## Concepts
+
+This specification defines metadata schemas and methods for encoding metadata.
+
+**Schemas** contain a set of **classes** and **enums**. A class represents a category of similar entities, defined as a set of **properties**. Each property describes values of a particular type. An enum defines a set of named values representing a single value type, and may be referenced by class properties. Schema definitions do not describe how entities or properties are stored, and may be represented in a file format in various ways. Schemas can be reused across multiple assets or even file formats.
```
ptrgags, line 75:
```suggestion
**Schemas** contain a set of **classes** and **enums**. A class represents a category of similar entities, and is defined as a set of **properties**. Each property describes values of a particular type. An enum defines a set of named values representing a single value type, and may be referenced by class properties. Schema definitions do not describe how entities or properties are stored, and may be represented in a file format in various ways. Schemas can be reused across multiple assets or even file formats.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+**Metadata**, as used throughout this specification, refers to any association of 3D content with entities and properties, such that entities represent meaningful units within an overall structure. Other common definitions of metadata, particularly in relation to filesystems or networking as opposed to 3D content, remain outside the scope of the document.
+
+Property values are stored with flexible representations to allow compact transmission and efficient lookups. This specification defines two possible [storage formats](#storage-formats).
+
+## Schemas
+
+### Overview
+
+A schema defines the organization and types of metadata used in 3D content, represented as a set of classes and enums. Class definitions are referenced by entities whose metadata conforms to the class definition, providing a consistent and machine-readable structure for all entities in a dataset.
```
ptrgags, line 89:
```suggestion
A schema defines the organization and types of metadata used in 3D content, represented as a set of classes and enums. Class definitions are referenced by entities whose metadata conforms to the class definition. This provides a consistent and machine-readable structure for all entities in a dataset.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+> **Example:** A typical ID / Name pair, in English, would be `localTemperature` and `"Local Temperature"`. In Japanese, the name might be represented as "きおん". Because IDs are restricted to alphanumeric characters and underscores, use of helpful property names is essential for clarity in many languages.
+
+#### Description
+
+Descriptions (`description`) provide a human-readable explanation of a property, its purpose, or its contents. Typically at least a phrase, and possibly several sentences or paragraphs. To provide a machine-readable semantic meaning, a property must also define a [semantic](#semantic).
+
+#### Semantic
+
+Property IDs, names, and descriptions do not impute meaning. To provide such a meaning, properties may be assigned a semantic identifier string (`semantic`), indicating how the property's content should be interpreted. Semantic identifiers may be defined by the [Cesium Metadata Semantic Reference](./Semantics/) or by external semantic references, and may be application-specific. Identifiers should be uppercase, with underscores as word separators.
```
ptrgags, line 152: TIL impute is a word... let's pick a more common wording that's consistent with other sections.

```suggestion
Property IDs, names, and descriptions do not have an inherent meaning. To provide a machine-readable meaning, properties may be assigned a semantic identifier string (`semantic`), indicating how the property's content should be interpreted. Semantic identifiers may be defined by the [Cesium Metadata Semantic Reference](./Semantics/) or by external semantic references, and may be application-specific. Identifiers should be uppercase, with underscores as word separators.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+>
+> | property  | componentType | required | noData    |
+> |:----------|:--------------|:---------|:----------|
+> | height    | "FLOAT32"     | ✓        |           |
+> | species   | "STRING"      |          | "Unknown" |
+> | leafColor | "STRING"      | ✓        |           |
+
+#### ID
+
+IDs (`id`) uniquely identify a property within a class, and must contain only alphanumeric characters and underscores. IDs should be human-readable (wherever possible) and camel-case. When IDs subject to these restrictions are not sufficiently clear for human readers, applications should also provide a property *name*.
```
ptrgags, line 138:
```suggestion
IDs (`id`) uniquely identify a property within a class, and must contain only alphanumeric characters and underscores to be machine-parseable. IDs should be camel case strings that are human-readable (wherever possible). When IDs subject to these restrictions are not sufficiently clear for human readers, applications should also provide a property *name*.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+| STRING  | A sequence of characters                                                  |
+| ENUM    | An enumerated type                                                        |
+
+Floating-point properties (`FLOAT32` and `FLOAT64`) must not include values `NaN`, `+Infinity`, or `-Infinity`.
+
+[Enum properties](#enums) are denoted by `ENUM`. An enum property must additionally provide the ID of the specific enum it uses, referred to as its enum type (`enumType`).
+
+> **Implementation Note:** Developers of authoring tools should be aware that many JSON implementations support only numeric values that can be represented as IEEE-754 double precision floating point numbers. Floating point numbers should be representable as double precision IEEE-754 floats when encoded in JSON. When those numbers represent property values (such as `noData`, `min`, or `max`) having lower precision (e.g. single-precision float or 8-bit or 16-bit integer), the values should be rounded to the same precision in JSON to avoid any potential boundary mismatches. Numeric property values encoded in binary storage are unaffected by these limitations of JSON implementations.
```
ptrgags, line 207:
```suggestion
> **Implementation Note:** Developers of authoring tools should be aware that many JSON implementations support only numeric values that can be represented as IEEE-754 double precision floating point numbers. Floating point numbers should be representable as double precision IEEE-754 floats when encoded in JSON. When those numbers represent property values (such as `noData`, `min`, or `max`) having lower precision (e.g. single-precision float, 8-bit integer or 16-bit integer), the values should be rounded to the same precision in JSON to avoid any potential boundary mismatches. Numeric property values encoded in binary storage are unaffected by these limitations of JSON implementations.
```

ptrgags: also i'm not sure what "boundary mismatches" mean here, I would reword this to be more specific.

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+#### Normalized Values
+
+Normalized properties (`normalized`) provide a compact alternative to larger floating-point types. Normalized values are stored as integers, but when accessed are transformed to floating-point form according to the following rules:
+
+* Unsigned integer values (`UINT8`, `UINT16`, `UINT32`, `UINT64`) must be rescaled to the range `[0.0, 1.0]` (inclusive)
+* Signed integer values (`INT8`, `INT16`, `INT32`, `INT64`) must be rescaled to the range `[-1.0, 1.0]` (inclusive)
+
+> **Implementation Note:** Depending on the implementation and the chosen integer type, there may be some loss of precision in values after denormalization. For example, if the implementation uses 32-bit floating point variables to represent the value of a normalized 32-bit integer, there are only 23 bits in the mantissa of the float, and lower bits will be truncated by denormalization. Client implementations should use higher precision floats when appropriate for correctly representing the result.
```
ptrgags, line 216: Not 100% if it's "unnormalize" or "denormalize", but we've been using "unnormalize" in our impelementation. 

```suggestion
> **Implementation Note:** Depending on the implementation and the chosen integer type, there may be some loss of precision in values after unnormalizing. For example, if the implementation uses 32-bit floating point variables to represent the value of a normalized 32-bit integer, there are only 23 bits in the mantissa of the float, and lower bits will be truncated after unnormalizing. Client implementations should use higher precision floats when appropriate for correctly representing the result.
```

specification/Metadata/README.md
```diff
@@ -0,0 +1,446 @@
...
+
+`VECN` and `MATN` types are treated as fixed-length numeric arrays.
+
+Each expression in the table above defines an index into the underlying property array. For a property array of `FLOAT32` components, index `3` would correspond to <u>_byte_</u> offset `3 * sizeof(FLOAT32) = 12` within that array. For an array of `BOOLEAN` components, offset `3` would correspond to <u>_bit_</u> offset `3`.
+
+> **Example:** Five variable-length arrays of UINT8 components, binary-encoded in a buffer. The associated property definition would be `type = "ARRAY"`, and `componentType = "UINT8"`, `componentCount = undefined` (variable-length).
+>
+> <img src="figures/array-of-ints.png"  alt="Variable-length array" width="640px">
+
+> **Example:** Two variable-length arrays of strings, binary-encoded in a buffer. The associated property definition would be `type = "ARRAY"`, `componentType = "STRING"`, `componentCount = undefined` (variable-length).
```
ptrgags, line 348: Noticed one thing to clarify in this example:

```suggestion
> **Example:** Two variable-length arrays of strings, binary-encoded in a buffer. The associated property definition would be `type = "ARRAY"`, `componentType = "STRING"`, `componentCount = undefined` (variable-length). Observe that the last element of the array offset buffer points to the last element of the string offset buffer. This is because the last valid string offset is the next-to-last element of the string offset buffer.
```

ptrgags (commented):
I had a few small things I noticed.

specification/Metadata/Semantics/README.md
lilleyse: > a more specific bounding volume...

This wording sounds good to me

specification/Metadata/README.md
donmccurdy: I'm not sure what "machine-parseable" means here, or why non-alphanumeric characters aren't. We probably don't want to get into "should be valid variable names in [language]" here, would it be OK to include just the 2nd change?

specification/Metadata/README.md
donmccurdy: OK – I've edited your suggestion to remove the word "boundary", I think just "mismatches" is likely clear enough here.

Context: The phrase "bounding mismatch" is taken [from the glTF specification](https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#accessors-bounds), albeit in a section on min/max bounds where the term "boundary" is defined by context.

specification/Metadata/README.md
donmccurdy: I think I disagree with this change – the two words have different meanings and this is the correct one for the context:

- unnormalized: value has not been normalized.
- denormalized: value has been decoded from a normalized form.

Wikipedia makes similar distinctions between [denormalized](https://en.wikipedia.org/wiki/Denormalization) and [unnormalized](https://en.wikipedia.org/wiki/Unnormalized_form) forms, referring to relational tables and not GPU attributes though.

specification/Metadata/README.md
donmccurdy: Ah good clarification! 👍

specification/Metadata/Semantics/README.md
donmccurdy:
```suggestion
In particular, `TILE_BOUNDING_BOX`, `TILE_BOUNDING_REGION`, and `TILE_BOUNDING_SPHERE` semantics each define a more specific bounding volume for a tile than is implicitly calculated from [3DTILES_implicit_tiling](../../../extensions/3DTILES_implicit_tiling). If more than one of these semantics are available for a tile, clients may select the most appropriate option based on use case and performance requirements.
```

donmccurdy:
Most threads resolved – two questions left @ptrgags

specification/Metadata/README.md
ptrgags: that's fair, just do the second change

specification/Metadata/README.md
ptrgags: okay, leave it as-is.

ptrgags:
Looks good, thanks @donmccurdy!

PR #535, opened 2021-11-01
3D Metadata Semantic Reference: remove overview sections from TOC

lilleyse:
Removes the "Overview" sections from the table of contents to improve scanability of the listed semantics. Not super tied to this change if others feel like the "Overview" sections are helpful to keep.

### DIFF
--- a/specification/Metadata/Semantics/README.md
+++ b/specification/Metadata/Semantics/README.md
@@ -1,6 +1,7 @@
 <!-- omit in toc -->
 # Cesium Metadata Semantic Reference
 
+<!-- omit in toc -->
 ## Overview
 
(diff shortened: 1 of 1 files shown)


donmccurdy (approved):
The change looks good to me!

I'm using the _Overview_ header (rather than a paragraph directly following the parent header) based on the [ISO directive on "hanging paragraphs"](https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor314), but we aren't using numbered sections here anyway, and it certainly isn't necessary to follow that guideline if we'd rather not.

ptrgags:
Oh @lilleyse I forgot about this PR, could we rebase this on `main`?

lilleyse:
@donmccurdy ah, thanks for the context there!

@ptrgags should be ready to go now

ptrgags:
Looks good, thanks @lilleyse

PR #536, opened 2021-11-01
Core Metadata: Apply feedback from #530 (pt 2)

donmccurdy:
Syncing `3d-tiles-next` to latest feedback on #530. 

~~There are a couple comment threads still open on #530 so this PR can probably be left open until I add those changes here.~~ This PR is caught up with the feedback.


ptrgags:
Looks good, thanks @donmccurdy!

PR #542, opened 2021-11-03
Make the 3D Tiles Next page to a landing page

ptrgags:
the `3D_TILES_NEXT.md` page is not how we usually like to store Markdown documentation. Usually we like to make a folder and have a README.md so it shows up nicer in GitHub.

@donmccurdy I did this targeting the `3d-tiles-next` branch, wasn't sure what's best for your workflow when you work on revising this document.

PR #543, opened 2021-11-04
Move the 3D Tiles Next page to a subfolder

ptrgags:
Same thing as #542 , but targeting `main` this time, since the `3d-tiles-next` branch is now considered outdated.

@donmccurdy could you review? I adjusted the relative links and mention of 3D Metadata Specification.


lilleyse:
From https://github.com/CesiumGS/3d-tiles/pull/542#issuecomment-959554547

> - The other top-level directories are lowercase, should we follow that convention? Maybe just `next/README.md`?

👍 I like `next` or `3d-tiles-next`

ptrgags:
@lilleyse updated!

PR #545, opened 2021-11-08
Added 3D Tiles Next overview

javagl:
The structure is analogous to the 3D Tiles Overview:

- The main PDF is [`next/3d-tiles-next-overview.pdf`](https://github.com/javagl/3d-tiles/blob/add-3d-tiles-next-overview/next/3d-tiles-next-overview.pdf)
- A combined image (for showing in a README) is in [`next/figures/3d-tiles-next-overview-pages-combined-3x2.jpg`](https://github.com/javagl/3d-tiles/blob/add-3d-tiles-next-overview/next/figures/3d-tiles-next-overview-pages-combined-3x2.jpg)
- The source files (SVG and resources and a small README) are in the [`next/overview`](https://github.com/javagl/3d-tiles/tree/add-3d-tiles-next-overview/next/overview) directory.


donmccurdy:
Are these SVGs the source files you'd open for editing in Inkscape? I use Affinity Designer, and for me these particular SVGs are not really edit-friendly (each line of text is a separate object, so any wording changes quickly cause wrapping issues) but the PDF is relatively easy to edit, with lines of text wrapping as expected. That was surprising to me since it looks like the PDF is generated _from_ the SVGs. 🤔

In any case ... the structure, level of detail, and diagrams look great! I saw Peter has already left some wording suggestions in an annotated PDF; I'll stay out of wording comments for the time being. I did spend some time in https://github.com/CesiumGS/3d-tiles/pull/546 coming up with ~1 paragraph summaries of each extension, in case any of that language is helpful.

javagl:
> Are these SVGs the source files you'd open for editing in Inkscape? I use Affinity Designer, and for me these particular SVGs are not really edit-friendly (each line of text is a separate object, so any wording changes quickly cause wrapping issues) 

The SVGs indeed are the source. The text should not be one object per line, but one object per paragraph (even though e.g. headlines are usually one object). I'm usually doing the line breaks in these paragraphs manually, depending on the place where the wrapping should occur. Maybe that's a level of "tweaking" that reduces the edit-friendliness for others...

I don't have an option to test this with Affinity Designer, but am curious: Do you see any difference in editability in the attached SVG?

[TextLayoutTest.zip](https://github.com/CesiumGS/3d-tiles/files/7500799/TextLayoutTest.zip)

(If this is the case, I could try to use the approach that is easier to edit for others in the future).

javagl:
>  I did spend some time in #546 coming up with ~1 paragraph summaries of each extension, in case any of that language is helpful.

These summaries are indeed _good_, along multiple dimensions of the term "good": Concise, precise, convincing, easy to understand, ...  The PDF could indeed be _better_ if I adopted some of these summaries accordingly. But given the timeline, maybe that's something to consider for a version 1.1.0 of the PDF...

lilleyse:
Could we swap out the plane with a different model, like a building or tree model?

javagl:
I just used the plane because of https://sandcastle.cesium.com/?src=3D%20Models.html - but if course, is there any preferred glTF asset available (assuming that it should really _be_ a glTF asset), then I can replace that accordingly. 

There don't seem to be buildings or trees in https://github.com/CesiumGS/cesium/tree/main/Apps/SampleData/models , though (the "WoodTower" from https://sandcastle.cesium.com/?src=Shadows.html would be ... meh, it's nice for the shadows demo, but not really a building ...)

ptrgags:
@javagl I read through the parts of the document you changed since my last round of feedback. Looks good! just two small tweaks (I wasn't clear enough in my previous comments)

[3d-tiles-next-overview (Annotated).pdf](https://github.com/CesiumGS/3d-tiles/files/7507475/3d-tiles-next-overview.Annotated.pdf)

javagl:
@ptrgags The comments have been integrated locally. I'll wait to see whether there is further feedback, and do one "consistent" update of the update of the PDF and the `combined-3x2` image then.

lilleyse:
@javagl maybe use this lantern model: https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/Lantern

image

My issue with the plane is that it's not the type of content typically in a 3D Tiles payload

javagl:
The lantern is ... how should I say... not really "realistic", in that it will also not be found in a real-world 3D Tiles payload. I think that the plane conveyed the point that one can use _any_ glTF model directly in 3D Tiles (even if it's not a classical "tile"), and one can imagine some real-world context like a flight tracker application. @ptrgags @donmccurdy  any opinions on this?

donmccurdy:
I don't think the illustration will ruin anyone's experience of the Overview PDF, so feel free to ignore this comment and maybe we file an issue to revise later, but — the ideal example would be something that looks like a spatial 'chunk' of a larger whole, probably? If so, lots of [outdoor photogrammetry scans on Sketchfab](https://sketchfab.com/search?features=downloadable&licenses=322a749bcfa841b29dff1e8a1bb74b0b&licenses=b9ddc40b93e34cdca1fc152f39b9f375&licenses=72360ff1740d419791934298b8b6d270&licenses=bbfe3f7dbcdd4122b966b85b9786a989&licenses=2628dbe5140a4e9592126c8df566c0b7&licenses=34b725081a6a4184957efaec2cb84ed3&licenses=7c23a1ba438d4306920229c12afcb5f9&licenses=783b685da9bf457d81e829fa283f3567&licenses=5b54cf13b1a4422ca439696eb152070d&q=rock+photogrammetry+scan&sort_by=-relevance&type=models) that would fit here.

[Example](https://sketchfab.com/3d-models/rocks-photogrammetry-scan-34022df84eb04b8a96606e5c58b0fa3b):

Screen Shot 2021-11-09 at 1 26 38 PM

javagl:
Two "concerns" here are 

1. The model has to be "iconic", as in "easy to recognize", because it is small (at least, I think that people might wonder what they see in an image like the one above)
2. The example is under a "Creative Commons Attribution-NonCommercial" license (but maybe one can find one that is closer to CC0...).

donmccurdy:
If iconic implies that it should be a common object readers can recognize and name, like a plane or a lamp, then I don't fully agree – but certainly we do want readers to be able to know what they're looking at. :) 

A few other ideas:

- [Rock Terrain](https://sketchfab.com/3d-models/rock-terrain-fa015203576b44dd9743d24d6c34b5e1) (CC Attribution)
- [Farm House](https://sketchfab.com/3d-models/abandoned-and-overgrown-farm-house-virginia-21f607c1110849f1a42745b8fc5b7274) (CC Attribution)

The building used in the [EXT_mesh_features screenshot](https://github.com/CesiumGS/3d-tiles/tree/main/next#metadata) might also be an option, if we hide the colored overlay.

ptrgags:
@javagl I'd say that the lantern is actually fine. While yes, it might be rare to find a _single_ lamp in isolation in a tileset, you certainly see them.

This example is photogrammetry, but I've seen instanced lampposts and trees used before in other datasets.

image

(from https://sandcastle.cesium.com/?src=3D%20Tiles%20Next%20Photogrammetry%20Classification.html&label=3D%20Tiles%20Next)

next/overview/README.md
```diff
@@ -0,0 +1,39 @@
+
+# 3D Tiles Next Overview
+
+[![](../figures/3d-tiles-next-overview-pages-combined-3x2.jpg)](../3d-tiles-next-overview.pdf)
```
ptrgags, line 4: Adding some alt text

```suggestion
[![Overview pages](../figures/3d-tiles-next-overview-pages-combined-3x2.jpg)](../3d-tiles-next-overview.pdf)
```

next/overview/README.md
```diff
@@ -0,0 +1,39 @@
+
+# 3D Tiles Next Overview
```
ptrgags, line 2: Now that #546 is merged, could you pull in `main` and add a link to the overview page (and perhaps preview image) to the `next` page?

ptrgags (commented):
@javagl The updates to the PDF look good. A couple small tweaks needed in the READMEs and then I think we can merge

javagl:
@ptrgags  I have added the "pages" image and link to the PDF on one main `next/README.md`

ptrgags:
@javagl updates look good. Thank you!

PR #552, opened 2021-11-12
Fix S2 cell ID examples

newtonallen3:


sanjeetsuhag:
@newtonallen3 Thank you for the fix! The suggested change looks correct to me. Please sign the [Cesium Contributor License Agreement](https://github.com/CesiumGS/cesium/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) and then I can merge this.

newtonallen3:
Thanks. After a delay related to learning about CLAs... I'm contributing this on behalf of Google, which has already signed a corporate CLA. So I believe I don't need to sign anything further. Is that correct?

sanjeetsuhag:
> I'm contributing this on behalf of Google, which has already signed a corporate CLA. So I believe I don't need to sign anything further. Is that correct?

Yes, you're correct. I can merge this now. Thanks again for the contribution @newtonallen3!

newtonallen3:
Thanks!

PR #565, opened 2021-11-15
Add an implementation note about tilesetVersion as a query parameter

### LINKED ISSUE #331: Add implementation note about how to use tilesetVersion
Add an implementation note explaining that `tilesetVersion` is often used as a query parameter when requesting tiles, so as not to grab outdated content from any caches.

CC @mramato

ptrgags:
Fixes #331

Specification bug fix: adds an implementation note that tilesetVersion can be used as a query parameter to avoid unwanted caching

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -540,6 +540,8 @@ The tileset JSON has four top-level properties: `asset`, `properties`, `geometri
 
 `asset` is an object containing metadata about the entire tileset. The `asset.version` property is a string that defines the 3D Tiles version, which specifies the JSON schema for the tileset and the base set of tile formats.  The `tilesetVersion` property is an optional string that defines an application-specific version of a tileset, e.g., for when an existing tileset is updated.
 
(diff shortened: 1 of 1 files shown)


specification/README.md
```diff
@@ -540,6 +540,8 @@ The tileset JSON has four top-level properties: `asset`, `properties`, `geometri
 
 `asset` is an object containing metadata about the entire tileset. The `asset.version` property is a string that defines the 3D Tiles version, which specifies the JSON schema for the tileset and the base set of tile formats.  The `tilesetVersion` property is an optional string that defines an application-specific version of a tileset, e.g., for when an existing tileset is updated.
 
+> **Implementation Note**: `tilesetVersion` can be used as a query parameter when requesting tiles to avoid using outdated content from a cache.
```
lilleyse, line 543:
```suggestion
> **Implementation Note**: `tilesetVersion` can be used as a query parameter when requesting content to avoid using outdated content from a cache.
```

ptrgags:
@lilleyse updated!

Issue #396, opened 2019-11-22
Conflict in Positions for Feature Table


shehzan10:
According to https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/TileFormats/FeatureTable/README.md#json-header, the feature table may contain arrays with property values directly. The example given there is that of the `POSITION` attribute for Instanced 3D Models.

However, https://github.com/AnalyticalGraphicsInc/3d-tiles/blob/master/specification/TileFormats/Instanced3DModel/README.md#instance-semantics says
> The value for each instance semantic must be a reference to the Feature Table binary body; they cannot be embedded in the Feature Table JSON header.

The same is said for Point Clouds, and Batched 3D Models don't have any per-feature semantics.

Confirmed with @lilleyse that this is a conflict in the spec.

lilleyse:
@Samulus this may a good 3D Tiles spec starter issue

PR #566, opened 2021-11-15
Allow property values to be stored in JSON header

javagl:
May fix https://github.com/CesiumGS/3d-tiles/issues/396 :  

The section for "Instance Semantics" said

> The value for each instance semantic must be a reference to the Feature Table binary body; they cannot be embedded in the Feature Table JSON header.

(analogously, for PNTS and "Point Semantics"). This statement is removed with this PR, because it is allowed to store these values in the JSON header as well. 

Note that this may be conisdered as a "breaking change": Existing implementations may have _relied_ on this restriction. Specifically, they may not handle the case where the values are supposed to be obtained from the JSON header arrays, as it is currently done at https://github.com/CesiumGS/cesium/blob/2fd0e8f7e4212bd1e7084299187f70597a6bbfd8/Source/Scene/Cesium3DTileFeatureTable.js#L97

### DIFF
--- a/specification/TileFormats/Instanced3DModel/README.md
+++ b/specification/TileFormats/Instanced3DModel/README.md
@@ -81,7 +81,6 @@ See [Property reference](#property-reference) for the `i3dm` feature table schem
 #### Instance semantics
 
 These semantics map to an array of feature values that are used to create instances. The length of these arrays must be the same for all semantics and is equal to the number of instances.
-The value for each instance semantic must be a reference to the Feature Table binary body; they cannot be embedded in the Feature Table JSON header.
 
 If a semantic has a dependency on another semantic, that semantic must be defined.
 If both `SCALE` and `SCALE_NON_UNIFORM` are defined for an instance, both scaling operations will be applied.
--- a/specification/TileFormats/PointCloud/README.md
+++ b/specification/TileFormats/PointCloud/README.md
@@ -76,7 +76,6 @@ See [Property reference](#property-reference) for the `pnts` feature table schem
 #### Point semantics
 
 These semantics map to an array of feature values that define each point. The length of these arrays must be the same for all semantics and is equal to the number of points.
-The value for each point semantic must be a reference to the Feature Table binary body; they cannot be embedded in the Feature Table JSON header.
 
 If a semantic has a dependency on another semantic, that semantic must be defined.
 If both `POSITION` and `POSITION_QUANTIZED` are defined for a point, the higher precision `POSITION` will be used.


lilleyse:
I think the better change would be to remove the second option in https://github.com/CesiumGS/3d-tiles/blob/main/specification/TileFormats/FeatureTable/README.md#json-header.

In the schema all the per-instance and per-point semantics refer to `featureTable.schema.json#/definitions/binaryBodyReference` so it's not actually possible to use the second option. Therefore I would consider removing the second option to be a "bug fix" rather than a breaking change.

javagl:
@Sean This may be a less trivial change then. I'll have to re-read some of the schemas to wrap my head around that, but from a quick glance, this may cause some inconsistencies:

- https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/pnts.featureTable.schema.json is said to be a `featureTable`. (All the properties that it defines are either `global*` or `binaryBodyReference`)
- https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/featureTable.schema.json#L76 says that each property may be a single value, a binary body reference, **or** an array of numbers. This part would have to be removed, potentially making existing data sets invalid.

(Maybe there is a claim or assumption that these `properties` may only ever be ones that are defined in the corresponding `pnts`/`i3dm` schema files, but that's not obvious for me right now...)

lilleyse:
> Maybe there is a claim or assumption that these `properties` may only ever be ones that are defined in the corresponding `pnts`/`i3dm` schema files, but that's not obvious for me right now...

In theory someone might have a custom content type (not `b3dm`, `pnts`, or `i3dm`) that has a feature table with array-based storage, but it's not very likely.

> https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/featureTable.schema.json#L76 says that each property may be a single value, a binary body reference, or an array of numbers. This part would have to be removed, potentially making existing data sets invalid.

Since it's unlikely that there is existing data to break I think it's fine to remove.

Just for more background, binary storage was a slightly later addition to the feature table spec and we just weren't diligent about removing the array-based storage option, but the intent was that per-feature semantics would always be stored in binary.

javagl:
To be sure: This means that the part that allows an array of numbers at https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/featureTable.schema.json#L83L89 should actually be removed from the *schema* (and not only from the bullet-point text), right?

If so, I'd probably close this PR and open another one (because it's the _opposite_ of what I did here). 

I'm a bit hesitant, because it's hard to know the implications. For example, are you sure that there is **not** some unit test in `CesiumJS/Specs` that tests whether *exactly this* (arrays in JSON) works, and which should, technically speaking, fail in the future...?

lilleyse:
> To be sure: This means that the part that allows an array of numbers at https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/featureTable.schema.json#L83L89 should actually be removed from the *schema* (and not only from the bullet-point text), right?

Yeah, let's remove it in both places.

> I'm a bit hesitant, because it's hard to know the implications. For example, are you sure that there is **not** some unit test in `CesiumJS/Specs` that tests whether *exactly this* (arrays in JSON) works, and which should, technically speaking, fail in the future...?

It looks like there's some unit test code that creates synthetic data with arrays instead of binary - https://github.com/CesiumGS/cesium/blob/10112c84d7de58067b04ccb641772165af4ba81b/Specs/Cesium3DTilesTester.js#L260-L264 - but that's an easy enough fix when we get around to it. I don't think we have any test datasets that use the array form.

If it's later revealed that there are some larger implications we can always revert the change. I _think_ we're ok now though.

javagl:
I'll close this for now, because there is another PR now, where some details are still sorted out. 

(*This* one might be reopened and merged eventually, depending on the outcome there...)

PR #567, opened 2021-11-15
Clarify that asset.tilesetVersion in an external tileset overrides the parent tileset

### LINKED ISSUE #381: tilesetVersion and external tilesets
The spec should be more clear about the interaction between `asset.tilesetVersion` and external tilesets. Do external tilesets always use their parent's `tilesetVersion`? Can an external tileset override it by using a different `tilesetVersion`? Or is this completely implementation specific?

CC https://github.com/AnalyticalGraphicsInc/cesium/pull/8108

ptrgags:
Fixes #381

Clarifies that for external tilesets:

* the external tileset's asset.tilesetVersion overrides that of its parent
* if not specified, the value is inherited from the parent

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -560,6 +560,8 @@ When a tile points to an external tileset, the tile:
 
 ![](figures/tileTransformExternalTileset.png)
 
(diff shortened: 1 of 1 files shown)


lilleyse:
Thanks @ptrgags!

Issue #461, opened 2021-03-24
Adding explanation for URIs in i3dm files to specification


dzungpng:
Currently in the spec, there is a [section ](https://github.com/CesiumGS/3d-tiles/tree/master/specification#uris) to explain URIs in tileset.json. However, there is no such section for URIs in i3dm files. Although this can be figured out by looking at [this code snippet](https://github.com/CesiumGS/cesium/blob/master/Source/Scene/Instanced3DModel3DTileContent.js#L315) in `Instanced3DModel3DTileContent.js`, it would be helpful to document it in the spec as well.

lilleyse:
Fixed in https://github.com/CesiumGS/3d-tiles/pull/568

PR #568, opened 2021-11-15
Add details for URIs in i3dm specification

javagl:
Addresses https://github.com/CesiumGS/3d-tiles/issues/461

The wording is largely taken from the main tileset specification. There, the text was contained in a dedicated `URIs` section, but here, I inserted it directly where the glTF URIs are described - direct preview: https://github.com/javagl/3d-tiles/tree/add-uri-specification-to-i3dm/specification/TileFormats/Instanced3DModel#gltf 

Someone should verify that the statement there is still technically true. Particularly:

- There is no concept of a "data URI" here
- The relative resources are relative to the `i3dm` file


javagl:
@ptrgags As discussed yesterday, this probably fixes #461 when merged

lilleyse:
> There is no concept of a "data URI" here

Yeah, no concept of data URI needed here.

> The relative resources are relative to the `i3dm` file

Yes

lilleyse:
Thanks @javagl!

PR #571, opened 2021-11-15
Add a clarification about scaling geometric error with the `transform` property

### LINKED ISSUE #367: Allow tileset transform to scale geometric error
The tileset transform should be able to scale geometric error. The two areas in the spec that need to be revised are:

In `specification/README.md`
> The transform property does not apply to geometricError—i.e., the scale defined by transform does not scale the geometric error—the geometric error is always defined in meters.

In `tile.schema.json`:
> transform does not apply to geometricError

In CesiumJS, see https://github.com/AnalyticalGraphicsInc/cesium/issues/5330 and https://github.com/AnalyticalGraphicsInc/cesium/pull/7411

ptrgags:
Fixes #367 

This clarifies how transform should be applied to `geometricError`, it _should_ scale the error, though it does so along the largest dimension.

The one tricky bit is how to explain what is the "largest scaling factor" is, technically speaking it would be good to do a [principal component analysis (PCA)](https://www.wikiwand.com/en/Principal_component_analysis) or [Singular value decomposition (SVD)](https://www.wikiwand.com/en/Singular_value_decomposition), albeit this may get too deep into the weeds to be understandable.

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -309,7 +309,7 @@ The `transform` property applies to
 * `tile.boundingVolume`, except when `tile.boundingVolume.region` is defined, which is explicitly in EPSG:4979 coordinates.
 * `tile.viewerRequestVolume`, except when `tile.viewerRequestVolume.region` is defined, which is explicitly in EPSG:4979 coordinates.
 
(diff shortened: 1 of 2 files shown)


lilleyse:
Thanks @ptrgags!

PR #577, opened 2021-11-17
Disallow per-feature property values in JSON header

javagl:
This may fix https://github.com/CesiumGS/3d-tiles/issues/396. It replaces https://github.com/CesiumGS/3d-tiles/pull/566 , where the conflict was resolved in the opposite way. According to a comment there, it should not be possible to store per-feature properties in arrays in the JSON header, and the schema should be updated accordingly. 

But... some things to sort out: 

The schema supported storing arrays in the header. Only the *text* said that these arrays were not allowed to be **per-feature** values. But these arrays may have served the purpose of having a **global** "vector" or "matrix" (represented as an array). Now, the only way to store a global array-typed value is with a `binaryBodyReference`...

@lilleyse This might be an unintended side-effect, and be a limitation that may be hard to justify. 

Or more specifically, in https://github.com/CesiumGS/3d-tiles/pull/566#issuecomment-970429559 you said

> I think the better change would be to remove the second option in https://github.com/CesiumGS/3d-tiles/blob/main/specification/TileFormats/FeatureTable/README.md#json-header.

and I now wondered what "better" means...

---

An aside: The `featureTable.schema.json` contained a definition for a type `numericArray`. While this could be useful (and it could have been used in place of the part of the definition that is now removed), it actually has not been used at all. So I also removed it from the definition here.

### DIFF
--- a/specification/schema/featureTable.schema.json
+++ b/specification/schema/featureTable.schema.json
@@ -75,7 +75,19 @@
                     "$ref": "#/definitions/binaryBodyReference"
                 },
                 {
-                    "type": "number"
+                    "$ref": "#/definitions/globalPropertyBoolean"
+                },
+                {
+                    "$ref": "#/definitions/globalPropertyInteger"
+                },
+                {
+                    "$ref": "#/definitions/globalPropertyNumber"
+                },
+                {
+                    "$ref": "#/definitions/globalPropertyCartesian3"
+                },
+                {
+                    "$ref": "#/definitions/globalPropertyCartesian4"
                 }
             ]
         },
--- a/specification/schema/featureTable.schema.json
+++ b/specification/schema/featureTable.schema.json
@@ -67,24 +67,12 @@
                 "byteOffset"
             ]
         },
-        "numericArray": {
-            "type": "array",
-            "items": {
-                "type": "number"
-            }
-        },
         "property": {
             "title": "Property",
-            "description": "A user-defined property which specifies per-feature application-specific metadata in a tile. Values either can be defined directly in the JSON as an array, or can refer to sections in the binary body with a `BinaryBodyReference` object.",
+            "description": "A user-defined property which specifies application-specific metadata in a tile. Values can refer to sections in the binary body with a `BinaryBodyReference` object. Global values can be also be defined directly in the JSON.",
             "oneOf": [
                 {
                     "$ref": "#/definitions/binaryBodyReference"
-                },
-                {
-                    "type": "array",
-                    "items": {
-                        "type": "number"
-                    }
                 },
                 {
                     "type": "number"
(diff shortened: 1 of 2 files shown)


specification/schema/featureTable.schema.json
```diff
@@ -67,25 +67,13 @@
...
                 },
-                {
-                    "type": "array",
-                    "items": {
-                        "type": "number"
-                    }
-                },
                 {
                     "type": "number"
                 }
```
lilleyse, line 79: > The schema supported storing arrays in the header. Only the *text* said that these arrays were not allowed to be **per-feature** values. But these arrays may have served the purpose of having a **global** "vector" or "matrix" (represented as an array). Now, the only way to store a global array-typed value is with a `binaryBodyReference`...

To avoid this problem should a `property` be one of the existing types?

```json
"oneOf": [
  {
    "$ref": "#/definitions/binaryBodyReference"
  },
  {
    "$ref": "#/definitions/globalPropertyBoolean"
  },
  {
    "$ref": "#/definitions/globalPropertyInteger"
  },
  {
    "$ref": "#/definitions/globalPropertyNumber"
  },
  {
    "$ref": "#/definitions/globalPropertyCartesian3"
  },
  {
    "$ref": "#/definitions/globalPropertyCartesian4"
  }
]
```

lilleyse: There's no `globalPropertyMatrix4` but that could be ok.

javagl: I have to admit that I may not have thoroughly understood some of the design decisions, that led to the current structure, and am now wondering about things like

- why is the `globalPorpertyBoolean` not allowed to be defined with a `byteOffset` (like the `...Integer` and `...Number` counterparts)? (Is this related to the data type size?)
- Why do `globalPropertyCartesian3`/`4` extend `"$ref": "tilesetProperty.schema.json"`? (Is this because they are "structured" (and not _primitive_) types?)

but that may not be utterly important right now.

The last suggestion (of listing the `binaryBodyReference` and the `global...` ones explicitly) could be one way of solving this. Assuming that this does not have undesired side-effects that I'm currently overlooking, I'll update this PR accordingly (early tomorrow) and mark it as ready.

javagl:
@lilleyse I have updated the type definition according to your suggestion.

specification/schema/featureTable.schema.json
lilleyse: > - why is the `globalPropertyBoolean` not allowed to be defined with a `byteOffset` (like the `...Integer` and `...Number` counterparts)? (Is this related to the data type size?)

Most likely because we didn't specify a data type for booleans.

- Why do `globalPropertyCartesian3`/`4` extend `"$ref": "tilesetProperty.schema.json"`? (Is this because they are "structured" (and not _primitive_) types?)

Yes, every object type inherits `tilesetProperty.schema.json` so it can have extras/extensions.

lilleyse: The changes here are good for the known feature table implementations - b3dm, i3dm, and pnts - and with glTF around the corner I don't think we have to focus too much on the smaller details.

PR #549, opened 2021-11-10
3DTILES_implicit_tiling and 3DTILES_multiple_contents revisions

lilleyse:
I did a pass over `3DTILES_implicit_tiling` mainly to tighten up the introduction and remove redundancy in a few areas (including removing the glossary). I also added an example of implicit tiling's interaction with multiple contents.

Since we already have sections in other extensions showcasing `3DTILES_metadata` + `3DTILES_multiple_contents` and `3DTILES_implicit_tiling` + `3DTILES_multiple_contents` I pared down `3DTILES_multiple_contents` to the essentials. I may take another pass over it tomorrow to add back more details.

Previews

* [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/tree/implicit-revisions/extensions/3DTILES_implicit_tiling)
* [3DTILES_multiple_contents](https://github.com/CesiumGS/3d-tiles/tree/implicit-revisions/extensions/3DTILES_multiple_contents)

PR #578, opened 2021-11-17
Add clarifications about metadata groups in multiple contents

ptrgags:
Added a couple of clarifications about metadata groups:

* contents do not have to be added to a group, it's optional
* mention that group metadata is constant across one or more tile contents.

@javagl could you review?

### DIFF
--- a/extensions/3DTILES_multiple_contents/README.md
+++ b/extensions/3DTILES_multiple_contents/README.md
@@ -90,7 +90,7 @@ When this extension is used the tile's `content` property must be omitted.
 
 ### Metadata Groups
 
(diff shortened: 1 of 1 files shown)


lilleyse:
Should this clarification go in the [Content Group Properties](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_metadata#content-group-properties) section of `3DTILES_metadata` given that https://github.com/CesiumGS/3d-tiles/pull/549 trims down the multiple contents spec?

lilleyse:
I think these points are already covered well enough by the [Content Group Properties](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_metadata#content-group-properties) section, but if not it could be a separate PR.

PR #593, opened 2021-11-19
Relax requirements for bounding volumes

ptrgags:
This PR relaxes the requirements for bounding volumes to allow for more future bounding volume types in the future (e.g. [`3DTILES_bounding_volume_s2`](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_bounding_volume_S2) provides S2 bounding volumes). 

This doesn't introduce any breaking changes, but the following relaxations:

* Bounding volumes can now be a `box`, `region` or `sphere` or an extension may define new types
* I relaxed things from "exactly one" bounding volume to "at least one". This allows a tile to have multiple bounding volumes. Clients can choose which one is best to use.

@lilleyse could you review?

### DIFF
--- a/specification/README.md
+++ b/specification/README.md
@@ -910,7 +910,7 @@ Application-specific data.
 <a name="reference-bounding-volume"></a>
 ### Bounding Volume
 
-A bounding volume that encloses a tile or its content. Exactly one `box`, `region`, or `sphere` property is required.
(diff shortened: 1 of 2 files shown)


lilleyse:
Thanks @ptrgags!

PR #594, opened 2021-11-19
Enforce no empty children arrays

ptrgags:
This PR just adds `minItems: 1` to the `tile.children` property to enforce the requirement that leaf tiles may not have an empty `children` array, it's better to leave this undefined to avoid an additional `if` check for the array length.

To be clear, this detail was already in the schema description, this just allows the schema to enforce it automatically.

@lilleyse

### DIFF
--- a/specification/schema/tile.schema.json
+++ b/specification/schema/tile.schema.json
@@ -74,7 +74,8 @@
             "items": {
                 "$ref": "tile.schema.json"
             },
-            "uniqueItems": true
+            "uniqueItems": true,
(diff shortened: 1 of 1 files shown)


lilleyse:
Also this is more consistent with glTF which disallows empty arrays for top-level properties: [glTF.schema.json#L33](https://github.com/KhronosGroup/glTF/blob/24ae1bb305f4be06ce675dfbfcbe8b9b2f0ef671/specification/2.0/schema/glTF.schema.json#L33)

PR #596, opened 2021-11-23
Renamed tile.content to content

javagl:
This was done in https://github.com/CesiumGS/3d-tiles/commit/7379ffde3ccb9e3b5e54e63631f77ea0c9e8fa16 but lost while merging 3d-tiles-next into the main branch.

There are still some plain _textual_ appearances of `tile.content`, but these are used as a "path description inside JSON", and do not explicitly refer to the schema file or a "type name".


lilleyse:
Thanks @javagl!

PR #601, opened 2021-11-29
Implicit tiling schema tweaks

lilleyse:
Fixed up some issues in the implicit tiling schema so that classes can be autogenerated correctly in cesium-native.

* The subtree format no longer references `tilesetProperty.schema.json` from 3D Tiles since this dependency was causing issues in cesium-native, and conceptually they are different file formats. Now `tilesetProperty.schema.json`, `extension.schema.jas`, and `extra.schema.json` have been duplicated for the subtree format. `tilesetProperty.schema.json` is now called `subtreeProperty.schema.json`
* Renamed `tile.3DTILES_implicit_tiling.subtree.schema.json` to `tile.3DTILES_implicit_tiling.subtrees.schema.json` (plural) to match the name of the property. Also changed the title to match. Now there aren't two titles called "Subtree".
* Added descriptions and types to the `constant` enum values.

### DIFF
--- a/extensions/3DTILES_implicit_tiling/schema/subtree/availability.schema.json
+++ b/extensions/3DTILES_implicit_tiling/schema/subtree/availability.schema.json
@@ -6,7 +6,7 @@
     "description": "An object describing the availability of a set of elements.",
     "allOf": [
         {
-            "$ref": "tilesetProperty.schema.json"
+            "$ref": "subtreeProperty.schema.json"
         }
     ],
     "properties": {
--- a/extensions/3DTILES_implicit_tiling/schema/subtree/availability.schema.json
+++ b/extensions/3DTILES_implicit_tiling/schema/subtree/availability.schema.json
@@ -24,10 +24,14 @@
             "description": "Integer indicating whether all of the elements are available (1) or all are unavailable (0).",
             "anyOf": [
                 {
-                    "const": 0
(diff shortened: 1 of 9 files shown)


javagl (approved):
- Differentiating from the `tilesetProperty` makes sense. They _really_ are different file formats, and should be "self-contained" (even though this implies some duplication - _coupling_ them would be worse)
- The renaming seems to be fine - it's one object describing "the subtree**S**", FWIW
- The descriptions ... when I saw `description: "UNAVAILABLE"` I first thought: "Hey, why is there no description available?" 😁 but IIRC, they are essentially used for generating the enum values at https://github.com/CesiumGS/cesium-native/blob/gltf-writer/Cesium3DTiles/generated/include/Cesium3DTiles/Availability.h#L25 , right? In any case, adding the description should be fine.

(The latter might be related to https://github.com/CesiumGS/cesium-native/issues/258 , but I'm not entirely up to date with the latest state of the code generator: I've seen that these are no longer enums, but members of a `struct`, so that issue might already be resolved or have become obsolete, beyond what was done in https://github.com/CesiumGS/cesium-native/pull/350 )

javagl:
So @ptrgags I think this could be merged if there is no further feedback.

ptrgags:
Looks good to me, thanks @lilleyse!

Issue #607, opened 2021-12-10
Avoid duplication in schema - consider a `metadataEntity`


javagl:
The specification currently says

> **Entities** are instantiations of a class, populated with property values conforming to the class definition

Such entities can appear on different levels. Right now, the JSON schema definition for the structure of an entity exists in [`tileset.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/d4116f48e43beb7cf9d7c691100f8bba52e7779b/extensions/3DTILES_metadata/schema/tileset.schema.json#L12), [`tile.3DTILES_metadata.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/d4116f48e43beb7cf9d7c691100f8bba52e7779b/extensions/3DTILES_metadata/schema/tile.3DTILES_metadata.schema.json#L12) and [`group.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/d4116f48e43beb7cf9d7c691100f8bba52e7779b/extensions/3DTILES_metadata/schema/group.schema.json#L12). 

These definitions are structurally equal, and only differ in the title and description. (Except for one [accidental difference](https://user-images.githubusercontent.com/5597569/145638212-663b4767-b9c7-4158-8b96-b280dd2dba66.png) - and that's one argument for avoiding this sort of duplication...). We should consider combining these to something like a `metadataEntity`, to only have the schema definition _once_. I think this could avoid confusion and it could be nicely in line with the textual description. Such a `metadataEntity` could be created by just moving the common `class`/`properties` blocks from the schemas into a common `metadataEntity.schema.json`. 

---

If we decide to do that, then we have to sort out whether the respective objects (i.e. the `tileset`, `tile.3DTILES_metadata`, and `group`) should then **be** a `metadataEntity`, or whether they should **contain** a `metadataEntity`. 

For example, the `tileset.schema.json` currently is
```
{
    "$id": "tileset.schema.json",
    "properties": {
        "class": { ... }
        "properties": { ... }
    }
}
```
And an instance of that could be
```
      "tileset": {
        "class": "Example",
        "properties": {
          "numberOfTiles": "100",
          "date": "2021-03-23"
         }
      }
```

We could let the `tileset` then **be** a `metadataEntity`. This could happen by letting `tileset` be `"allOf": [ { "$ref": "metadataEntity.schema.json" } ]`. In this case, the instance JSON would not change. 

Alternatively, the `tileset` could **contain** a `metadataEntity`, as in
```
[...]

PR #608, opened 2021-12-12
Avoid duplication in schema with metadataEntity

javagl:
A draft that may eventually fix https://github.com/CesiumGS/3d-tiles/issues/607 - further details to be discussed there.

### DIFF
--- a/extensions/3DTILES_metadata/README.md
+++ b/extensions/3DTILES_metadata/README.md
@@ -334,7 +334,7 @@ Most property values are encoded as JSON within the entity to which they are ass
 
 ### Tileset Properties
 
-*Defined in [tileset.schema.json](./schema/tileset.schema.json)*.
+*Defined in [tileset.schema.json](./schema/tileset.schema.json) and [metadataEntity.schema.json](./schema/metadataEntity.schema.json)*.
 
 Properties assigned to tilesets provide metadata about the tileset as a whole. Common examples might include year of collection, author details, or other general context for the tileset contents.
 
@@ -373,7 +373,7 @@ The `tileset` object within a tileset's `3DTILES_metadata` extension must specif
 
 ### Tile Properties
 
-*Defined in [tile.3DTILES_metadata.schema.json](./schema/tile.3DTILES_metadata.schema.json)*.
+*Defined in [tile.3DTILES_metadata.schema.json](./schema/tile.3DTILES_metadata.schema.json) and [metadataEntity.schema.json](./schema/metadataEntity.schema.json)*.
 
 Property values may be assigned to individual tiles, including (for example) spatial hints to optimize traversal algorithms. The example below uses the built-in semantic `TILE_MAXIMUM_HEIGHT` from the [3D Metadata Semantic Reference](../../specification/Metadata/Semantics).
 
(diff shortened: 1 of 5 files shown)


lilleyse:
This change looks good to me. It's essentially equivalent to what glTF does for [material.normalTextureInfo.schema.json](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/material.normalTextureInfo.schema.json) and [material.occlusionTextureInfo.schema.json](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/material.occlusionTextureInfo.schema.json) which "inherit" from `textureInfo.schema.json` and can add their own additional properties. The code generators in cesium-native handle these types.

One difference I noticed is that `material.normalTextureInfo.schema.json` and `material.occlusionTextureInfo.schema.json` declare the properties of the base schema.

```
        "index": { },
        "texCoord": { },
```

Should that be happening here as well?

```
{
    "$schema": "https://json-schema.org/draft/2020-12/schema",
    "$id": "group.schema.json",
    "title": "Group Metadata",
    "type": "object",
    "description": "An object containing metadata about a group.",
    "allOf": [
        {
            "$ref": "metadataEntity.schema.json"
        }
    ],
    "properties": {
        "class": {},
        "properties": {},
        "extensions": {},
        "extras": {}
    }
}
```

javagl:
> Should that be happening here as well?

This could be done here, and maybe it _should_ be done, but only for consistency. Long(er)-term, these properties may be omitted when they are already mentioned in the `allOf` base schema. (This might eventually also be done in glTF, as of https://github.com/KhronosGroup/glTF/issues/2062#issuecomment-934201170 , but probably only after a few other updates).

javagl:
I could mark this as "ready for review", but before merging, we should probably check whether the spec text should be updated - for example, places that talk about a `group` might have to be reworded, from

> This-and-that is a group, see the `[group.schema.json]`

to 

> A group is a metadata entity, see `[metadataEntity.schema.json]`

lilleyse:
Yes, do a pass over the spec text and then bump me for review.

javagl:
I've added the links to the schema where appropriate, and did a minor wording update. 

I stumbled over the fact that the word 'entity' was apparently used with a meaning that is different from the definition in the https://github.com/CesiumGS/3d-tiles/tree/main/specification/Metadata#concepts section:

> Entities are instantiations of a class, populated with property values conforming to the class definition.

The wording here said

> ... properties do not take on particular values until a metadata is assigned (i.e. the class is "instatiated") to a particular entity within the 3D Tiles hierarchy

Specifically:

- "Entity" once means "that thing that is created from the metadata values" (i.e. the instance of the metadata class)
- "entity" can also be the object that this instance is _associated with_ (e.g. the group or tile)

Maybe the wording would be less confusing if it was

> ... properties do not take on particular values until a metadata is assigned (i.e. the class is "instatiated") to a particular <strike>entity within</strike> **element of** the 3D Tiles hierarchy

but maybe some of these ambiguities are inevietable, due to the genericity of the terms. I hope that the updated wording is OK, though...

lilleyse:
Thanks @javagl. The edits looks good to me.

PR #615, opened 2022-01-15
Clarify meaning of "internal buffer"

javagl:
The text that referred to the "internal buffer" should have been indented as being part of the example. But the term "internal buffer" was not explicitly defined before that, so added this accordingly. 

Preview: https://github.com/javagl/3d-tiles/tree/internal-subtree-buffer-wording/extensions/3DTILES_implicit_tiling#buffers-and-buffer-views


lilleyse:
Thanks @javagl

PR #619, opened 2022-01-24
Make `properties` optional in schema

lilleyse:
The `properties` field for tile metadata and subtree metadata should not be marked as required because there are valid cases where all properties may be omitted.

For example... a class with two optional properties:

```json
"classes": {
  "tile": {
    "properties": {
      "name": {
        "componentType": "STRING",
        "required": false
      },
      "id": {
        "componentType": "STRING",
        "required": false
      }
    }
  }
}
```

Some tiles may provide values, but other may not

```json
    "children": [
      {
        "extensions": {
          "3DTILES_metadata": {
            "class": "tile",
            "properties": {
              "name": "Tile A",
              "id": 0
            }
          }
        }
      },
      {
        "extensions": {
          "3DTILES_metadata": {
            "class": "tile"
          }
        }
      }
    ]
```

I opened a similar PR for [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features): https://github.com/CesiumGS/glTF/pull/49

### DIFF
--- a/extensions/3DTILES_metadata/schema/3DTILES_implicit_tiling/subtree.3DTILES_metadata.schema.json
+++ b/extensions/3DTILES_metadata/schema/3DTILES_implicit_tiling/subtree.3DTILES_metadata.schema.json
@@ -26,7 +26,6 @@
         "extras": {}
     },
     "required": [
-        "class",
-        "properties"
+        "class"
     ]
 }
--- a/extensions/3DTILES_metadata/schema/metadataEntity.schema.json
+++ b/extensions/3DTILES_metadata/schema/metadataEntity.schema.json
@@ -54,7 +54,6 @@
         "extras": {}
     },
     "required": [
-        "class",
-        "properties"
+        "class"
     ]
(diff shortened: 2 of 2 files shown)


javagl (commented):
I first thought "Of course they can be made optional, why not?"

The problem: I _really_ tried to find an answer to that 'why not?' part (i.e. I might just be overthinking here). 

- One could argue that each _entry_ of the dictionary can be optional (when the correponding property is not marked as `required`), but not the dictionary _itself_ (and one could make a case for storing an _empty_ dictionary in the example)
- One could say that when there are no properties for that tile, then the enclosing `3DTILES_metadata` should not exist in the first place

I'm not opposed to this, but will try to think about the reason for the "empty" `3DTILES_metadata` object, or whether making the properties optional can have undesired side-effects. (So I might hit 'merge' soon, if I don't come up with a _specific_ reason to not do this...)

Issue #617, opened 2022-01-24
Summary of possible changes for the specification of `3DTILES_implicit_tiling`


javagl:
There are several smaller possible changes for the `3DTILES_implicit_tiling` specification. This is a short summary of these, with cross-links to existing issues, or summaries that may be used to start issues when there is the need for further disussion. 

### Meaning of "Internal buffer"

The specification referred to the "internal buffer" of a subtree file, without explicitly saying what this is. This is clarified via the (pending) https://github.com/CesiumGS/3d-tiles/pull/615. 

### Handling of leftover bits in availability bitstreams

<sup>Via slack</sup>

The specification currently does not say anything about the unused bits of availability bitstreams. Specifically: The value of the last 3 or 7 bits of the last byte are not determined by the availability. The specification should _explicitly require_ these bits to be set to 0. This makes sure that two binary files that carry the same availability information are _bitwise equal_. 

(This could be trivial to fix, but should be done after #615 is merged, assuming that it should be mentioned in the same section)

### Subtree JSON format

(Details to be discussed)

### Upsampling requirements based on availability information

A rough summary of this question is: What should a client do when there is an implicit quadtree tile that is neither a leaf _nor_ a tile that has 4 children?

A rough summary of the possibilities seem to be to show the child tiles that are available and...

- omit the child tiles that are not available and show "holes" there
- or upsample the geometry from the tile to fill the holes that are caused by unavailable children

Alternatively, there could be stricter requirements for the availablility (basically disallowing the case of only having 1,2,3 children). Further details can be found at https://github.com/CesiumGS/3d-tiles/issues/489#issuecomment-1011472667 .

### Consistent naming for "number of levels" in trees

<sup>Via slack</sup>

The "size of a tree" is represented with different terms:

- `maximumLevel` for the whole implicit tree, where `maximumLevel+1` yields the number of levels
- `subtreeLevels` for the subtrees, where `subtreeLevels-1` yields the maximum level in a subtree

For consistency, the terms could be

- `maximumLevel` and `maximumSubtreeLevel` or
- `availableLevels` and `subtreeLevels` (this seemed to be the preferred solution)

This will affect only small parts of the spec, **but** will affect all existing data sets, and the implementations. (At least, if one wants the variable names to match the properties - i.e. it may imply a "Replace all `maxmimumLevel` with `availableLevels`", with a diligent review to get the `+1`'s right...)

### Fold extensions into implicit tiling spec

[...]

PR #620, opened 2022-01-28
Added note about zeroing unused bits in bitstreams

javagl:
This was brought up via slack and mentioned in https://github.com/CesiumGS/3d-tiles/issues/617 . It originally (only) referred to the [availability bitstreams of `3DTILES_implicit_tiling`](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_implicit_tiling#availability-packing), but this section mainly links to the "Storage Formats --> Binary Table Format --> Booleans" section of the Metadata spec, so we thought that it could make sense to add this as a general requirement there:

https://github.com/CesiumGS/3d-tiles/blob/c0d63ec8cf36b4b5d75c0ac031ea9bc4d5a53ae9/specification/Metadata/README.md#booleans

(This ensures that two "data sets" that describe "the same data" are _bitwise equal_, by making sure that there are no random, unspecified bits contained in the data)


lilleyse:
Looks good, thanks @javagl

Issue #621, opened 2022-02-02
BATCH_LENGTH in batch table


martin-schilling:
Hi, thanks for your amazing work on Cesium! 

I noticed a small thing in the [reference card](https://github.com/CesiumGS/3d-tiles/blob/main/3d-tiles-reference-card.pdf) that I found weird. On page 11 there are examples regarding the basic structure of the feature and batch table headers. The JSON of the batch table header includes a `BATCH_LENGTH` property which, at least as far as I can tell, is not supposed to be there. Based on the following statement on page 12 the `BATCH_LENGTH` property should instead be part of the feature table:

>The Feature Table of Batched 3D Models only contains global properties. The size of the Batch Table is given as the BATCH_LENGTH property, and the RTC_CENTER may store the center point in the case that the positions are given relative to the center.

I came across this issue when I was trying to load a tileset that works perfectly fine with the normal model using the new experimental model today. I simply followed the reference card when I created it and it looks like I also put the `BATCH_LENGTH` into the batch table header by accident because of that. 

In case I missed something, I would appreciate a brief clarification. Thank you!

javagl:
@martin-schilling That indeed seems to be a bug. The BatchTable does not define or allow any global semantics, as of https://github.com/CesiumGS/3d-tiles/blob/main/specification/schema/batchTable.schema.json#L92 

I addressed that with https://github.com/CesiumGS/3d-tiles/pull/622

martin-schilling:
@javagl great thanks for fixing that so quickly

PR #622, opened 2022-02-02
Removed wrong BATCH_LENGTH in reference card

javagl:
Fixes https://github.com/CesiumGS/3d-tiles/issues/621

Page 11 of the reference card contained a JSON snippet and a statement that suggested that `BATCH_LENGTH` was a global property in the BatchTable. This is not the case: The `BATCH_LENGTH` is a global property of the FeatureTable.

(Also, the text did not mention that the data can be stored in arrays. Now the text mentions it shortly. If desired, it could be elaborated further).

A quick "visual diff" from a screenshot showing the old (top) and new (bottom) version:

Cesium Reference Card Fix Issue 621

### DIFF
--- a/reference-card/3d-tiles-reference-card-page0011.svg
+++ b/reference-card/3d-tiles-reference-card-page0011.svg
@@ -2,23 +2,23 @@
 <!-- Created with Inkscape (http://www.inkscape.org/) -->
 
 <svg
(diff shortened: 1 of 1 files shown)


lilleyse:
Thanks @javagl!

PR #623, opened 2022-02-08
Change maximumLevel to availableLevels in implicit tiling extension

lilleyse:
This is targeting a staging branch where we'll be pushing revisions to the 3D Tiles Next extensions during the next couple of weeks as we prepare for 3D Tiles 1.1. The staging branch will be merged into main once everything stabilzes.

Changes `maximumLevel` to `availableLevels` for consistency with `subtreeLevels`. Note that `maximumLevel` is an index whereas `availableLevels` is a length, so `availableLevels` is equivalent to `maximumLevel + 1`.

See discussion in https://github.com/CesiumGS/3d-tiles/issues/617


lilleyse:
I'm going to merge this into the staging branch where the full set of subtree changes can be reviewed together later.

PR #628, opened 2022-02-15
Mention restrictions to numeric component types in schema

javagl:
The `description` in the property schema for `min/max/offset/scale` now mentions that these are only applicable when the `componentType` is numeric.


javagl:
This is obsolete as of https://github.com/CesiumGS/glTF/pull/53#issuecomment-1040431925

Issue #504, opened 2021-10-13
Add a property to specify the CRS of the tileset


lilleyse:
A long running source of confusion is that 3D Tiles and glTF have different coordinate systems. 3D Tiles uses a generic right-handed cartesian coordinate system where the z-axis can mean different things depending on whether the tileset is in a global or local coordinate system, while glTF is strictly y-up.

This breaks down quickly once you're working with geospatial data that doesn't have an up-axis. This tile here is glTF but it doesn't _really_ have an up axis like a regular game asset would.

<img src="https://user-images.githubusercontent.com/915398/137203709-e9d3c1cd-b52b-481b-a162-d9cae2504ea9.png" width="400" />

[<sub>Sandcastle</sub>](https://sandcastle.cesium.com/#c=vVNhS8MwEP0roZ8mjHSiIMw61FlEmJvo9Iv1Q9aeWzBLSnLdVsX/7nXJpsx91VJo3rv37u6VdiEsW0hYgmVnTMOS9cHJas6f1lwri/I17huNQmqwWXRwmulMew93OWjgU2UmwAsocTYGhxdTkjocg7V0oL5oK/AummZFIStH7GHHX7xz6ivUC3f36AuLdBL6qBWYW4Eznt493AxGw5M269B9sG2+IndYjZpJlOC4KIrWR6YZK42qFYXosjVsCEcSo12XPXuGhSXaG7h/FZ8hDA/al/BcygJnXXYcoLD5uC5pZuhy4TEfjoZpkMwFTZRCbTV9o4zl9+lVU/8k1ed3wvp/ElKybcg/S3h9n6bDPRnf/y3jNuafZbwcPKa7EUO0FYIuNp+1Jz240a6EHI29lSupG1PUjhKHtYLeJtK5nJfGIqusanEeI8xLRRu4eFLlb4A8d64xNtIk/mlNCrlgsjjb82uzXAnnqPJaKfUg3yGLeklM+l9WZeit6eloAVaJupHNDnsDT3LOk5jgficaoybC7nT+Ag)

[...]

PR #633, opened 2022-02-21
Add `TILESET_CRS_GEOCENTRIC` semantic

lilleyse:
Related to https://github.com/CesiumGS/3d-tiles/issues/504.

This PR adds a new tileset metadata semantic called `TILESET_CRS_GEOCENTRIC` that allows a tileset to specify its geocentric coordinate reference system. This does not go as far as allowing arbitrary CRS support in 3D Tiles. Values are limited to geocentric (Cartesian) coordinate systems like WGS 84 and its realizations, as well as [ITRF](https://en.wikipedia.org/wiki/International_Terrestrial_Reference_System_and_Frame) realizations. The metadata is informational only and should not require runtime changes.

Known values include:<ul><li>`"EPSG:4978"` - WGS 84</li><li>`"EPSG:7656"` - WGS 84 (G730)</li><li>`"EPSG:7658"` - WGS 84 (G873)</li><li>`"EPSG:7660"` - WGS 84 (G1150)</li><li>`"EPSG:7662"` - WGS 84 (G1674)</li><li>`"EPSG:7664"` - WGS 84 (G1762)</li><li>`"EPSG:9753"` - WGS 84 (G2139)</li></ul>

The [Coordinate Reference System](https://github.com/CesiumGS/3d-tiles/tree/main/specification#coordinate-reference-system-crs
) section in the spec still applies here

> 3D Tiles uses a right-handed Cartesian coordinate system; that is, the cross product of _x_ and _y_ yields _z_. 3D Tiles defines the _z_ axis as up for local Cartesian coordinate systems. A tileset's global coordinate system will often be in a [WGS 84](http://earth-info.nga.mil/GandG/publications/tr8350.2/wgs84fin.pdf) earth-centered, earth-fixed (ECEF) reference frame ([EPSG 4978](http://spatialreference.org/ref/epsg/4978/)), but it doesn't have to be, e.g., a power plant may be defined fully in its local coordinate system for use with a modeling tool without a geospatial context.

Some of the open questions are:

* Does the CRS apply to region bounding volume types? Yes, it should. What about for CRSs besides WGS 84?
* Does the CRS apply to glTF content? Yes, but only after the glTF up-axis conversion and tile transforms have been applied. E.g. a glTF could be defined in local coordinates and then transformed into WGS 84 with a tile transform.
* What other tileset metadata should be included?

[Preview](https://github.com/CesiumGS/3d-tiles/tree/tileset-crs-semantic/specification/Metadata/Semantics#tileset-semantics)

To do:

* [x] Add epoch


specification/Metadata/Semantics/README.md
```diff
@@ -48,6 +48,8 @@ Units for all linear distances are meters, and all angles are radians.
...
 --|--|--
 `TILESET_FEATURE_ID_SETS`|<ul><li>Type: `STRING`</li><li>Array: `true`</li></ul>|The union of all the feature ID sets in glTF content using the [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_mesh_features) and [`EXT_instance_features`](https://github.com/CesiumGS/glTF/tree/3d-tiles-next/extensions/2.0/Vendor/EXT_instance_features) extensions.
+`TILESET_CRS_GEOCENTRIC`|<ul><li>Type: `STRING`</li></ul>|The geocentric coordinate reference system (CRS) of the tileset. Known values include:<ul><li>`"EPSG:4978"` - WGS 84</li><li>`"EPSG:7656"` - WGS 84 (G730)</li><li>`"EPSG:7658"` - WGS 84 (G873)</li><li>`"EPSG:7660"` - WGS 84 (G1150)</li><li>`"EPSG:7662"` - WGS 84 (G1674)</li><li>`"EPSG:7664"` - WGS 84 (G1762)</li><li>`"EPSG:9753"` - WGS 84 (G2139)</li></ul>`region` bounding volumes are assumed to use the same reference ellipsoid as the geocentric coordinate reference system specified here.<br><br>For more details on coordinate reference systems in 3D Tiles, see [Coordinate Reference System (CRS)](../../#coordinate-reference-system-crs).
```
ptrgags, line 51: "Known values include" sounds a bit ambiguous to me (maybe that's just me :sweat_smile: ). Does this mean "The following values are allowed:" or "Known values include, but are not limited to:"?

specification/Metadata/Semantics/README.md
```diff
@@ -48,6 +48,8 @@ Units for all linear distances are meters, and all angles are radians.
...
+`TILESET_CRS_COORDINATE_EPOCH`|<ul><li>Type: `SCALAR`</li><li>Component Type: `FLOAT32` or `FLOAT64`</li></ul>|The coordinate epoch for coordinates that are referenced to a dynamic CRS such as WGS 84. Coordinates include glTF vertex positions after transforms have been applied — see [glTF transforms](https://github.com/CesiumGS/3d-tiles/tree/main/specification#gltf-transforms). Expressed as a decimal year (e.g. `2019.81`). See [WKT representation of coordinate epoch and coordinate metadata](http://docs.opengeospatial.org/is/18-010r7/18-010r7.html#128) for more details.
```
ptrgags, line 52: Why is this a float and not a string or a VEC2 of UINTN? The WKT spec describes it as `<unsigned integer> . <unsigned integer>`. A float value would result in a loss of precision.

The example of `2019.81` is actually a case where this is not exactly representable:

image

specification/Metadata/Semantics/README.md
ptrgags: Also, is this future proof for future epochs of WGS 84?

lilleyse: Yes, with this wording it should be future proof for new WGS 84 realizations.

specification/Metadata/Semantics/README.md
lilleyse: Good point. `STRING` seems like a better choice here.

lilleyse:
@ptrgags updated!

ptrgags:
@lilleyse changes look good, thank you!

PR #630, opened 2022-02-18
Revisions to the 3D Tiles Next extensions

lilleyse:
This PR includes revisions to many of the 3D Tiles Next extensions based on feedback since November.

* Arrays are now distinct from the type system: [#595](https://github.com/CesiumGS/3d-tiles/issues/595)
* Added content metadata and subtree metadata to round out the granularities at which metadata can be provided
* Added support for attribution (aka "credits" or "copyrights"). Attribution can be assigned at any granularity - including per-vertex or per-texel granularity - but will typically be assigned to tile contents or the tileset as a whole. [#55](https://github.com/CesiumGS/3d-tiles/issues/55)
* Added `offset` and `scale` to remap quantized metadata to different ranges
* Added back `default` values for when property values are missing
* Changed `maximumLevel` to `availableLevels` for consistency with `subtreeLevels`
* Other schema changes. The full revision history can be found [here](https://github.com/CesiumGS/3d-tiles/blob/extension-revisions/next/REVISION_HISTORY.md).

There have also been changes to the glTF extensions that are closely associated with 3D Tiles Next. The `EXT_mesh_features` extension is now split into three separate extensions:
  * `EXT_mesh_features` that only defines the concept of feature IDs
  * `EXT_structural_metadata` that allows associating the features with metadata
  * `EXT_instance_features` that supports GPU instance feature IDs.

For the CesiumJS implementation of these changes, see https://github.com/CesiumGS/cesium/tree/extension-revisions

#

Direct links to specs:

* [3D Metadata Specification](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/specification/Metadata)
* [3D Metadata Semantics Reference](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/specification/Metadata/Semantics)
* [3DTILES_metadata](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/extensions/3DTILES_metadata)
* [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/extensions/3DTILES_implicit_tiling)
* [3DTILES_multiple_contents](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/extensions/3DTILES_multiple_contents)
* [EXT_mesh_features](https://github.com/CesiumGS/glTF/tree/ext-mesh-features-revision/extensions/2.0/Vendor/EXT_mesh_features)
[...]

PR #634, opened 2022-02-22
Initial draft for 3D Tiles 1.1 schema

lilleyse:
This PR folds all the 3D Tiles Next extension from https://github.com/CesiumGS/3d-tiles/pull/630 into the 3D Tiles 1.1 tileset.json schema. There are no breaking changes but there are some deprecations.

Opening for early feedback.


specification/schema/boundingVolume.schema.json
```diff
@@ -38,7 +38,15 @@
             "minItems": 4,
             "maxItems": 4
         },
+        "s2": {
+            "description": "S2 cell",
+            "allOf": [
+                {
+                    "$ref": "s2.schema.json"
```
lilleyse, line 45: Should box, region, sphere also have dedicated schema files?

specification/schema/boundingVolume.schema.json
```diff
@@ -38,7 +38,15 @@
             "minItems": 4,
             "maxItems": 4
         },
+        "s2": {
```
lilleyse, line 41: Should this be called `s2` or `s2Cell`?

specification/schema/tile.schema.json
```diff
@@ -66,7 +66,32 @@
...
+            "deprecated": true
+        },
+        "contents": {
+            "type": "array",
+            "description": "An array of contents.",
+            "items": {
+                "$ref": "content.schema.json"
+            },
+            "minItems": 1
+        },
```
lilleyse, line 79: Unfortunately we can't change `content` to be an array because that would be a breaking change. And it would be awkward to allow `content` to be either an object or an array of objects.

specification/schema/tileset.schema.json
```diff
@@ -19,8 +19,54 @@
...
+        },
+        "groups": {
+            "type": "object",
+            "description": "A dictionary, where each key is a group ID and each value is an object defining the group. Group IDs may contain only alphanumeric and underscore characters",
+            "minProperties": 1,
+            "additionalProperties": {
+                "$ref": "group.schema.json"
             }
         },
+        "metadata": {
```
lilleyse, line 54: Most of the properties in `3DTILES_metadata` have been hoisted up to root scope. There are a few reasons for doing that:

* Schemas are so central to 3D Tiles 1.1 that it feels right to make it a top level property
* `statistics` is a replacement for `properties` which was also at the top level
* `groups` are really a distinct concept from metadata and don't belong under a metadata sub-object. Arguably `3DTILES_metadata` should not have groups; that should be a separate extension.
* Tileset, tile, and content all have a `metadata` property that is a `metadataEntity.schema.json`. The symmetry is nice.

specification/schema/tileset.schema.json
```diff
@@ -19,8 +19,54 @@
...
+            "allOf": [
+                {
+                    "$ref": "metadataEntity.schema.json"
+                }
+            ],
+            "description": "Tileset metadata."            
+        },
+        "contentInfo": {
```
lilleyse, line 62: `contentInfo` is similar `3DTILES_content_gltf`. The sub-object `contentInfo.gltf` says what glTF extensions are used/required. In the future it might say what profile the glTF contents conform - for example a point cloud profile. 

More sub-objects could be added for other content formats in the future.

lilleyse: I'm not really happy with the name `contentInfo`. Suggestions welcome...

lilleyse (commented):
Left some inline comments to start discussion on a few things

specification/schema/tile.schema.json
j9liu: Should we enforce that it should be only either `content` or `contents`, i.e. put these in a `"oneOf"`?

lilleyse: Yes! I pushed a fix for that in https://github.com/CesiumGS/3d-tiles/pull/634/commits/ee3514cf9de0469cd61d71bf6d72c0dcfc92a665. If my understanding of JSON schema is correct this snippet below should allow just `content`, just `contents`, or neither.

```json
"oneOf": [
  {
    "required": [
      "content"
    ]
  },
  {
    "required": [
      "contents"
    ]
  },
  {}
]
```

specification/schema/boundingVolume.schema.json
lilleyse: Changed to `s2Cell`

lilleyse:
@j9liu @ptrgags 
* `s2` was renamed to `s2Cell`
* `content` and `contents` are now mutually exclusive.

Any thoughts about `contentInfo`?

specification/schema/tileset.schema.json
lilleyse: I removed `contentInfo` for now. Will open up a dedicated issue to discuss this some more.

specification/schema/boundingVolume.schema.json
lilleyse: Not addressed yet, but doesn't need to hold up this PR.

lilleyse:
This was helpful for figuring out where the 3D Tiles Next capabilities would fit into the tileset structure but can be closed now.

javagl:
Just to have the pointer: Some further considerations are at https://github.com/CesiumGS/3d-tiles/pull/654

Issue #638, opened 2022-02-23
Clarification for `offset` and `scale` of metadata properties


javagl:
<sub>Updated based on [the comment below](https://github.com/CesiumGS/3d-tiles/issues/638#issuecomment-1050282668)</sub>

The [section that defines `offset` and `scale`](https://github.com/CesiumGS/3d-tiles/tree/18a0ae693a963ca322c340a39e3844bcfeb56c10/specification/Metadata#offset-and-scale) shows the computation of the actual property value, based on the "raw" value, and mentions how this concept is applied to different types (e.g. vector- or array types), but some further clarification for this part - and analogously, the part that defines `minimum` and `maximum` - may be necessary.

The points that should be made clearer:

- The `offset` and `scale` properties are only applicable when the property either has a floating-point `componentType`, or when `normalized` is set to `true`.
- An implementation note should point out the possible lossiness of the conversion (beyond what is already said in the "Normalized Values" section)
- The structure of the `offset` and `scale`, must match the `type` of the property, but their components are treated as floating point values. 
- The structure of `minimum` and `maximum` must match the `type` of the property. If the `normalized` flag is set to `true`, then their components will be floating point values. Otherwise, their components will match the `componentType` of the property. 

---

Note: I had created some [example computations for the `EXT_structural_metadata` specification](https://github.com/javagl/glTF/blob/fc0fe024228ee212f85ec9034f8452c518d6ac64/extensions/2.0/Vendor/EXT_structural_metadata/README.md#linear-value-transformations) that show the use of `offset` and `scale` for some example types, but the JSON-specific part is not applicable here.

---

Note: There are no _explicit_ statements about overflows that may appear when computing `result = offset + scale * normalize(value)`. The implementation note in the "Normalized Values" section mentions that some bits may be lost under certain conditions when `normalized=true`. An implementation note in the updated section emphasizes that the computations should be done in floating-point, and should retain as much precision as possible. Stricter requirements can hardly be stated here, considering that there may be implementations of this specification that focus on the GPU, where usually, only 32 bit floating point values are available. 

---

The preview of the updated section can now be found in https://github.com/CesiumGS/3d-tiles/pull/640

lilleyse:
> For properties that are `normalized` or define an `offset` or `scale`, the component type of `minimum` and `maximum` is assumed to be a floating point type, and their values represent the the bounds of the property values _after_ normalization or `offset`- or `scale` computations have been applied. 
>
[...]

PR #640, opened 2022-02-26
Added clarification for offset and scale

javagl:
This PR corresponds to the changes described in https://github.com/CesiumGS/3d-tiles/issues/638 

Preview: https://github.com/javagl/3d-tiles/tree/extension-revisions-offset-scale/specification/Metadata#offset-and-scale

### DIFF
--- a/specification/Metadata/README.md
+++ b/specification/Metadata/README.md
@@ -310,16 +310,7 @@ Normalized properties (`normalized`) provide a compact alternative to larger flo
 
 A property may declare an offset (`offset`) and scale (`scale`) to apply to property values. This is useful when mapping property values to a different range. 
 
-The `offset` and `scale` can be defined for types that either have a floating-point `componentType`, or when `normalized` is set to `true`. This applies to `SCALAR`, `VECN`, and `MATN` types, and to fixed-length arrays of these types. The structure of `offset` and `scale` must match the `type` of the property:
-
-- For `SCALAR` (non-array) types, they are single numeric values.
-- For `SCALAR` array types with fixed length `count`, they are arrays with length `count`.
-- For `VECN` types, they are arrays, with length `N`.
-- For `MATN` types, they are arrays, with length `N * N*`.
-- For `VECN` array types with fixed length `count`, they are arrays with length `count`, where each array element is itself an array of length `N`
-- For `MATN` array types with fixed length `count`, they are arrays with length `count`, where each array element is itself an array of length `N * N`.
-
-The components of `offset` and `scale` are floating-point values. 
+The `offset` and `scale` can be defined for types that either have a floating-point `componentType`, or when `normalized` is set to `true`. This applies to `SCALAR`, `VECN`, and `MATN` types, and to fixed-length arrays of these types. The structure of `offset` and `scale` is explained in the [Property Values Structure](#property-values-structure) section.
 
 The following equation is used to transform the original property value into the actual value that is used by the client:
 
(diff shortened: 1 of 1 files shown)


specification/Metadata/README.md
```diff
@@ -308,25 +308,38 @@ Normalized properties (`normalized`) provide a compact alternative to larger flo
...
 
 #### Minimum and Maximum Values
 
-Properties representing numeric values — such scalar, vector, and matrix types — may specify a minimum (`minimum`) and maximum (`maximum`). Minimum and maximum values represent component-wise bounds of the valid range for a property.
+Properties may specify a minimum (`minimum`) and maximum (`maximum`) value. Minimum and maximum values represent component-wise bounds of the valid range of values for a property. The `minimum` and `maximum` value can be defined for `SCALAR`, `VECN`, and `MATN` types with numeric component types, and for fixed-length arrays of these types. The structure of `minimum` and `maximum` must match the `type` of the property, as described in the [section about Offset and Scale](#offset-and-scale). 
+
+For properties that are `normalized`, the component type of `minimum` and `maximum` is a floating point type, and their values represent the the bounds of the property values _after_ normalization or `offset`- or `scale` computations have been applied.
```
lilleyse, line 340: ... or after other transformations have been applied that aren't part of the class definition, e.g. `offset`/`scale` in the property table. I think it's worth making this point because a class definition like the one below doesn't seem valid but it is

```json
{
  "type": "SCALAR",
  "componentType": "UINT8",
  "normalized": true,
  "min": -5.98,
  "max": 6.23
}
```

lilleyse, line 340: > The structure of `minimum` and `maximum` must match the `type` of the property, as described in the [section about Offset and Scale](#offset-and-scale). 

Do the `noData` and `default` sections need similar treatment?

Should there be a common section that `offset`, `scale`, `min`, `max`, `noData`, `default` reference?

lilleyse: Thinking more about `noData` and `default`...

It makes sense that `normalized`, `offset`, and `scale` have no effect on the `noData` value since the `noData` should be the "packed" type.

I think `default` should be the "unpacked" type. Currently it's not specified very clearly.

specification/Metadata/README.md
```diff
@@ -308,25 +308,38 @@ Normalized properties (`normalized`) provide a compact alternative to larger flo
...
+
+When the `offset` for a property is not given, then is is assumed to be `0` for each component of the respective type. When the `scale` value of a property is not given, then it is assumed to be `1` for each component of the respective type. 
 
 #### Minimum and Maximum Values
 
-Properties representing numeric values — such scalar, vector, and matrix types — may specify a minimum (`minimum`) and maximum (`maximum`). Minimum and maximum values represent component-wise bounds of the valid range for a property.
+Properties may specify a minimum (`minimum`) and maximum (`maximum`) value. Minimum and maximum values represent component-wise bounds of the valid range of values for a property. The `minimum` and `maximum` value can be defined for `SCALAR`, `VECN`, and `MATN` types with numeric component types, and for fixed-length arrays of these types. The structure of `minimum` and `maximum` must match the `type` of the property, as described in the [section about Offset and Scale](#offset-and-scale).
```
lilleyse, line 338: Unrelated question - should the shorthand names in the parentheses (`minimum`), (`maximum`), and (`defaultValue`) actually be (`min`), (`max`), and (`default`) to match the json schema more closely?

lilleyse (commented):
Thanks @javagl. After this PR is merged can you update the relevant places in the 3D Tiles and glTF specs/schemas? This affects statistics as well, those should be the values after `normalized`, `scale`, `offset` and storage-specific transformations have been applied.

lilleyse:
The restructuring looks good. Thanks!

lilleyse:
> Unrelated question - should the shorthand names in the parentheses (`minimum`), (`maximum`), and (`defaultValue`) actually be (`min`), (`max`), and (`default`) to match the json schema more closely?

I pushed this change in https://github.com/CesiumGS/3d-tiles/commit/aa724ac0f5a70e30f54facf943033c1df22f4608

Issue #641, opened 2022-03-01
Metadata inheritance


lilleyse:
3D Tiles Next defines different levels of metadata granularity but it doesn't define the relationship or inheritance model of this metadata. There are a lot of ways this could be done.

<img src="https://user-images.githubusercontent.com/915398/156205230-54e8fb74-76b2-4b7f-b192-cebf09a479c7.png" width="400" />

First, metadata could be inherited based on the tileset structure — a feature would inherit its content's metadata, which would inherit its tile's metadata, which would inherit its tileset's metadata. This is a "BELONGS-TO" rather than "IS-A" relationship. A feature is not a tileset, but it belongs to a tileset and has access to the tileset's metadata.

This is how it's implemented in CesiumJS in [getPropertyInherited](https://github.com/CesiumGS/cesium/blob/9b93f47d43e95ec7dc610f7da45d4fc049a6ea90/Source/Scene/Cesium3DTileFeature.js#L263).

One aspect that needs further discussion is whether a tile should inherit its parent tile's metadata. It seems like right thing to do but the implementation may not be straightforward or performant.

Metadata could also be inherited in a more abstract sense. [Batch Table Hierarchy](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_batch_table_hierarchy) allows a metadata entity to belong to other metadata entities and get access to their properties. In `EXT_structural_metadata` terms this would be a row in a property table pointing to a row in another property table. This is also a "BELONGS-TO" relationship, but more abstract since it doesn't use the tileset structure. Arguably content groups are similar.

Then there is the "IS-A" relationship which implies class hierarchies. Unlike the previous approaches, the class hierarchy would need to be defined in the class definition. This has some benefits but it isn't as good for composability.

I've left out some details here to keep this brief to start. What metadata inheritance model makes the most sense for 3D Tiles?

lilleyse:
Also CC https://github.com/CesiumGS/3d-tiles/issues/558. This issue is essentially a superset of that.

javagl:
One question is on which level this inheritance should be defined or specified. 

---

Right now, the [3D Metadata Specification](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Metadata) has no connection to tilesets, or any sort of hierarchically structured data. So any form of inheritance that was supposed to be defined on this level would have to be defined solely on the level of the concepts that are contained in this specification - and this is essentially on the level of Classes and Entities. 

[...]

PR #642, opened 2022-03-02
Property table array in subtree format

lilleyse:
Instead of inlining a property table in `tileMetadata` and `contentMetadata` there is now a dedicated section for property tables. This is consistent with [EXT_structural_metadata](https://github.com/CesiumGS/glTF/tree/ext-mesh-features-revision/extensions/2.0/Vendor/EXT_structural_metadata) and is more future proof if we ever need to support property tables referring to other property tables for metadata inheritance (https://github.com/CesiumGS/3d-tiles/issues/641).


extensions/3DTILES_implicit_tiling/schema/subtree/contentMetadata.schema.json
```diff
@@ -0,0 +1,24 @@
+{
```
ptrgags, line 1: Do we forsee that `contentMetadata` and tileMetadata` will diverge? the schema files are pretty much identical. Might be okay, just thought I'd ask.

extensions/3DTILES_implicit_tiling/README.md
```diff
@@ -387,7 +387,7 @@ Subtree metadata (`subtreeMetadata`) is encoded in JSON according to the [JSON F
 <!-- omit in toc -->
 #### Property Tables
 
-Binary property values are stored in a **property table**. A property table must specify its class (`class`), which refers to a class ID in the `3DTILES_metadata` extension of root tileset JSON, and a dictionary of properties (`properties`), where each key is a property ID correspond to a class property and each value is the index of the buffer view containing property values. The property table may provide value arrays for only a subset of the properties of its class, but class properties marked `required: true` must not be omitted.
+Binary property values are stored in a **property table**. A property table must specify its class (`class`), which refers to a class ID in the `3DTILES_metadata` extension of root tileset JSON,  a dictionary of properties (`properties`), where each key is a property ID correspond to a class property and each value is the index of the buffer view containing property values, and a count (`count`) for the number of elements in the property table. The property table may provide value arrays for only a subset of the properties of its class, but class properties marked `required: true` must not be omitted.
```
ptrgags, line 390:
```suggestion
Binary property values are stored in a **property table**. A property table must specify its class (`class`), which refers to a class ID in the `3DTILES_metadata` extension of root tileset JSON, a dictionary of properties (`properties`), where each key is a property ID correspond to a class property and each value is the index of the buffer view containing property values, and a count (`count`) for the number of elements in the property table. The property table may provide value arrays for only a subset of the properties of its class, but class properties marked `required: true` must not be omitted.
```

ptrgags (commented):
While I haven't yet started the implementation, this change looks sensible to me.

extensions/3DTILES_implicit_tiling/schema/subtree/contentMetadata.schema.json
lilleyse: Unlikely that they would diverge, but I think two files is ok for now.

javagl (commented):
I think it's good to align the structure more with that of `EXT_structual_metadata`. 
I'm a bit confused about `tileMetadata` and `contentMetadata`, though: They are ... wrappers for an integer. If I had to find a reason for that, could only come up with ~"well, maybe, some day, we want to store other things there, beyond the property table index". But that sounds a bit vague. Is there any _specific_ reason to wrap this index into objects?

lilleyse:
> I'm a bit confused about `tileMetadata` and `contentMetadata`, though: They are ... wrappers for an integer. If I had to find a reason for that, could only come up with ~"well, maybe, some day, we want to store other things there, beyond the property table index". But that sounds a bit vague. Is there any _specific_ reason to wrap this index into objects?

Future proofing was the main reason. Maybe there would be other encodings in the future? Metadata inheritance is also in the back of my mind, but now that I think about it some more those fields would be part of the property table definition. It should be ok to simplify and make them integers instead of wrappers for integers. I'll make the change.

lilleyse:
@javagl updated

javagl:
> Future proofing was the main reason. 

So now it will be my fault if we hit a road block in the future :-) But I don't see any reason now to not merge it.

PR #644, opened 2022-03-03
Update descriptions of offset/scale/min/max in schema

javagl:
Based on the changes in the 3D Metadata Specification from https://github.com/CesiumGS/3d-tiles/pull/640 , the descriptions in the schema have been updated.

The corresponding PR for `EXT_structural_metadata` is at https://github.com/CesiumGS/glTF/pull/58


lilleyse:
`noData` and `default` properties should be clarified as well

* `noData` is before transforms are applied
* `default` - I think this would be the final transformed value, i.e. offset/scale are not applied to it

PR #646, opened 2022-03-04
Clarfiy what it means to be an alphanumeric identifier

lilleyse:
Follow up to https://github.com/CesiumGS/glTF/pull/60. Includes more strict definition for "alphanumeric identifier".

### DIFF
--- a/extensions/3DTILES_metadata/README.md
+++ b/extensions/3DTILES_metadata/README.md
@@ -159,7 +159,7 @@ Schemas may be embedded in tilesets with the `schema` property, or referenced ex
 
 Template for entities. Classes provide a list of property definitions. Every entity must be associated with a class, and the entity's properties must conform to the class's property definitions. Entities whose properties conform to a class are considered instances of that class.
 
-Classes are defined as entries in the `schema.classes` dictionary, indexed by an alphanumeric class ID.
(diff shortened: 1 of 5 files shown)


javagl (commented):
I wondered whether it could be worth moving the block from **ID** into an own paragraph like "Identifiers", explain the regex and further constraints there once, and then just link to that from every **ID** section.

(I could create a draft PR on top of this one, just to see how this could look like)

Also, I wondered if there should be any constraints for Semantic names (maybe not even at https://github.com/CesiumGS/3d-tiles/blob/637e53f2d0af1968b1e0a60a0863fbe2debb2759/specification/Metadata/README.md#semantic , but directly in the semantics docs). The constraints that are described there verbally could be captured in a RegEx as well. (An aside: I thought there was a requirement for "custom" semantics to start with `_`, but I didn't find it right now...?)

lilleyse:
> I wondered whether it could be worth moving the block from **ID** into an own paragraph like "Identifiers", explain the regex and further constraints there once, and then just link to that from every **ID** section.
> 
> (I could create a draft PR on top of this one, just to see how this could look like)

Sure, go for it!

lilleyse:
@javagl and I reviewed each other's changes here, merging.

PR #647, opened 2022-03-04
Added dedicated identifiers section

javagl:
Moved the part that defines alphanumeric identifiers into an own section, on top of https://github.com/CesiumGS/3d-tiles/pull/646

The section itself is at https://github.com/javagl/3d-tiles/tree/alphanumeric-clarification-section/specification/Metadata#identifiers (Note: It is not added to the TOC - it mainly serves as an anchor to link to it explicitly). Directly below that is an example of its usage, in `Schema -> ID`. 

The `3DTILES_metadata` has also be updated (where it mentioned IDs), to point to this section, e.g. as in https://github.com/javagl/3d-tiles/tree/alphanumeric-clarification-section/extensions/3DTILES_metadata#class


lilleyse:
Looks good!

Issue #637, opened 2022-02-23
Should tile content support multiple groups?


javagl:
The [current `content.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/5a4bc564360a96623315d5496c3b51c1230bcf38/specification/schema/content.schema.json#L29) allows assigning a `group` name to tile content. The idea is to define groups like `trees`, `buildings` and `cars` (as in the [example image in `3DTILES_multiple_contents`](https://github.com/CesiumGS/3d-tiles/tree/5a4bc564360a96623315d5496c3b51c1230bcf38/extensions/3DTILES_multiple_contents#overview)).

One could make a case for allowing the definitions of *multiple* groups for each content. 

The arbitrary nature of what could be considered to be a "group" in each possible domain makes it easy to come up with examples, even though they might not be _specific_, _realistic_ examples, but are rather intended to underline the question of whether this should be possible _conceptually_:

![Cesium multiple groups for content](https://user-images.githubusercontent.com/5597569/155340233-cdeb23d7-5949-4b4f-813c-a1cd92f99ea1.png)

Thinking this further, it raises questions about the details of what could or should be specified here. 

One could just consider these group names as some sort of "tags", that might have a _meaning_ that depends on the domain and use case, but without any further constraints. 

If this was supposed to be used and defined a bit more formally, it might be necessary to define something like a "group**ing**", so that the group assignments yield a [partition of a set](https://en.wikipedia.org/wiki/Partition_of_a_set). Referring to the example: If one content has the groups `["greenBuildings", "officeBuildings"]` then this is fine. If one content has the groups `[ "greenBuildings", "blueBuildings" ]`, then this could raise questions. A "grouping" in that manner could be defined as the set of group names for one partition, and each content may only belong to (at most) one of them. This would have to be defined in a "top-level" object. But further details have to be sorted out if the option to assign multiple groups is actually considered, and depending on the constraints that should be applied for these group names.

lilleyse:
> One could just consider these group names as some sort of "tags", that might have a _meaning_ that depends on the domain and use case, but without any further constraints. 

This is how I think of groups, the most general interpretation possible. Additional constraints might be best left out of scope for now.

[...]

PR #524, opened 2021-10-25
3DTILES_metadata: Remove 'name' and 'description' from entity schemas

donmccurdy:
_Branched from https://github.com/CesiumGS/3d-tiles/pull/518 – no need to review until after that PR is merged._

Removed `name` and `description` from entity schemas. Entities should use properties with equivalent semantics instead.

Diff:

- https://github.com/donmccurdy/3d-tiles/compare/3DTILES_metadata-v2.1...donmccurdy:3DTILES_metadata-v2.2

PR #648, opened 2022-03-04
Changed top-level groups from dictionary to array

javagl:
As one result from https://github.com/CesiumGS/3d-tiles/issues/637 , the top level `groups` (that are defined for the tileset) should not be a dictionary, but an array, where each group has a unique identifier.

### DIFF
--- a/extensions/3DTILES_metadata/schema/group.schema.json
+++ b/extensions/3DTILES_metadata/schema/group.schema.json
@@ -10,9 +10,22 @@
         }
     ],
     "properties": {
+        "id": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Unique identifier for the group."
+        },
+        "name": {
+            "type": "string",
+            "minLength": 1,
+            "description": "The name of the group, e.g. for display purposes."
+        },
         "class": {},
         "properties": {},
         "extensions": {},
         "extras": {}
-    }
+    },
+    "required": [
+        "id"
+    ]
 }
--- a/extensions/3DTILES_metadata/README.md
+++ b/extensions/3DTILES_metadata/README.md
@@ -481,9 +481,9 @@ Metadata assigned to implicit tiles is stored in a more compact binary form. See
 
 Tiles may contain more than one content (see: [`3DTILES_multiple_contents`](../3DTILES_multiple_contents)), or multiple tiles may reference content sharing the same metadata. In these cases, metadata assigned to the tile would be inadequate or inefficient for describing tile contents. This extension allows content to be organized into collections, or "groups", and metadata may be associated with each group. Groups are useful for supporting metadata on only a subset of a tile's content, or for working with collections of contents as layers, e.g. to manage visibility or visual styling.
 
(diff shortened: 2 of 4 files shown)


extensions/3DTILES_metadata/schema/group.schema.json
```diff
@@ -10,9 +10,22 @@
...
+        "id": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Unique identifier for the group."
+        },
+        "name": {
+            "type": "string",
+            "minLength": 1,
+            "description": "The name of the group, e.g. for display purposes."
+        },
```
lilleyse, line 22: `name` and `description` were removed from metadata entities in https://github.com/CesiumGS/3d-tiles/pull/524.

lilleyse:
Also, make sure to include the `dict -> array` change in the revision history

extensions/3DTILES_metadata/schema/group.schema.json
```diff
@@ -10,9 +10,22 @@
         }
     ],
     "properties": {
+        "id": {
+            "type": "string",
+            "minLength": 1,
+            "description": "Unique identifier for the group."
+        },
```
lilleyse, line 17: From offline discussion: Now that https://github.com/CesiumGS/3d-tiles/pull/646 is merged this section can be updated similar to `schema.schema.json`.

javagl:
The inlined comments have been addressed.

extensions/3DTILES_metadata/schema/group.schema.json
lilleyse: Hm this is a bit of an edge case.. no other metadata entities have an id property (unless they use the `ID` semantic which doesn't have the alphanumeric requirement). This might be an inconsistency we'll have to think more about.

javagl: I might have misunderstood the statement from

> Slightly orthogonal - but I wonder if the top level groups should be an array instead of a dictionary, and references should be indices instead of strings **(each group definition would have an id property).**

<sup>Emphasis by me</sup>

from https://github.com/CesiumGS/3d-tiles/issues/637#issuecomment-1049005959 - what did this refer to if not an `id` property like the one above?

lilleyse: Oh I didn't meant to say this PR was doing anything unexpected.

I'm just rethinking my original suggestion to add an `id` since no other metadata entities have an `id` and I don't think groups are necessarily a special case.

javagl: Yeah, one could argue a bit here: 

Q: "Which tile is this?" 
A : "The tile with url `example.com/t.b3dm`"

vs. 

Q: "Which group is this?"  
A: 42!

Groups are "special" in that they do not have any representation in the _structure_. One of their main purposes is giving a _meaning_ to an otherwise random collection of things. I'd consider that a reason why giving them an _id_ as well as a _name_ and _description_ may be more justified.

lilleyse: `id`, `name`, `description` could be given by metadata semantics. It depends if we want to enforce `id` or not.

PR #652, opened 2022-03-07
Disallow default and noData if required is true

lilleyse:
This came up in a recent CesiumJS PR: https://github.com/CesiumGS/cesium/pull/10172#discussion_r821151177.

`default` and `noData` must not be defined if `required` is true.

Once this is merged I'll update `EXT_structural_metadata`


javagl (approved):
That seems to make sense.

lilleyse:
Fixed in EXT_structural_metadata in https://github.com/CesiumGS/glTF/pull/56/commits/7e3a7701eb1067b0d413ccc13ec2be5f0206cb7e

Issue #632, opened 2022-02-21
Avoiding duplications for the JSON schema and implementations


javagl:
The [3D Metadata Specification](https://github.com/CesiumGS/3d-tiles/tree/extension-revisions/specification/Metadata) describes a standard format for metadata. This specification itself does not define a certain serialization format for the metadata schema. This is left to the _implementations_ of the 3D Metadata Specification.

Currently, there are two implementations:

- [`3DTILES_metadata`](https://github.com/CesiumGS/3d-tiles/tree/39ab24b1d194fd0263c62f188786cbc83e239263/extensions/3DTILES_metadata/schema)
- [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/tree/6daf0eb3ac8e174fbf6b80e04f82c0ae84acee35/extensions/2.0/Vendor/EXT_structural_metadata/schema)

These JSON schemas essentially define _the same structure_ (and intentionally so!). The only difference is that the first one refers to `TilesetProperty` and the second one refers to `glTFProperty` as the "root" object (but these, in turn, are also equal except for their name).

The result of this duplication is that clients may be forced to maintain different implementations of the same thing. For example, the code generator that generates the structures for `cesium-native` from the JSON schema currently generates [a `Property` class in the `Cesium3DTiles` namespace](https://github.com/CesiumGS/cesium-native/blob/a312bf42d6de1e1ef04e930c2b86acb4bc0eb1ec/Cesium3DTiles/generated/include/Cesium3DTiles/Property.h), and a [`ClassProperty` in the `CesiumGltf` namespace](https://github.com/CesiumGS/cesium-native/blob/a312bf42d6de1e1ef04e930c2b86acb4bc0eb1ec/CesiumGltf/generated/include/CesiumGltf/ClassProperty.h). (Differences in these classes are due to the use of different states of the schema - that's exactly the point...)

We should try to find a solution that avoids this duplication. There are, roughly speaking, three options:

- Define these strucures only in `3DTILES_metadata`, and let `EXT_structural_metadata` *depend* on that
- Define these strucures only in `EXT_structural_metadata`, let `3DTILES_metadata` *depend* on that
- Define these structures independently, and let `EXT_structural_metadata` and `3DTILES_metadata` *depend* on that.

[...]

PR #654, opened 2022-03-08
Draft of directory layout for schemas for 3D Tiles 1.1

javagl:
This is a **draft** for the directory layout of the JSON schema files for 3D Tiles 1.1. It is supposed to eventually address https://github.com/CesiumGS/3d-tiles/issues/632. 

There is the `specification/schema` directory that contains the main JSON schema for 3D Tiles 1.1. This will just contain the updated/extended schema files from 1.0, but hardly any _new_ schema files. The `schema` directory contains subdirectories that define different elements of 3D Tiles 1.1:
- The `Schema` directory defines the elements that are required for describing the Metadata Schema (corresponding to `3DTILES_metadata`). This is supposed to be used by 3D Tiles 1.1 and by `EXT_structural_metadata`. 
- The `PropertyTables` directory contains the definition of property tables. This is supposed to be used by 3D Tiles 1.1. (for the Metadata of implicit tilesets), and `EXT_structural_metadata`. 
- The `Statistics` directory contains the Metadata statistics structure that is used in 3D Tiles 1.1
- The `subtrees` directory contains the JSON schema for the subtree JSON in `subtree` files for implicit tiling in 3D Tiles 1.1

(This is roughly inspired by the image shown in https://github.com/CesiumGS/3d-tiles/issues/632#issuecomment-1061165381 )

The idea is that these directories could be referred to as "search paths". For example, the `EXT_structural_metadata` extension could contain a `"$ref": "propertyTable.schema.json"` at some point. This reference would then assumed to be resolved against the schema in the `PropertyTables` subdirectory.

Some things that have to be sorted out:

- All elements for the Schema definition in `3DTILES_metadata` and `EXT_structural_metadata` inherited from `tilesetProperty` and `glTFProperty`, respectively. This "root object" only defines the `extras/extensions`. 
- Several parts referred to a `definitions.schema.json` with common definitions of "numeric values" etc. 

One option would have been to add the files `rootProperty.schema.json`, `extras.schema.json`, `extension.schema.json`, and `definition.schema.json` into _each_ subdirectory. But for now, I moved them into a `common` directory. This directory could be considered as a "search path" for all other schemas, but it has no other dependencies for itself.


lilleyse:
I like where this is heading. Some thoughts:

* Is the `README.md` in each subdirectory useful? The cross-linking information might be useful for us but it might just look like clutter for other readers.
* Could `metadataEntity.schema.json` move to `common` and  `subtreeMetadata.schema.json` be removed?
* There are some relative paths to subfolders, like `Statistics/statistics.schema.json`, but this isn't consistent everywhere. Should all the refs be relative paths? The only exception would be `EXT_structural_metadata` which can't use relative paths.
* Should `PropertyTable`, `Schema`, and `Statistics` folders be lower case? Or should `subtree` be upper case? Is there a reason to go one way or the other?
* Should feature table and batch table-related schemas move to their own folder? This would help declutter. Perhaps even `properties.schema.json` since it's deprecated. I'm not sure of a good name. Maybe `legacy` or `deprecated`?
* Can `tilesetProperty.schema.json` (and `extension.schema.json` and `extras.schema.json`) be removed? All reference to `tilesetProperty` could go to `rootProperty` instead.

specification/schema/content.schema.json
```diff
@@ -18,6 +18,18 @@
...
+            "allOf": [
+                {
+                    "$ref": "metadataEntity.schema.json"
+                }
+            ],
+            "description": "Content metadata."
+        },
+        "group": {
+            "type": "string",
+            "description": "The group this content belongs to. The value must correspond to a group ID declared in the `groups` dictionary."
```
lilleyse, line 31: The `draft-1.1-schema` branch was out of date here. It should now be:

```suggestion
            "description": "The group this content belongs to. The value is an index into the array of `groups` that is defined for the containing tileset."
```

specification/schema/content.schema.json
```diff
@@ -18,6 +18,18 @@
             "type": "string",
             "description": "A uri that points to tile content. When the uri is relative, it is relative to the referring tileset JSON file."
         },
+        "metadata": {
+            "allOf": [
+                {
+                    "$ref": "metadataEntity.schema.json"
+                }
+            ],
+            "description": "Content metadata."
```
lilleyse, line 27: At some point we'll need to flesh out the new properties' descriptions since I hastily put those together the first time around. Doesn't have to be this PR though.

specification/schema/tile.schema.json
```diff
@@ -66,7 +66,32 @@
...
+            },
+            "minItems": 1
+        },
+        "metadata": {
+            "allOf": [
+                {
+                    "$ref": "metadataEntity.schema.json"
+                }
+            ],
+            "description": "Content metadata."
```
lilleyse, line 86: This looks like my mistake. It should be `"Tile metadata."`

javagl:
> Could metadataEntity.schema.json move to common and subtreeMetadata.schema.json be removed?

The `common` part until now is completely agnostic of everything else. I _think_ that it makes sense to move `metadataEntity` to `Schema` (although it's not part of the schema, it's so strongly related to that _conceptually_). I moved it there for now, and removed `subtreeMetadata`, but I can move it to `common`, if desired. 

> There are some relative paths to subfolders, like Statistics/statistics.schema.json, but this isn't consistent everywhere. Should all the refs be relative paths? The only exception would be EXT_structural_metadata which can't use relative paths.

That was a leftover from my attempts to validate this. I think that there should be no such relative paths (so that was fixed). 

> Should PropertyTable, Schema, and Statistics folders be lower case? Or should subtree be upper case? Is there a reason to go one way or the other?

I should have exepcted that. The distinction was roughly that of `Concepts` vs. `technicalities`, but that's a bit vague. A subfolder called `schema/schema` may look a bit more odd than `schema/Schema`, but that's also subtle. I don't have a strong preference. If there is one, drop me a note. Otherwise,... let's see how this turns out.

> Should feature table and batch table-related schemas move to their own folder? This would help declutter. Perhaps even properties.schema.json since it's deprecated. I'm not sure of a good name. Maybe legacy or deprecated?
> Can tilesetProperty.schema.json (and extension.schema.json and extras.schema.json) be removed? All reference to tilesetProperty could go to rootProperty instead.

I also wondered about this. But I'd be _reeeally_ hesitant to move (or even remove) a `schema.json` file for a _minor_ version update. One could probably think deeply about the meaning of "backward compatibility" (and I did that, and could try to summarize some of my thoughts here). But my first shot would be to *not* remove anything that existed in 1.0, unless we're _absolutely_ sure that removing it will not cause any form of incompatibility.

> Is the README.md in each subdirectory useful? The cross-linking information might be useful for us but it might just look like clutter for other readers. I think it's clear enough that these are groupings that can be linked to each other.

The original `schema` directory contained a `README.md` (and admittedly, I found that a bit odd - the `schema` should contain the `schema`...). But that top-level README will have to be extended anyhow, so I'll collect the information there for now (and remove the sub-READMEs)

lilleyse:
> The `common` part until now is completely agnostic of everything else. I _think_ that it makes sense to move `metadataEntity` to `Schema` (although it's not part of the schema, it's so strongly related to that _conceptually_). I moved it there for now, and removed `subtreeMetadata`, but I can move it to `common`, if desired. 

Hm, my preference is to keep `Schema` focused on the schema definition and not mixed with the encoding. Maybe `metadataEntity` could stay in the root folder and subtree could just refer to it from there?

> I should have exepcted that. The distinction was roughly that of `Concepts` vs. `technicalities`, but that's a bit vague. A subfolder called `schema/schema` may look a bit more odd than `schema/Schema`, but that's also subtle. I don't have a strong preference. If there is one, drop me a note. Otherwise,... let's see how this turns out.

I figured the distinction was based on something like that. Uppercase for all of them would be my preference, or at least subtree as uppercase.

> I also wondered about this. But I'd be _reeeally_ hesitant to move (or even remove) a `schema.json` file for a _minor_ version update. One could probably think deeply about the meaning of "backward compatibility" (and I did that, and could try to summarize some of my thoughts here). But my first shot would be to *not* remove anything that existed in 1.0, unless we're _absolutely_ sure that removing it will not cause any form of incompatibility.

Sounds good.

lilleyse:
> I also wondered about this. But I'd be _reeeally_ hesitant to move (or even remove) a `schema.json` file for a _minor_ version update. One could probably think deeply about the meaning of "backward compatibility" (and I did that, and could try to summarize some of my thoughts here). But my first shot would be to *not* remove anything that existed in 1.0, unless we're _absolutely_ sure that removing it will not cause any form of incompatibility.

Another idea is to move them into a `TileFormats` folder. In that case `properties.schema.json` wouldn't be moved there. Is there a reason you're hesitant to move schema files around?

javagl:
The `metadataEntity` was moved to the root directory. 
The `subtree` folder was renamed to `Subtree`. (Quirky git magic involved here...). 

> Another idea is to move them into a TileFormats folder. In that case properties.schema.json wouldn't be moved there. Is there a reason you're hesitant to move schema files around?

I just created such a `TileFormats` folder and moved the batch- and feature table parts there. It is definitely _cleaner_ from a "modularity perspective": These JSON schemas are "unrelated" to the tileset, in that they have their own "context" (namely, the binary representation of the tile formats). I don't see a _specific_ reason to _not_ move them there, except for https://xkcd.com/1172/

PR #655, opened 2022-03-09
Add ImplicitTiling subdirectory to 1.1 spec

lilleyse:
Part of https://github.com/CesiumGS/3d-tiles/issues/651

Adds an `ImplicitTiling` subdirectory to the `specification` folder. This is nearly identical to [`3DTILES_implicit_tiling`](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_implicit_tiling) with extensions removed, relative paths fixed, etc.

PR #656, opened 2022-03-09
Update links in 3D Metadata and Semantics

lilleyse:
Like https://github.com/CesiumGS/3d-tiles/pull/655, removed contributors list and links to extensions for the metadata docs. Some links are still `TODO` until those sections are added.

### DIFF
--- a/specification/Metadata/README.md
+++ b/specification/Metadata/README.md
@@ -2,22 +2,6 @@
 # 3D Metadata Specification
 
 <!-- omit in toc -->
-## Contributors
-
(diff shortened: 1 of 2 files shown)


lilleyse:
> I assume that before merging the `draft-1.1` branch, there will a dedicated review, **not** so much referring to technicalities, but rather on whether there are still `TODO`s, and maybe a "click-through" to check that the `../../LinkTargets` are correct.

Yes, we'll do a final pass before `draft-1.1` is merged into `main`.

PR #657, opened 2022-03-09
Move revision entry under correct date

j9liu:
I noticed that the `groups` change #648 was logged under October 2021 instead of February 2022 in `REVISIONS.md`, so here's a quick fix.


lilleyse:
Good catch!

PR #658, opened 2022-03-09
Move `Styling/schema` into `schema/Styling`

lilleyse:
Now that we have subdirectories in the schema folder for dedicated concepts it makes sense to include Styling there as well.


javagl (approved):
Looks fine.

(An aside: I always wondered about _which_ tool exactly uses `omit in toc` (and how) - seems to be VSCode in some form?)

lilleyse:
> (An aside: I always wondered about _which_ tool exactly uses `omit in toc` (and how) - seems to be VSCode in some form?)

It's this extension: https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one

PR #659, opened 2022-03-10
Generalize common schema definitions

javagl:
I had added common, basic building blocks in `definitions.schema.json` that could be reused for other definitions (during https://github.com/CesiumGS/3d-tiles/pull/654 ). This has _not_ yet extensively been reviewed and validated, but I'll open it as a draft PR for now.


lilleyse:
Looks good!

PR #664, opened 2022-03-10
Update specification for 1.1 (except metadata)

lilleyse:
* Updated version number
* Added intro section about glTF content
* Added section about implicit implicit
* Updated "File Extensions and Media Types" section
* Added multiple contents section
* Removed property reference (for now) until we have a chance to regenerate it


IanLilleyT:
We reviewed everything except the last commit in person, and the last commit looks good

PR #667, opened 2022-03-13
Remove tilesetProperty.schema.json

lilleyse:
Removes `tilesetProperty.schema.json`, `extension.schema.json`, and `extras.schema.json` in favor of `rootProperty.schema.json`, `extension.schema.json`, and `extras.schema.json` in the `common` folder.


lilleyse:
I think this is a case where the benefits outweigh the risk since it's such a structurally useful change for the schema. Though it is good to know the risks.

My hope is that there aren't too many code generators that need to be fixed since `tilesetProperty.schema.json` itself is a relatively new addition (since https://github.com/CesiumGS/3d-tiles/pull/497). And technically we already broke subtrees by removing `subtreeProperty.schema.json`.

Issue #631, opened 2022-02-21
Maintaining specifications for 3D Tiles 1.0, 1.1, and beyond


javagl:
There are efforts for the standardizing 3D Tiles version  **1.1**. This version will be based on what is currently referred to as "3D Tiles Next", which currently is a set of (draft) extensions for 3D Tiles 1.0, that offer the core functionality that is supposed to be standardized as version 1.1. 

We have to find a way to sensibly maintain and organize different versions in this repository. The approaches here could roughly be categorized as follows:

- Always maintaining the "latest" state, and using Git tags for versioning
- Creating different subdirectories for the different versions

3D Tiles 1.0, 'Next', and 1.1 should coexist, should easily be accessible, and may have to be maintained independently (for smaller bugfixes). And some of the current repository contents will be (largely) independent of the underlying version (but that might change if there is a version 2.0 at some point). So I think that trying to find a _directory_-based versioning mechanism makes more sense in our case.

#### Current state

Right now, the directory structure of the repository is

    
    \                             Main README
    
    \specification                Specification README for 1.0
    \specification\schema         Schema for 1.0
    \specification\TileFormats    Tile formats of 1.0     

    \specification\Metadata       Specification for Metadata  
    \specification\Styling        Styling functionality
    
    \extensions\3DTILES_...       Extensions for 1.0 - these "ARE" "3D Tiles Next"!

    \reference-card               Reference card for 1.0  
    
    \next                         Main README for 3D Tiles Next
    \next\reference-card          Reference card for 3D Tiles Next

<sup>(Omitted some additional files, like `figures` directories that refer to the respective README etc.)</sup>

#### Directory layout options

The options for organizing different versions in such a directory structure differ in the detail. This basically refers to the level of "granularity" on which the version numbers are inserted, affecting the question of _which part a certain version applies to_.

Very roughly speaking, one could start "pragmatically", with a layout like this:

    \specification\1.0                Specification README for 1.0
    \specification\1.0\schema         Schema for 1.0
    \specification\1.0\TileFormats    Tile formats of 1.0     

    \specification\1.1                Specification README for 1.1
    \specification\1.1\schema         Schema for 1.1

With the obvious caveats:

- The `reference-card` subfolders require versioning. But are not part of the "specification" in the narrow sense.
- The `extensions` should be separate of the "core" specification. But they strictly belong to one version. And right now, **all** extensions belong to 3D Tiles 1.0, and are what is referred to as "3D Tiles Next"
[...]

PR #668, opened 2022-03-13
Delete 1.0 specification documents from draft-1.1 branch

lilleyse:
Removes the 3D Tiles 1.0 specification documents (.docx and .pdf) from the draft-1.1 branch until we have a chance to regenerate these from the 1.1 spec.


javagl:
They will always be in the version history. But I mentioned the point that we might need some "releases" (or "archive") folder at some point in https://github.com/CesiumGS/3d-tiles/issues/631#issuecomment-1068048239 . So removing them here for now should be OK.

javagl:
The `CONTRIBUTING.md` contained instructions of how the `docx` and `pdf` files that have been deleted here had been created. 

This `CONTRIBUTING.md` is about to be deleted (because it is obsolete _for now_). The link to its latest version here, for convenience: https://github.com/CesiumGS/3d-tiles/blob/b114fc3c9ac82b869634ae4ef57d564da725887a/CONTRIBUTING.md

PR #670, opened 2022-03-14
Regenerate implicit tiling diagrams as PNG and fix morton index mistake

lilleyse:
Replaced implicit tiling diagrams with png images instead of jpg. This slightly improves image crispness but not entirely since most of the images are being rescaled by the browser.

Also fixes a mistake with the morton indexes in two diagrams:

screnn_implicit


javagl (commented):
Looks fine for me

PR #671, opened 2022-03-14
Add tileset metadata sections

javagl:
This mainly adds a "Metadata" section to the main 3D Tiles (1.1) README.md. The contents is largely based on the README from `3DTILES_metadata`.

Preview https://github.com/javagl/3d-tiles/blob/draft-1.1-tileset-metadata-sections/specification/README.md#metadata


javagl:
> It wouldn't be a bad idea to copy the metadata paragraph in the introduction and use it as the first paragraph in the metadata section

There is some overlap with the "granularity levels" description directly after that, but having that an introductory paragraph is OK. 

> Should the Metadata Statistics section go after the Assigning Metadata? 

That makes sense. Changed it accordingly. 

> diagram for statistics. I think it's worth including here. 

I added this diagram as well.

> I agree with removing some examples

I found it hard to decide which ones to remove. But I think that describing one `metadataEntity` can cover the tileset/tile/content case anyhow, and the overlap with the multiple contents example could also be avoided. Now, the sections that are called **"... Properties"** are pretty short. One could restructure this further, to omit them, and  essentially "inline" these paragraphs into the "Granularities" bullet points. For example,

>  - **Tileset** - The tileset as a whole may be associated with global metadata, such as the year of publication.
> 
> ...
> 
> ##### Tileset Properties 
> 
> Properties assigned to tilesets provide metadata about the tileset as a whole. Common examples might include year of collection, author details, or other general context for the tileset contents.

could just become

>  - **Tileset** - The tileset as a whole may be associated with global metadata. Common examples might include year of collection, author details, or other general context for the tileset contents.
> 
> <sub>(Omitting the 'Tileset Propeties' section)</sub>

lilleyse:
> One could restructure this further, to omit them, and essentially "inline" these paragraphs into the "Granularities" bullet points.

Yeah, this seems like a good idea to keep all the examples in a single place.

lilleyse:
@javagl The rest of the changes look good. After that last restructuring this should be ok to merge.

Issue #651, opened 2022-03-07
3D Tiles 1.0 to 1.1 checklist


lilleyse:
Checklist as we update 3D Tiles 1.0 to 3D Tiles 1.1. Most of these changes will go into the [draft-1.1](https://github.com/CesiumGS/3d-tiles/tree/draft-1.1) branch.

### Prerequisites

* [x] Finish https://github.com/CesiumGS/3d-tiles/pull/630
* [x] Finish https://github.com/CesiumGS/glTF/pull/56
* [x] Finish https://github.com/CesiumGS/3d-tiles/pull/634

### General
* [ ] https://github.com/CesiumGS/3d-tiles/issues/631
* [x] https://github.com/CesiumGS/3d-tiles/issues/632
* [x] Fix broken links
* [x] Remove mentions of 3D Tiles Next
* [x] Refer to 1.1 sections instead of extensions wherever applicable
* [x] What is the best way to incorporate s2, gltf, implicit tiling, metadata, and multiple contents?
* [x] Use .gltf or .glb everywhere in examples

### Specification

* [x] Metadata
  * [x] Restructure `Metadata` folder: https://github.com/CesiumGS/3d-tiles/issues/632
  * [x] Add `3DTILES_metadata` concepts to the main specification. Link to `Schema` and `PropertyTable` folders in the reference implementation.
  * [x] Move implicit tiling property table definition to `Metadata/ReferenceImplementation/PropertyTable` folder
* [x] Implicit tiling
  * [x] Add `ImplicitTiling` folder and README: https://github.com/CesiumGS/3d-tiles/pull/655
  * [x] Spatial data structures - rework to explain difference between explicit and implicit tiling. Add short summary and image for implicit tiling and link to `ImplicitTiling`.
* [x] Multiple contents
  * [x] Add to `specification/README`
  * [x] Explain distinction between `content` and `contents` and rationale for having both (i.e. avoids a breaking change), and when to use one over the other (prefer `content` for single content, better for backwards compatibility)
* [x] TileFormats
  * [x] Update individual pages with a deprecation message. Point to migration guide.
  * [x] Add `glTF` folder
    * [x] Expand on intro summary in the main spec. Show how glTF can be used for heterogeneous 3D models, instanced models, and point clouds
    * [x] Include recommended glTF extensions: https://github.com/CesiumGS/3d-tiles/issues/602
    * [x] Include details about feature identification and metadata
    * [x] Include migration guide in appendix
[...]

Issue #560, opened 2021-11-14
3D Tiles Next - list of developer resources and the ecosystem


pjcozzi:
Looking for someone to diligently facilitate this one, any takers?  😄 

I believe that one of the main drivers of the early success of 3D Tiles and glTF was how community contributions were encouraged, organized, and findable.  For example:
* List of [glTF Presentations and Articles](https://github.com/KhronosGroup/glTF#presentations-and-articles)
* glTF projects in the [Project Explorer](https://github.khronos.org/glTF-Project-Explorer/) (which replaced GitHub issue https://github.com/KhronosGroup/glTF/issues/1058 with a mere 262 comments)

In this same spirit, I think we should start tracking all developer resources and the software ecosystem for 3D Tiles Next.  The list could be as simple as a README.md with a GitHub issue full of comments to add things to it as we discover them (and as contributors post to it) or something else sophisticated if it is needed at this point.

---

* The [The State of 3D Tiles Next](https://cesium.com/blog/2021/11/10/introducing-3d-tiles-next/#the-state-of-3d-tiles-next) is a good starting point with a list of developer resources and projects
* There's also
   * [3D Tiles Next README.md](https://github.com/CesiumGS/3d-tiles/blob/main/next/README.md)
   * [Web3D Conference 2021](https://cesium.com/learn/presentations/#web3d-conference-2021) video and slides by @ptrgags 
   * [minimal-pointcloud-gltf](https://github.com/wallabyway/minimal-pointcloud-gltf) by @wallabyway
   * [Anything from Maxar](https://cesium.com/industries/mission-support/owt/)?
   * Interest from 3DTilesRendererJS, https://github.com/NASA-AMMOS/3DTilesRendererJS/issues/185.  CC @gkjohnson and @dbuck

pjcozzi:
> I assume that a first solution for this could just be a file similar to https://github.com/javagl/3d-tiles/blob/main/RESOURCES.md ...

@javagl sure, please also link directly to this from the main README.md.

We should also duplicate the resources from the announcement blog as that list will grow quickly.

javagl:
This has been addressed in https://github.com/CesiumGS/3d-tiles/pull/579 . Maybe someone wants to have a short look (@ptrgags ?)

The list will probably be refactored when the number of resources increases, but I hope that the current structure is a reasonable start.

(BTW: I noticed that the main `RESOURCES.md` was not linked ... from anywhere... so I also added a link to that in the main `README.md`).

[...]

PR #674, opened 2022-03-17
Cleanup pass for 1.1

javagl:
This addresses some points from https://github.com/CesiumGS/3d-tiles/issues/651 : 

> Update or remove CONTRIBUTING.md

Removed it, because it was obsolete as of https://github.com/CesiumGS/3d-tiles/pull/668 

> Update Q-and-A.md

This was done. Some questions have been updated to reflect the current state. Some other questions have been removed. When the question of an answer points to a _seven_ year old open issue with 0 comments, then the question cannot be so relevant...

> Update RESOURCES.md

This was largely done by moving _some_ of the `next/RESOURCES.md` into this file. Currently, some of them are still mentioned in a "Next" section at the bottom. When the respective parts (e.g. Demos) are available for 1.1 (maybe as Sandcastles), this part can be removed as well.

---

Other things could be added to this PR, depending on the preferred solution:

> Remove examples folder. Instead there should be a prominent link in the README to [3d-tiles-samples](https://github.com/CesiumGS/3d-tiles-samples).

The `examples` folder also bugged me. One example is linked from the main README. Others are used in "hidden" places that generally talk about JSON schema. If there are no objections, I'd remove the `examples` folder entirely, and see how the updates in the "hidden" places can be made in the most sensible form.

> Update or remove CREDITS.md

It probably cannot be removed, due to attribution requirements. I think it could make sense to move the attribution to the place where it is due (i.e. into an image caption). I can _try_ that, if I'm confident to be able to figure out what the attribution refers to. (Once this is figured out, it will no longer be a problem in the future.... doing it _once_ can be hard...)

> Update or remove ECOSYSTEM.md

Deleting it would be simple. Should there be any form of replacement? It could make sense to distribute this into the "Implementations" and "Developer Resources" sections from https://github.com/javagl/3d-tiles/blob/draft-1.1-structure-pass-2/RESOURCES.md , if this sounds reasonable...


lilleyse:
> If there are no objections, I'd remove the `examples` folder entirely, and see how the updates in the "hidden" places can be made in the most sensible form.

No objections from me

> It probably cannot be removed, due to attribution requirements. I think it could make sense to move the attribution to the place where it is due (i.e. into an image caption). I can try that, if I'm confident to be able to figure out what the attribution refers to. (Once this is figured out, it will no longer be a problem in the future.... doing it once can be hard...)

Yeah, makes sense to move the attribution to the specific images.

> Deleting it would be simple. Should there be any form of replacement? It could make sense to distribute this into the "Implementations" and "Developer Resources" sections from https://github.com/javagl/3d-tiles/blob/draft-1.1-structure-pass-2/RESOURCES.md , if this sounds reasonable...

I don't have to much to say about the specific structure, but I do think combining `RESOURCES.md`, `next/RESOURCES.md`, and `ECOSYSTEM.md` into a single document is a good idea.

lilleyse:
> I don't have to much to say about the specific structure, but I do think combining `RESOURCES.md`, `next/RESOURCES.md`, and `ECOSYSTEM.md` into a single document is a good idea.

Actually a few more thoughts here... Instead of a separate document this could go in the main `README`. There's already an [ecosystem](https://github.com/CesiumGS/3d-tiles#3d-tiles-ecosystem) section that could be expanded into several sections like demos, tools, resources, presentations, etc that would pull together the material that is currently spread across a few different pages:

* `RESOURCES.md`
* `next/RESOURCES.md`
* `ECOSYSTEM.md`
* `README/#ecosystem`
* https://github.com/CesiumGS/3d-tiles/issues/560

The [glTF repo](https://github.com/KhronosGroup/glTF) is a good reference for this. The [CesiumJS repo](https://github.com/CesiumGS/cesium) is also good, especially the featured demos section.

javagl:
- Removed `examples` folder
- Deleted `CREDITS.md` and moved credits into captions of images
- Removed mention of ion in Q&A

---

Regarding the "sitemap": 

> I have mixed feelings about this page. It's a convenient way to locate things. But it's another page to keep up to date and it's hard to know the value it will bring vs. just navigating the directory structure.

This was a first draft, mainly to see which links it could contain. I considered to actually create some "table"

|  |  |
|---|---|
| - [Metadata](/specification/Metadata/README.md) | Some description here |
| &nbsp;&nbsp;&nbsp;&nbsp;- [Semantics Reference](/specification/Metadata/Semantics/README.md) |  Some decription here |
| ... | ... |

I also see that it has to be kept up to date, although I think that it refers to such a "coarse" and "settled" structure that it should probably not change very frequently. Still, the "effort-to-usefulness" ratio has to be considered. The file can be removed at any time (from this PR), and **iff** we consider to add something like this, it can be a dedicated PR.

---

Regarding the resources/ecosystem:

> Actually a few more thoughts here... Instead of a separate document this could go in the main README. There's already an [ecosystem](https://github.com/CesiumGS/3d-tiles#3d-tiles-ecosystem) section that could be expanded into several sections like demos, tools, resources, presentations, etc that would pull together the material that is currently spread across a few different pages:
> 
> The [glTF repo](https://github.com/KhronosGroup/glTF) is a good reference for this. The [CesiumJS repo](https://github.com/CesiumGS/cesium) is also good, especially the featured demos section.

I think that the main README should be relatively short, and rather "dispatch" the user to the page that contains the desired information. But regardless of whether it will be a section in the main README or a dedicated file, I wonder about the best structure.

The `ECOSYSTEM.md` contains mainly screenshots, without information what these things actually are. The "Ecosystem" section in the main README contains some information, but only a small selection (and I'm not sure whether the validator should be featured so prominently, as long as it does not support 1.1). 

The glTF repo originally listed all the tools (and they have been moved to the glTF-Project-Explorer in the meantime), roughly classified into "Viewers/Importers/Exporters...". I can try to go through the list in the `ECOSYSTEM.md`, and have a look what this actually **is** - maybe it can also be classified in a sensible way.

lilleyse:
Let's remove the sitemap to help move this PR along. It can be a separate PR.

The ecosystem/resources restructuring can be a separate PR as well.

RESOURCES.md
```diff
@@ -12,6 +23,7 @@
    * _3D Tiles: Beyond 2D Tiling_ ([pdf](https://cesium.com/presentations/files/FOSS4GNA2016/3DTiles.pdf), [video](https://www.youtube.com/watch?v=I1vYCrMKKEE)) at FOSS4G NA 2016.
    * _3D Tiles motivation and ecosystem update_ ([pdf](https://cesium.com/presentations/files/3D-Tiles-OGC-DC.pdf)) at the OGC Technical Committee Meeting (March 2016).
    * _3D Tiles intro_ ([pdf](https://cesium.com/presentations/files/SIGGRAPH2015/Cesium3DTiles.pdf)) at the Cesium BOF at SIGGRAPH 2015.
+  * _Introducing 3D Tiles Next_, at Web3D Conference 2021. [Video and slides](https://cesium.com/learn/presentations/#web3d-conference-2021)
```
lilleyse, line 26: These are listed from most recent to least recent, so this should go first

javagl:
If this is OK, I'd update the checklist at https://github.com/CesiumGS/3d-tiles/issues/651 after this is merged.

PR #678, opened 2022-03-28
Remove `group.id` property

lilleyse:
Removes the `group.id` property so the groups are consistent with other metadata entities. Metadata entities such as tilesets, tiles, contents, and groups should use the [`ID`](https://github.com/CesiumGS/3d-tiles/tree/draft-1.1/specification/Metadata/Semantics#general-semantics) semantic.

There was some discussion about this in https://github.com/CesiumGS/3d-tiles/pull/648#pullrequestreview-900412973.

### DIFF
--- a/extensions/3DTILES_metadata/schema/group.schema.json
+++ b/extensions/3DTILES_metadata/schema/group.schema.json
@@ -10,11 +10,6 @@
         }
     ],
     "properties": {
-        "id": {
-            "type": "string",
-            "pattern": "^[a-zA-Z_][a-zA-Z0-9_]*$",
-            "description": "Unique identifier for the group within the tileset. This must be an alphanumeric identifier matching the regular expression `^[a-zA-Z_][a-zA-Z0-9_]*$`."
-        },
         "class": {},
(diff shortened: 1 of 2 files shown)


javagl (approved):
Related to the broader question of "id vs. index" : One could make a case that

- things that are values in a dictionary should not have an "id"
- things that are entries in an array should have an "id"

(The reason being related to the question of what the ID is used for. If it's used for _identifying_ things, then it has many constraints in terms of uniqueness that do not apply directly when there already are "ids", namely, the dictionary keys)

But in general, that change is fine for me, at least to have consistency.

Now becoming more relevant: https://github.com/CesiumGS/cesium/issues/10223

(Also, I'll have to update the samples that needed ID to circumvent this issue)

javagl:
> Does that just include https://github.com/CesiumGS/3d-tiles-samples/tree/main/1.1/MetadataGranularities?

Yes, this seemed to be the only model that defined groups for now. The PR is at https://github.com/CesiumGS/3d-tiles-samples/pull/46

PR #679, opened 2022-03-28
Cleanup pass for 1.1 - main README.md

javagl:
Addresses some points from https://github.com/CesiumGS/3d-tiles/issues/651 : 

> Update JSON encoding and URIs - https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#json-encoding

I extended the JSON encoding section with some information from the glTF spec. If this should include the examples, just let me know. (But at some point, it becomes a verbatim copy ... hm). 

> Update Coordinate reference system to include tileset semantics

Added a link to the "Tileset Semantics", mentioning the option to specify the CRS via metadata. 

(I hesitated a bit: The `region` is supposed to be given in "EPSG 4979". But the semantics say that the region uses the same ellipsoid as the CRS of the tileset. If this was **not** the WGS84 ellipsoid, the values would probably no longer be in "EPSG 4979", in the strictest sense. Someone who is more familiar with these concepts could have an idea about how this possible contradiction could be resolved...)

> Clean up transforms section. Move b3dm, i3dm, pnts specific information to their respective documents.

Done. The main README now focusses on the glTF case (the image was updated as well). The information that was removed from the main README has been moved into the `TileFormats` READMEs. Some of that may now be a bit "hidden" (e.g. the statement about first performing dequantizations, before applying other transforms). But I think that for someone who _dedicatedly_ looks at one of the tile formats specs, this should still be OK...)

> Add version history (include deprecations)

I drafted a first version history at https://github.com/javagl/3d-tiles/blob/51aa9ceabaaa6bdf7d9645e54c610480b4122a2d/specification/README.md#version-history - not sure whether it's the right level of detail, but think that it could be OK.


specification/TileFormats/Batched3DModel/README.md
```diff
@@ -140,7 +140,14 @@ When a Batch Table is present or the `BATCH_LENGTH` property is greater than `0`
...
 
-Vertex positions may be defined relative-to-center for high-precision rendering, see [Precisions, Precisions](http://help.agi.com/AGIComponents/html/BlogPrecisionsPrecisions.htm). If defined, `RTC_CENTER` specifies the center position that all vertex positions are relative to after the coordinate system transform and glTF node hierarchy transforms have been applied.
+Vertex positions may be defined relative-to-center for high-precision rendering, see [Precisions, Precisions](http://help.agi.com/AGIComponents/html/BlogPrecisionsPrecisions.htm). If defined, `RTC_CENTER` specifies the center position that all vertex positions are relative to after the coordinate system transform and glTF node hierarchy transforms have been applied. Specifically, when the `RTC_CENTER` is defined in the feature table of a Batched 3D Model, the computation of the [tile transform](../../README.md#tile-transforms) is done as follows:
+
+1. [glTF node hierarchy transformations](#gltf-node-hierarchy)
+2. [glTF _y_-up to _z_-up transform](#y-up-to-z-up)
+3. The transform for the `RTC_CENTER`, which is used to translate model vertices
+4. [Tile transform](#tile-transforms)
```
lilleyse, line 148: These links should point to the sections in `../../README.md#`

specification/TileFormats/Instanced3DModel/README.md
```diff
@@ -263,6 +263,15 @@ When the glTF field contains a URI, then this URI may point to a [relative exter
 By default glTFs use a right handed coordinate system where the _y_-axis is up. For consistency with the _z_-up coordinate system of 3D Tiles, glTFs must be transformed at runtime. See [glTF transforms
 ](../../README.md#gltf-transforms) for more details.
 
+When the [`RTC_CENTER`](#rtccenter) is defined in the feature table of an Instanced 3D Model, the computation of the [tile transform](../../README.md#tile-transforms) is done as follows:
+
+1. [glTF node hierarchy transformations](#gltf-node-hierarchy)
+2. [glTF _y_-up to _z_-up transform](#y-up-to-z-up)
+3. The per-instance positions and scales, as defined in the feature table of the Instanced 3D Model. 
+4. The transform for the `RTC_CENTER`, which is used to translate model vertices
+5. [Tile transform](#tile-transforms)
```
lilleyse, line 272: Same comment here. Also the `RTC_CENTER` link should be `#rtc_center`.

specification/README.md
```diff
@@ -1096,6 +1095,22 @@ The following example colors features with a height above 90 as red and the othe
 
 For complete details, see the [Declarative Styling](./Styling/) specification.
 
+
+## Version History
+
+- [3D Tiles 1.0](https://github.com/CesiumGS/3d-tiles/tree/1.0) _(2018-12-14)_, approved as the [3D Tiles Specification 1.0](http://docs.opengeospatial.org/cs/18-053r2/18-053r2.html) by the Open Geospatial Consortium
```
lilleyse, line 1101:
```suggestion
- [3D Tiles 1.0](https://github.com/CesiumGS/3d-tiles/tree/1.0) _(2018-12-14)_, approved as an [OGC Community Standard](http://docs.opengeospatial.org/cs/18-053r2/18-053r2.html)
```

specification/README.md
```diff
@@ -1096,6 +1095,22 @@ The following example colors features with a height above 90 as red and the othe
 
 For complete details, see the [Declarative Styling](./Styling/) specification.
 
+
+## Version History
+
+- [3D Tiles 1.0](https://github.com/CesiumGS/3d-tiles/tree/1.0) _(2018-12-14)_, approved as the [3D Tiles Specification 1.0](http://docs.opengeospatial.org/cs/18-053r2/18-053r2.html) by the Open Geospatial Consortium  
+- [3D Tiles 1.1](https://github.com/CesiumGS/3d-tiles)
+  - Additions:
+    - Support for structured Metadata that can be associated with tilesets, tiles, tile content, and tile content groups
```
lilleyse, line 1104:
```suggestion
    - Support for structured metadata that can be associated with tilesets, tiles, tile content, and tile content groups
```

specification/README.md
```diff
@@ -1096,6 +1095,22 @@ The following example colors features with a height above 90 as red and the othe
...
+    - Directly support glTF assets as tile contents
+    - Support for multiple tile contents
+    - Support for implicit tiling schemes
+  - Deprecations:
+    - The original tile formats (b3dm, i3dm, pnts, and cmpt) are deprecated in favor of glTF content
+    - The `tileset.properties` are deprecated, in favor of the more versatile metadata support
```
lilleyse, line 1110: The version history should go in the parent README to provide a better separation between the version history and the specification itself.

specification/README.md
lilleyse: Actually this suggestion isn't great because it's not obvious that `OGC Community Standard` links to the 3D Tiles document. But I think it's important to have the term "Community Standard" somewhere in here.

javagl:
So, I know that this is _totally_ not important, but wanted to try it out: 

The original logo that was shown in the README was not ideal for dark mode:

Cesium DarkLogo

But GitHub offers a way to switch images based on the mode that is used: https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

So now switching the mode switches the logo as well:

Cesium Logo Color

Issue #643, opened 2022-03-02
Ideas for structuring Metadata Semantics


javagl:
Each property of a metadata class can have a [Semantic](https://github.com/CesiumGS/3d-tiles/tree/5a4bc564360a96623315d5496c3b51c1230bcf38/specification/Metadata#semantic). This is just a string that can be used to assign an application-specific _meaning_ to these properties. A collection of semantics is currently summarized in the [3D Metadata Semantic Reference README](https://github.com/CesiumGS/3d-tiles/tree/5a4bc564360a96623315d5496c3b51c1230bcf38/specification/Metadata/Semantics). 

This form of maintaining the semantics reference has some limitations:

- It is not clear how clients/users could define their own custom semantics (e.g. whether there are naming conventions, or how to avoid ambiguities and name clashes) 
- It is never clear which set of semantics "exist" - where "existing" means "being defined in some README" or "being supported by some implementation"
- The set of existing semantics is not structured. Currently, there are (soft) conventions for the existing ones: They are separated into "General" ones, some for "Tiles", and some for "Content", and what they refer to is indicated by a prefix of the name (namely `TILE_*` or `CONTENT_*`), but there is no concept for enforcing or extending this.
- The _types_ of the semantics are not known at runtime

(The _list_ of available ones are known at runtime in CesiumJS - via the [MetadataSemantic](https://github.com/CesiumGS/cesium/blob/dc725e67fc8c2adf8033e93ce5513711ed43c2ac/Source/Scene/MetadataSemantic.js) class. But this does not include real type information, and has to be aligned with the README manually, and is not a scalable solution). 

---

One option to <strike>solve all problems</strike> alleviate some of these limitations could be to define the semantics in a _schema_. And this could actually be pretty much the [3DTILES_metadata schema](https://github.com/CesiumGS/3d-tiles/tree/5a4bc564360a96623315d5496c3b51c1230bcf38/extensions/3DTILES_metadata#schema). An example of two of the existing semantics and how they could be defined in a schema:

```
"schema": {
  "classes": {
    "CesiumContentMetadataSemantics": {
      "properties": {
        "CONTENT_BOUNDING_BOX": {
          "name": "Content bounding box",
          "description": "The bounding volume of the content of a tile, equivalent to tile.content.boundingVolume.box",
          "type": "SCALAR",
          "componentType": "FLOAT32"
          "array": true,
          "count": 12
        },
        "CONTENT_MINIMUM_HEIGHT": {
[...]

PR #681, opened 2022-03-31
Add `"UNKOWN"` to `"TILESET_CRS_GEOCENTRIC"`

lilleyse:
Adds a reserved value `"UNKNOWN"` to the list of possible values for `"TILESET_CRS_GEOCENTRIC"`.


erikdahlstrom:
A difference is that "UNKNOWN" can mean that there is knowledge that the region values in the dataset are not "EPSG 4979" (broken or mislabeled data could end up in this category).

javagl (commented):
Then, as such, the change is probably OK.

Broader questions about the definition of semantics are summarized in https://github.com/CesiumGS/3d-tiles/issues/643 , and _this_ change could probably be seen as the question of whether the type of this semantic should be 'enum', and whether `"UNKNOWN"` should be one of the enum values...

Issue #590, opened 2021-11-19
Use AsciiDoctor for the 3D Tiles specification


ptrgags:
To follow the lead of `glTF`, it might be good to upgrade our spec from Markdown to AsciiDoctor for nicer HTML and PDF output. It sounds like https://github.com/CesiumGS/wetzel has been updated within the past year. it would be good to see how well it fits 3D Tiles.

Some notes from @emackey about the latest details of wetzel:

>- It can convert JSON directly to AsciiDoctor now.
>- It has a new option to skip writing the TOC (because AsciiDoctor does its own TOC).
>- A parent AsciiDoctor file could use multiple include statements to deal with the multiple roots issue.
>
>[Over in glTF land we're using it](https://github.com/KhronosGroup/glTF/pull/1901/files), check out the [Makefile](https://github.com/KhronosGroup/glTF/blob/d8b8b9ac979ef0400a75d4a3dc48ae851f266185/specification/2.0/Makefile) that calls wetzel from `npx` on a command line, and the [CI.yml](https://github.com/KhronosGroup/glTF/blob/d8b8b9ac979ef0400a75d4a3dc48ae851f266185/.github/workflows/CI.yml) that runs this Makefile from within GitHub Actions.

>wetzel is run as part of the GitHub Actions CI, to convert JSON schemas directly to AsciiDoctor files You can go into the "[Actions](https://github.com/KhronosGroup/glTF/actions)" tab and download [build artifacts from the latest run](https://github.com/KhronosGroup/glTF/suites/1605015425/artifacts/29313241).

>This should give you a file `spec-outputs.zip` that has one self-contained PDF and one self-contained HTML file included.  Extract and double-click either one to see how the output is shaping up.  Section 5 is the main reference section that's auto-generated from the schema files.  There's also an Appendix A that has verbatim copies of the schema (also handled by wetzel now), so that if the PDF were sent to ISO, the real schemas would be included in the standardization process.  The remainder of the glTF specification document does not come from schemas, and was manually translated from original source Markdown to AsciiDoctor by Khronos veteran Jon Leech.

emackey:
There are some additional details that have been added to the [wetzel readme](https://github.com/CesiumGS/wetzel#common-usage).

javagl:
[...]

PR #682, opened 2022-04-07
AsciiDoc conversion experiments

javagl:
A **DRAFT** of the first experiments for converting the specification to AsciiDoc (see https://github.com/CesiumGS/3d-tiles/issues/590 )


javagl:
The current state is that it contains a `specification/Specification.adoc` that was largely auto-generated from the `specification/README.md`, and that can be used to create first, basic (not-so-pretty) PDF and HTML outputs. It also contains a `BUILDING.md` with some instructions for this process.

javagl:
An update and some sort of checklist

The conversion of the "core" of the specification to AsciiDoc is largely done. The JSON schema property references are still missing.

### Regarding the core specification

It is possible to generate the single-page HTML version of the specification. And it is possible to generate the single-document PDF version of the specification. 

What is still to be done:

- **Formatting tweaks**
  This refers to things like table column widths or image sizes, but nothing big. 
- **Fixing some links**
  _Most_ of the links should be fixed via the last ~40 commits or so. 
  The cross-linking between documents required some workaround, described [in this state of the instructions](https://github.com/javagl/3d-tiles/blob/67a5b447f1c25c13f54fc17b3fcb511ad323c55a/BUILDING.md#cross-linking-between-files). Suggestions for better solutions are welcome. 
  Systematically going through all documents and checking all links could be tedious, but doable. For Markdown, I once created a small tool that actually checks all the cross-links, and reports them as in

  ```
  Error in file:/C:/3d-tiles/specification/README.md
  Not valid   : file:/C:/3d-tiles/specification/README.md#property-reference
  Anchors are [3d-tiles-format-specification, contents, introduction, file-extensions-and-media-types, json-encoding, uris, units, coordinate-reference-system-crs, concepts, tiles, tile-content, geometric-error, refinement, replacement, additive, bounding-volumes, region, box, sphere, content-bounding-volume, extensions, viewer-request-volume, transforms, tile-transforms, gltf-transforms, gltf-node-hierarchy, y-up-to-z-up, example, implementation-example, tile-json, tileset-json, external-tilesets, bounding-volume-spatial-coherence, spatial-data-structures, quadtrees, k-d-trees, octrees, grids, implicit-tiling, metadata, metadata-schema, assigning-metadata, metadata-statistics, specifying-extensions-and-application-specific-extras, extensions-1, extras, tile-format-specifications, declarative-styling-specification, license] for file:/C:/3d-tiles/specification/README.md
  LinkInfo:
      text                  : Property reference
      destination           : #property-reference
      destinationUri        : #property-reference
      scheme                : null
      fullUriWithFragment   : file:/C:/3d-tiles/specification/README.md#property-reference
      fullUriWithoutFragment: file:/C:/3d-tiles/specification/README.md
      fragment              : property-reference
      fileName              : 
  ```
  and I'll consider generalizing this for AscciDoc, but it's hard to make promises about the timeline here, so the manual approach may be more predictable.

- **Reducing the PDF size**
  The single-page PDF has a whopping 45MB right now, so it should be compressed. Unfortunately, the common tools that seem to be used for this task cause all internal links of the document to break. For details, refer to  https://bugs.ghostscript.com/show_bug.cgi?id=699830 . In doubt: Batch-processing all images and converting them to JPG with a limited size and maybe 70% quality should already reduce the size _significantly_


### Regarding the property references

As usual, the "core" was the 80% of the work that took 20% of the time. The JSON schema property reference is supposed to be generated with [wetzel](https://github.com/CesiumGS/wetzel), and quite some work will have to go into that. 

A short summary of what has to be done:

- **Multiple search paths**
  The 3D Tiles JSON schema is split into multiple directories. Support for multiple seach paths in wetzel is addressed in https://github.com/CesiumGS/wetzel/pull/71 . But there is a more fundamental issue with linking to the JSON schema files that have been found in these search paths (even if it is only a single one). It is not yet clear how the inner workings have to be restructured in order to support this
- **Document 'additionalProperties'**
  The output of wetzel usually is *"Additional properties are (not) allowed"*, without information about the _type_ of the properties. But for the [`class.schema.json` properties](https://github.com/CesiumGS/3d-tiles/blob/56d4d0302640682b2f3d2c31062d524ed87c7dbf/specification/schema/Schema/class.schema.json#L22), it is important to include this information (namely, that they have to be `class.property.schema.json`). This is addressed in https://github.com/CesiumGS/wetzel/pull/72 , but this is only a _draft_, and has to be compared to or aligned with https://github.com/CesiumGS/wetzel/pull/68 
- **Circular references and definitions**
  In the 3D Tiles specification, there are some references that wetzel does not handle correctlly - for example, the fact that `tile` refers to `tile` (via its `children`). This is addressed in https://github.com/CesiumGS/wetzel/pull/74 , but this will require a deeper restructuring of wetzel (details in the issue)
- **Linking and structure**
  The structure of the generated property references will have to be tweaked a bit. Right now, wetzel inserts `#anchors-for-cross-linking`, but these have to be disambiguated for the case that multiple property references should end up in a single HTML document. (This is partially addressed in https://github.com/CesiumGS/wetzel/pull/75 , but far from done). Also, there usually are "table of contents" inserted in the references, and these will have to be flattened and/or cross-linked for 3D Tiles. The path forward here is not yet clear.

<strike>If</strike> When the property references can be created as individual AsciiDoc files, then it should be easy to integrate them here. A first experiment was done in https://github.com/javagl/3d-tiles/commit/226a5bf00c618b6dcdfc66f2f1ea6fcc8892ef3a (not yet part of this PR), and there was no obvious blocker for this part until now.

javagl:
To elaborate on one aspect of the **Linking and structure** point mentioned above:

wetzel inserts links from the "Property Reference" to the JSON schema. For example, see glTF:

- glTF Property Reference: https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#properties-reference
- glTF JSON Schema Reference: https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#appendix-a-json-schema-reference

<sup>(An aside: Note that the Property Reference is **normative**, while the JSON schema is **not** normative!)</sup>

I assume that a similar solution should also be found for 3D Tiles. On a _very_ basic level, this already works:

Cesium 3DTilesSpec Links

This cross-linking takes place has some caveats. In the current state, wetzel generates this by writing out `include::schema/example.schema.json` statements into the AsciiDoc document. Actually resolving that (when generating the HTML) is left to AsciiDoc itself. The directory - `schema` in the example - is configured via a command-line parameter. So does not work out-of-the-box when the `schema` directory could actually be `schema/TileFormats`. For now, I took the approach of actually _inlining_ the schema files into the AsciiDoc document. But this has to be re-evaluated at some point. 

On top of that, all this raises a bunch of questions regarding links:

- How to avoid redundancies and duplicate link anchors when two - otherwise independent - documents refer to common things like `rootProperty`? (Or to put it that way: Where _exactly_ should `extras` and `extensions` appear in the HTML output?)
- If the property references are supposed to be hosted on GitHub as well, then they have to be different `.adoc` files for the HTML generation and for hosting on GitHub. In one case, the link anchors are local ones (inside the same document). In the other case, they are actual `link`s to external files (in complicated, relative directories...)

But I'll try to sort all that out...

javagl:
An update with the latest state:

### Regarding the core specification

Most of the recent changes have been link fixes (details in previous comments). One goal was to make sure that links work (in general... and) on GitHub as well as the PDF/HTML output. For example, one can now browse at https://github.com/javagl/3d-tiles/blob/asciidoc-conversion/specification/README.adoc and click the first link to `glTF Tile Format`, and actually end up at the right page, and this link *still* works in the single-page HTML file or the PDF document.

(It will take a few iterations until _all_ broken links are weeded out, but most of them should already work...)

### Regarding the property reference

A large refactoring of wetzel is in progress. In the current form, it has so many issues with the 3D Tiles spec that the **tl;dr** can only be: "It does not work (period)". Some early experiments with that less-then-ideal wetzel output are in https://github.com/javagl/3d-tiles/tree/asciidoc-conversion-experiments , but more work will have to go into that. 

---

### Preview:

The current state of the PDF is attached here. Note that this is a _compressed_ version, with low-resolution images. (The original one has 45MB, because of the large PNG files. Compressing that PDF is a pain in the back, and requires a carefully selected version of GhostScript - details are given in the build instructions)

[Specification-1.1.0-compressed.pdf](https://github.com/CesiumGS/3d-tiles/files/8710924/Specification-1.1.0-compressed.pdf)

javagl:
An updated state of the specification is attached here. This time with a Property Reference and a JSON Schema Refrence section.


---

Some notes about the property reference and wetzel:

The property reference is based on a local version of wetzel that does not have much in common with what is currently in the `wetzel` repository (c.f. https://en.wikipedia.org/wiki/Ship_of_Theseus ). If this is supposed to ever be made public, I'll have to do _lots_ of cleanups, but even then, it's not unlikely that the changes are not considered to be "improvements", and therefore, might never be merged. 

At some points, the 3D Tiles spec is already hitting a limit of which sort of documentation can be auto-generated from the schema. While most the browsing/linking should currently work - for example, linking from `b3dm.featureTable.BATCH_LENGTH` to the type `featureTable-definitions-globalPropertyInteger` - there are [types that simply have no names](https://github.com/CesiumGS/3d-tiles/blob/40d8a8b3efc5046fd020a49569b1a5865aa3a426/specification/schema/featureTable.schema.json#L148), and thus, can hardly be squeezed into a form that does _not_ boil down to a plain ADOC/MD version of _the JSON schema itself_. We might even consider changing the schema if we wanted to improve browsability here.

---

[Specification-1.1.0-compressed-2022-05-22.pdf](https://github.com/CesiumGS/3d-tiles/files/8749635/Specification-1.1.0-compressed-2022-05-22.pdf)

javagl:
I'll mark this as "ready for review" now. There certainly are still some issues (probably some links not working), or possible improvements (as in ~"That's not how you do this in AsciiDoc!"), but it should currently be a reasonably consistent state.

A summary of some possibly relevant points:

- There now is basically one `README.adoc` file for each former `README.md`
  - The original MDs are still there, to avoid breaking _external_ links. But GitHub displays the `.adoc` file by default, For example, when looking at the directory https://github.com/javagl/3d-tiles/tree/asciidoc-conversion/specification/TileFormats/PointCloud , then it shows the AsciiDoc file. 
- Basic instructions for building the specification are in [BUILDING.md](https://github.com/javagl/3d-tiles/blob/asciidoc-conversion/specification/BUILDING.md)
  - The "root" for building the specification is `Specification.adoc`. It includes all `README.adoc` files, as well as the `APPENDIX.adoc` and the "properties reference" that is generated with wetzel, and contained in `PropertiesReference_3dtiles.adoc` and `PropertiesReference_3dtiles_schema.adoc`

Previews of the resulting specifications are attached here. 

The PDF file was compressed (as described in the build instructions). For the HTML version, I downscaled/compressed the images, but the HTML file can just be dropped into the local `3dtiles/specification` subdirectory to see it with the original images. (I didn't want to upload 90MB here, due to a slow connection). 


[Specification-1.1.0-compressed-2022-06-26.pdf](https://github.com/CesiumGS/3d-tiles/files/8779862/Specification-1.1.0-compressed-2022-06-26.pdf)

[3d-tiles-Specification-HTML-compressedImages-2022-06-26.zip](https://github.com/CesiumGS/3d-tiles/files/8779869/3d-tiles-Specification-HTML-compressedImages-2022-06-26.zip)

specification/BUILDING.md
```diff
@@ -0,0 +1,110 @@
+<!-- omit in toc -->
```
lilleyse, line 1: Are `PropertiesReference_3dtiles.adoc` and `PropertiesReference_3dtiles_schema.adoc` auto-generated from wetzel? And will those instructions eventually go in here?

specification/Specification.adoc
```diff
@@ -0,0 +1,63 @@
...
+v1.1.0
+:toc2:
+:toclevels: 2
+:sectnums:
+:source-highlighter: rouge
+:rouge-style: github
+
+image:figures/3DTiles_dark_color_small.png[]
+
+This document describes the specification for 3D Tiles, an open standard for streaming massive heterogeneous 3D geospatial datasets.
```
lilleyse, line 12: It's unfortunate that the logo and intro text is split off into this separate document since it leaves `README.md` looking a little bare. But I can understand why it's done this way.

specification/README.adoc
```diff
@@ -0,0 +1,1209 @@
...
+
+* Tileset files should use the `.json` extension and the https://www.iana.org/assignments/media-types/application/json[`application/json`] Media Type.
+* Tile content files should use the file extensions and Media Type specific to their <<core-tile-format-specifications,tile format specification>>.
+* Metadata schema files should use the `.json` extension and the https://www.iana.org/assignments/media-types/application/json[`application/json`] Media Type.
+* Tileset style files should use the `.json` extension and the https://www.iana.org/assignments/media-types/application/json[`application/json`] Media Type.
+* JSON subtree files should use the `.json` extension and the https://www.iana.org/assignments/media-types/application/json[`application/json`] Media Type.
+* Binary subtree files should use the `.subtree` extension and the https://www.iana.org/assignments/media-types/application/octet-stream[`application/octet-stream`] Media Type.
+* Files representing binary buffers should use the `.bin` extension and https://www.iana.org/assignments/media-types/application/octet-stream[`application/octet-stream`] Media Type.
```
lilleyse, line 96: Bulleted lists now have extra spacing between rows. Is this something we can control in asciidoc?

bullet-spacing

lilleyse: Same observation for tables

table-spacing

specification/ACKNOWLEDGEMENTS.adoc
```diff
@@ -0,0 +1,34 @@
+= Acknowledgements
```
lilleyse, line 1: I noticed that Acknowledgements is not an appendix in the glTF spec: https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#acknowledgments

Might be good to understand why to see if we should do the same thing.

specification/Styling/README.adoc
```diff
@@ -0,0 +1,1496 @@
...
+* Regular expressions
+* Arrays of lengths other than 2, 3, or 4
+* Mismatched type comparisons (e.g. `1.0 === false`)
+* Array index out of bounds
+====
+
+[#styling-file-extension-and-mime-type]
+== File extension and MIME type
+
+Tileset styles use the `.json` extension and the `application/json` mime type.
```
lilleyse, line 1495: Not directly related to this PR but I noticed that some pages still use "MIME type" instead of "Media Type"

specification/PropertiesReference_3dtiles.adoc
```diff
@@ -0,0 +1,3262 @@
+
```
lilleyse, line 1: Could the property references be grouped together in a more consistent way, like tileset properties first, then tile format properties, then metadata - roughly based on the folder structure on github? I think it will be easier to follow that way.

property-reference-ordering

specification/directories.adoc
```diff
@@ -0,0 +1,40 @@
...
+// links between ADOC files in sibling directories can be resolved.
+ifdef::env-github[]
+:url-specification: ../
+:url-specification-implicittiling: {url-specification}ImplicitTiling/
+:url-specification-implicittiling-availability: {url-specification-implicittiling}/AVAILABILITY.adoc
+:url-specification-metadata: {url-specification}Metadata/
+:url-specification-metadata-referenceimplementation: {url-specification-metadata}ReferenceImplementation/
+:url-specification-metadata-referenceimplementation-propertytable: {url-specification-metadata-referenceimplementation}PropertyTable/
+:url-specification-metadata-referenceimplementation-schema: {url-specification-metadata-referenceimplementation}Schema/
+:url-specification-metadata-semantics: {url-specification-metadata}Semantics/
```
lilleyse, line 11: 3D Metadata Semantic Reference should probably be an official section rather than an appendex

image

specification/BUILDING.md
lilleyse, line 1: I noticed that the 3D Tiles logo was small in `Specification-1.1.0-compressed-2022-06-26.pdf`. I didn't see this problem when I generated the PDF myself.

small-logo

specification/BUILDING.md
```diff
@@ -0,0 +1,110 @@
...
+      -sDEVICE=pdfwrite ^
+      -dCompatibilityLevel=1.4 ^
+      -dPDFSETTINGS=/screen ^
+      -dNOPAUSE ^
+      -dBATCH ^
+      -dDetectDuplicateImages ^
+      -sOutputFile=Specification-1.1.0-compressed.pdf ^
+      Specification-1.1.0.pdf   
+    ```
+    (This is a Windows .BAT file. On Linux, replace the `^` with `\`)
```
lilleyse, line 60: My notes on the build process on Linux:

* I had to build Ghostscript from source since I couldn't find precompiled Linux binaries. I downloaded `ghostscript-9.20-linux-x86_64.tgz`, ran `./configure`, then `make`, and that created `bin/gs`.
* Have you seen this warning? I'm not sure which character it's refering to.

```
~/Code/3d-tiles/specification$ asciidoctor-pdf --verbose --trace --attribute scripts=cjk --attribute pdf-theme=default-with-fallback-font Specification.adoc --out-file Specification-1.1.0.pdf
asciidoctor: WARNING: Could not locate the character `⁣' (\u2063) in the following fonts: Noto Serif, M+ 1p Fallback, Noto Emoji
```

specification/BUILDING.md
```diff
@@ -0,0 +1,110 @@
...
+  - `asciidoctor-pdf --verbose --trace --attribute scripts=cjk --attribute pdf-theme=default-with-fallback-font Specification.adoc --out-file Specification-1.1.0.pdf`
+  
+    > Note: The last call is derived from https://asciidoctor.org/docs/asciidoctor-pdf/#support-for-non-latin-languages and handles Japanese characters. Without them, the call could just be `asciidoctor -r asciidoctor-pdf -b pdf Specification.adoc -o Specification-1.1.0.pdf`
+
+<sup>If the call does not appear to do anything, neither generate a PDF nor print an error message, make sure you typed `asciidoctor` and not just `asciidoc`</sup>
+
+
+### Compressing the PDF
+
+By default, the PDF file is large, due to the large image files being inserted uncompressed. There are tools for compressing the PDF, including a dedicated `asciidoctor-pdf-optimize` tool, but they have caveats.
```
lilleyse, line 40: After compression some of the diagrams become difficult to read.


Selection_375
Selection_376

The PDF you posted in https://github.com/CesiumGS/3d-tiles/pull/682#issuecomment-1138670614 doesn't seem to be nearly as bad. That one is 4.4MB whereas the one I generated was 2.3MB. I'm not sure what would explain the difference.

lilleyse (commented):
You've seen some stuff @javagl :smile: 

I did a high level pass to start. I plan to do a deeper pass later to help catch broken links or spot discrepancies between the .md and .adoc versions. The HTML output looks beautiful.

specification/BUILDING.md
javagl: These files indeed are the current output from (my local version of) wetzel. Similar files are generated by wetzel in its original `main` state, except for that it cannot properly produce them for 3D Tiles, of course.

There is a note at the top of `BUILDING.md` saying that the process of generating them with wetzel will be documented in the future (and maybe wrapped into some Docker container when necessary).

specification/Specification.adoc
javagl: When I started the conversion, I did a few experiments regarding the https://docs.asciidoctor.org/asciidoc/latest/document/header/ in AsciiDoc. There are some options and degrees of freedom, and I assume that for a document that is about to be submitted for standardization, there will be some sort of "title page template" (and we'll have to see whether this will just be "prefixed" to the PDF, or has to be integrated into the PDF that we generate).

To avoid the "bare" `README.adoc`, we could do something with 
```
ifdef::env-github[]
// include some image here
endif::[]
ifndef::env-github[]
// Don't :-) 
endif::[]
```

specification/README.adoc
javagl: I think that it (sh/c)ould be possible to tweak the underlying CSS, but I haven't looked into the details here yet - will take a note to do that

specification/ACKNOWLEDGEMENTS.adoc
javagl: This seems to be a bit arbitrary. From a quick (!) search, Vulkan does not seem to contain Acknowledgements at all. OpenCL seems to insert them at the top (before all sections, but unnumbered. apparently with the AsciiDoc `[[acknowdlegements]]` section type). I'll take a note to investigate that further.

specification/PropertiesReference_3dtiles.adoc
javagl: Right now, this is listed alphabetically by the "type name", e.g.
```
...
  templateUri (linked)
  tile (linked)
  tile.implicitTiling (linked)
  tileset (linked)
...
```
Of course, abritrary other orders can be established (e.g. sorting by the "Title" - i.e. the string that is actually displayed at the end). But when it comes to "grouping" and "folders", I'll have to see how the "desired sorting" can be squeezed into wetzel - beyond the `this.knownTypeNames = Object.keys(this.knownTypes).sort();` that it is doing now...

specification/directories.adoc
javagl: The reason why I moved it into the Appendix is that I thought that it should not be part of the _Specification of the format itself_. I saw it rather as a convention/vocabulary that is specific for the _application_ that consumes the data.

Of course, one could just move it into a section, and probably (?) tag that section as "non-normative", if this is desired...

Somehow, this is related to https://github.com/CesiumGS/3d-tiles/issues/643 ...

specification/BUILDING.md
javagl: Strange, I have some PDF where this was still the right size, but at some point, that chnaged - will take a note

specification/BUILDING.md
javagl: I saw that warning, and tried to figure out where it came from, but my friggin Windows console just displayed this as "�" and did not include the info that it was `\u2063`, so I couldn't properly search for this. I'll take a note for that as well, the Unicode code may be helpful to figure out what this actually is... "Invisible Separator" for now, but I'll have a look...

specification/BUILDING.md
javagl: The build instructions mention the compression modes. The one in the example is `screen`, which is _very_ agressive in terms of reducing the image resolution. The one that I used for the PDF was `ebook`, which may be more appropriate here.

(I did not yet try them out all, and I think they are only "shortcuts" for a bunch of other parameters, including _specific_ parameters that govern the DPI settings for images and the compression/quality tradeoff for images).

javagl:
I wrote some comments to the inlined ones, and summarize them as a short TODO list here:

- Consider adding a title image/text with `ifdef::env-github[]` into the main `README.adoc` (right now, these are only in `Specification.adoc`)
- Look at the CSS to reduce the spacing between bullet point lists and table rows
- Check whether 'Acknowledgements' should be a section or appendix. Investigate AsciiDoc `acknowledgments` section type
- Check if the properties reference can be sensibly "grouped" to be more similar to the folder structure
- MAYBE move "Metadata Semantic Reference" to its own (non-normative?) section. Also see https://github.com/CesiumGS/3d-tiles/issues/643
- Investigate why the 3D Tiles logo was small in some versions of the PDF. (Maybe just an artifact from of a missing `pdfwidth` and some DPI changes due to the compression or so...)
- Figure out what causes the warning `WARNING: Could not locate the character `⁣' (\u2063) ...` when creating the PDF
- Experiment with the PDF compression modes. `screen` is too aggressive and makes images unreadable. `ebook` (the one that the preview PDF was created from) seems OK. Check the lower-level options for limiting image sizes and quality.
- OT: Replace "MIME type" with "Media Type"

specification/Styling/README.adoc
javagl: Added this as a note at the "TODO list" comment below

specification/README.adoc
javagl: This _appears_ to primarily refer to the preview on GitHub. For that, there are some issues like https://github.com/asciidoctor/asciidoctor/issues/1087 , but they seem to boil down to "GitHub has to fix this" (i.e. there is no way to influence that preview).

For the HTML/PDF output, there are options for tweaking: 

- For PDF, the spacing can be tweaked with a YML file. This can "inherit" from the "default" style, and setting `outline_list: item_spacing: 1` yields a smaller spacing. 
- For HTML, the spacing can be tweaked with CSS. If somebody thinks that CSS is more useable than [Malbolge](https://en.wikipedia.org/wiki/Malbolge), feel free to edit [the default stylesheet](https://github.com/asciidoctor/asciidoctor/blob/main/src/stylesheets/asciidoctor.css) as desired, and I'll integrate it.

javagl:
An aside: I added the logo and an introductory sentence to the (GitHub-rendered version of the) `README.adoc` that is shown at https://github.com/javagl/3d-tiles/tree/asciidoc-conversion/specification . But we might consider to create a dedicated `README.md` for this directory. The current `README.adoc` that contains the actual specification text could then be something like `MAIN_SPECIFICATION.adoc` or so. 

This new "top-level" README.md could then be something like

---

### Specification

Hi there, look, all this is ADOC now, here's then entry point for 3D Tiles 1.1: [MAIN_SPECIFICATION.md](http://example.com)

Here are some PDF files: [Specification-1.1.0.pdf](http://example.com) ...

...

---

so that this directory/README.md rather serves as "_the_ entry point" for all spec-related documents (similar to https://github.com/KhronosGroup/glTF/tree/main/specification )

(Not a strong opinion, just a thought...)

specification/ACKNOWLEDGEMENTS.adoc
lilleyse: Ok, the current approach is fine then

specification/README.adoc
lilleyse: If there's no option for fixing the github preview then I think we'll just have to get used to it. No need to tweak the PDF/HTML output.

specification/Specification.adoc
lilleyse: Ah good idea!

specification/TileFormats/README.adoc
```diff
@@ -0,0 +1,55 @@
...
+| 3D model instances
+
+| xref:{url-specification-tileformats-pointcloud}README.adoc#tileformats-pointcloud-point-cloud[Point Cloud (`pnts`)]
+| Massive number of points
+
+| xref:{url-specification-tileformats-composite}README.adoc#tileformats-composite-composite[Composite (`cmpt`)]
+| Concatenate tiles of different formats into one tile
+|===
+
+See the See xref:{url-specification-tileformats-gltf-migration}#tileformats-gltf-migration-from-legacy-tile-formats[migration guide]. for further information about how these use cases can be modeled based on the glTF tile format.
```
lilleyse, line 55:
```suggestion
See the xref:{url-specification-tileformats-gltf-migration}#tileformats-gltf-migration-from-legacy-tile-formats[migration guide] for further information about how these use cases can be modeled based on the glTF tile format.
```

specification/Styling/README.adoc
javagl: Created https://github.com/CesiumGS/3d-tiles/pull/690 and https://github.com/CesiumGS/3d-tiles/pull/691 - when the first one is merged, I'll update this (`asciidoc-conversion`) branch with that state.

specification/BUILDING.md
javagl: I'll consider this as 'resolved' for now (using the `/ebook` settings, as described in the updated build instructions)

specification/BUILDING.md
javagl: Added instructitions for generating the properties reference to the build instructions. 
(If that docker thing is tackled at one point, this will be added accordingly)

lilleyse:
Let's skip the TOC for now.

lilleyse:
I noticed a lot of changes when I generated the property reference with wetzel locally. Does the property reference need to be re-generated now that https://github.com/CesiumGS/3d-tiles/pull/688 is merged?

javagl:
Indeed, the state here was generated based on a state without `BinaryBodyOffset`. Updated it with the latest state.

PR #684, opened 2022-05-17
Draft 1.1 external tileset class clarification

javagl:
This only adds the hint

> Any class that is defined in the schema of an external tileset must also be defined in the top-level tileset schema.

in the main 'Metadata Schema' section, and thus, reflects the change from https://github.com/CesiumGS/cesium/pull/10387

If desired, the [schema description](https://github.com/CesiumGS/3d-tiles/blob/1c1b792c15f8f4c42a375d3822e78f5b57567259/specification/schema/metadataEntity.schema.json#L10) could also be extended by

> The class that property values conform to. The value must be a class ID declared in the `classes` dictionary of the metadata schema **of the *`<right>`* tileset** 

Where *`<right>`* could be "containing" or "root/top-level". 

<sup>There are some questions that could or should be investigated further, about namespacing or disambiguation (e.g. related to https://github.com/CesiumGS/cesium/issues/10085 ), but for now, this change is only an attempt to clarify the structure that is expected by CesiumJS</sup>


javagl:
Note to myself: Don't forget to mirror this change in https://github.com/CesiumGS/3d-tiles/pull/682 ...

lilleyse:
Thanks @javagl

javagl:
Mirrored in https://github.com/javagl/3d-tiles/commit/492be698e03657197491448995c4b85836aa5686

PR #688, opened 2022-05-25
Introduce BinaryBodyOffset as a defined type

javagl:
Certain [type definitions in `featureTable.schema.json`](https://github.com/CesiumGS/3d-tiles/blob/39ff902c75bd3620ac83e5327c40ea5b198e8315/specification/schema/TileFormats/featureTable.schema.json#L91) referred to multiple types with `oneOf`, where the first type did not have a name. In additon to causing trouble for generating the property reference with wetzel, this "unnamed type" was repeated several times. Now, this unnamed type is called `BinaryBodyOffset`.

---

Property reference before:
```
[#reference-featuretable-definitions-globalpropertyinteger]
=== featureTable#/definitions/globalPropertyInteger

An object defining a global integer property value for all features.

* **Type**: One of `object`, `integer`

* **Additional properties are allowed.**

```

Property reference after:
```
[#reference-featuretable-definitions-globalpropertyinteger]
=== featureTable-definitions-globalPropertyInteger

An object defining a global integer property value for all features.

* **Type**: One of <<reference-featuretable-definitions-binarybodyoffset,`featureTable-definitions-binaryBodyOffset`>>, `integer`

* **Additional properties are allowed.**
```
where the link leads to the definition...
```
[#reference-featuretable-definitions-binarybodyoffset]
=== featureTable-definitions-binaryBodyOffset

An object defining the offset into a section of the binary body of the features table where the property values are stored if not defined directly in the JSON.
...
```
The output is still not perfect. I'm playing "whack-a-mole" with wetzel right now. But this change should be independent of that, and at least _allow_ this sort of linking, instead of using `object`.

---

This seemed to be the only "unnamed type" for now. There are some properties that are still listed as `object`. Namely the ones where type constraints are established with `additionalProperties` -  for example, [`statistics.classes`](https://github.com/CesiumGS/3d-tiles/blob/39ff902c75bd3620ac83e5327c40ea5b198e8315/specification/schema/Statistics/statistics.schema.json#L8). One could make a case for giving this a name as well. But right now, these are documented with a line
```
    * **Type of each property**: <<reference-statistics-class,`statistics.class`>>
```
which should be fine here.

### DIFF
--- a/specification/schema/TileFormats/featureTable.schema.json
+++ b/specification/schema/TileFormats/featureTable.schema.json
@@ -9,16 +9,26 @@
         "$ref": "#/definitions/property"
     },
     "definitions": {
-        "binaryBodyReference": {
-            "title": "BinaryBodyReference",
+        "binaryBodyOffset": {
+            "title": "BinaryBodyOffset",
             "$ref": "rootProperty.schema.json",
-            "description": "An object defining the reference to a section of the binary body of the features table where the property values are stored if not defined directly in the JSON.",
+            "description": "An object defining the offset into a section of the binary body of the features table where the property values are stored if not defined directly in the JSON.",
             "properties": {
                 "byteOffset": {
                     "type": "integer",
                     "description": "The offset into the buffer in bytes.",
                     "minimum": 0
-                },
+                }
+            },
+            "required": [
+                "byteOffset"
+            ]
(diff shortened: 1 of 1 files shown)


lilleyse:
Thanks @javagl

PR #695, opened 2022-06-18
Fix extension name in description in subtree schema

javagl:
On today's episode of "Things that people wouldn't notice anyhow": The extension was referred to as `EXT_multiple_contents` but is called `3DTILES_multiple_contents`.


lilleyse:
Good catch!

Issue #703, opened 2022-07-19
Update reference card for 3D Tiles 1.1


javagl:
One point of https://github.com/CesiumGS/3d-tiles/issues/651 that had still been on my plate (and been waiting there for too long) is the update of the reference card for 3D Tiles 1.1. To avoid filling the repo with commits of SVG (or even worse, binaries like PDFs), certain points can probably be sorted out here.

I assume that the result should be a dedicated reference card for the new features of 3D Tiles 1.1. Specifically, I assume that it should _not_ combine the reference card of 1.0 and the reference card for 'Next'. Instead, it should rather be an update of the former 'Next' reference card. 

Two broader, structural questions:

- Where/how to present the S2 extension? 
  It could be a dedicated section at the end, to make clear that it is not part of "3D Tiles (core) 1.1". Or it might be omitted in the reference card altogether: It only was a short section to _present_ the extension, but in terms of contents, it was not adding much value beyond the main S2 README. 

- Where/how to present the glTF extensions? 
  There already is some confusion, and people think that the glTF extensions are somehow "part of" Next/1.1. Also, the glTF extensions have been split into `mesh_features` and `structural_metadata` in the meantime. One could mention `structural_metadata` as one implementations of the 3D Metadata Specification. But this should be done carefully, with a disclaimer: Technically, this extension is _not_ part of 3D Tiles. 
  For now, I think that it could make sense to omit the page "The EXT_mesh_features Extension" from the 'Next' reference card. This could be moved into a short (1-2 pages) but dedicated overview of the proposed `mesh_features` and `structural_metadata` extensions.

lilleyse:
> I assume that the result should be a dedicated reference card for the new features of 3D Tiles 1.1. Specifically, I assume that it should _not_ combine the reference card of 1.0 and the reference card for 'Next'. Instead, it should rather be an update of the former 'Next' reference card. 

So it would be like a "what's new in 3D Tiles 1.1?" reference card? Even though I wasn't thinking that originally that's probably the right thing to do.

> Where/how to present the S2 extension?

It would be fine to omit

> Where/how to present the glTF extensions?

There could be a separate resource that describes all the recommended glTF extensions. Which is reminiscent of the glTF geospatial profile: https://github.com/KhronosGroup/glTF/issues/2092

javagl:
[...]

PR #706, opened 2022-07-26
Update reference card for 1.1

javagl:
- Added reference card PDF for 1.1
- Created subfolder for reference cards
- Moved sources of 1.0 reference card into '1.0' subfolder
- Added sources of 1.1 reference card in '1.1' subfolder
- Updated READMEs and reference card images.

Fixes https://github.com/CesiumGS/3d-tiles/issues/703


lilleyse:
Excellent, thanks @javagl!

PR #710, opened 2022-08-15
Clarifications for bounding volumes and transforms

javagl:
Added 'Informative' sections explaining or justifying

- the use of _radians_ in bounding regions
- the representation of bounding boxes
- the column-major storage of the tile `transform`


lilleyse:
Thanks @javagl!

Issue #711, opened 2022-08-24
Clarifications and additions for the specification


javagl:
> **Edit 2023-01-11**:  Most of the points mentioned here have been addressed in https://github.com/CesiumGS/3d-tiles/pull/726#issuecomment-1370114018 - see the next comment for an updated state summarizing the points that are not addressed yet.

Possible constraints and clarifications to add to the schema and specification:

For [`boundingVolume`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/boundingVolume.schema.json):

- The `latitude` and `longitude` for the `region` bounding volume should be constrained to the ranges that are mentioned in https://spatialreference.org/ref/epsg/4979/ (-180.0000, -90.0000, 180.0000, 90.0000)
- The minimum height should not be larger than the maximum height
- The `south` may not be greater than the `north`
- The `radius` of the bounding sphere should not be negative

For [`properties`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/properties.schema.json) (although deprecated?)

- The `maximum` should not be smaller than the `minimum`

For [`statistics`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Statistics/statistics.schema.json):

- The `classes` should be required
- The desciption says *"A dictionary, where each key corresponds to a class ID in the `classes` dictionary and..."* - this is confusing. It should say *"... in the `classes` dictionary **of the metadata schema of the containing tileset**, and ..."*
- Similarly the `statistics.class.properties` description should refer to the _class of the metadata schema_

For [`statistics.class.property`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Statistics/statistics.class.property.schema.json#L8):

- One **could** add constraints (like `!(min>max)`) here. But one could go pretty far, e.g. `!(mean>min)`, and even make mathematically based constraints about the possible values of the standard deviation and such. Maybe this is a step too far for now (there are more important constraints/checks to be added)

For [`class.property`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Schema/class.property.schema.json#L94):

[...]

Issue #598, opened 2021-11-23
Limiting the number of files in a directory for implicit tiling


javagl:
The [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_implicit_tiling) specification allows defining quadtree- or octree tilesets that only exist _implicitly_. Each of these implicit tiles may have associated `tile.content`. This tile  content is stored in a file that is referred to with a template URI. This template URI includes variables for the tile coordinates (`level`, `x`, `y`, (`z`)).

For example, an implicit quadtree tileset may refer to `b3dm` content that is stored under the template URI `"content/{level}/{x}/{y}.b3dm"`.

The number of content files that may end up in one directory may be large. This is true _even when_ they are distributed into mutliple directory levels, by creating subdirectories for each level and each coordinate, as shown in the example above. The size of along one dimension, for a given level, is 2<sup>level</sup>. This means that for an implicit tileset with 16 levels, there may be content files

    content/0/0/0.b3dm
    ....
    content/0/0/65535.b3dm

Having so many files in one directory may not be possible at all for certain file systems, or cause severe performance penalties on others (see https://stackoverflow.com/q/466521 for some details).

There should be an alternative for addressing the content files of large implicit tilesets, beyond the template URIs that are currently allowed.

donmccurdy:
Loosely related suggestion in https://github.com/CesiumGS/3d-tiles/issues/553#issuecomment-971988626, loading multiple tile contents from a single file like `uri: lod0.glb#15`.

It is a bit regrettable to complicate the specification in order to support the specific combination of (1) an old filesystem, and (2) a server with no distinction between URLs and paths on disk. Cloud object storage (Amazon S3, Google Cloud Storage) is a great way to host large tilesets and has no such limitations. Other complexities exist in URL resolution (e.g. authentication, temporary tokens) which are left out of scope for good reason.

jerstlouis:
I agree that a distinction between paths on disk vs. URL paths should be made.

Regarding storing on disk, one file per tile might easily be too much overhead See the [CDB 2 proposal](https://github.com/sofwerx/cdb2-eng-report/blob/master/11-tiling-coverages.adoc#proposed-cdb-x-data-container) to handle arbitrarily large datasets by grouping multiple tiles within GeoPackages as tile pyramids based on LOD, but still splitting over multiple GeoPackages for repository use cases.

[...]

PR #726, opened 2023-01-02
Clarifications for the 1.1 specification

javagl:
Addressing https://github.com/CesiumGS/3d-tiles/issues/711 : I'll replicate the bullet point list here, and mark the parts that have already been solved with checkboxes.

- [x] For [`boundingVolume`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/boundingVolume.schema.json):

  - The `latitude` and `longitude` for the `region` bounding volume should be constrained to the ranges that are mentioned in https://spatialreference.org/ref/epsg/4979/ (-180.0000, -90.0000, 180.0000, 90.0000)
    - **NOTE**: It's actually given in radians, deviating from the EPSG spec. Emphasized this in the text.
  - The minimum height should not be larger than the maximum height
  - The `south` may not be greater than the `north`
  - The `radius` of the bounding sphere should not be negative

- [x] For [`properties`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/properties.schema.json) (although deprecated?)

  - The `maximum` should not be smaller than the `minimum`

- [x] For [`statistics`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Statistics/statistics.schema.json):

  - <strike>The `classes` should be required</strike> Based on internal discussion, it should remain optional
  - [x] The desciption says *"A dictionary, where each key corresponds to a class ID in the `classes` dictionary and..."* - this is confusing. It should say *"... in the `classes` dictionary **of the metadata schema of the containing tileset**, and ..."*
  - [x] Similarly the `statistics.class.properties` description should refer to the _class of the metadata schema_

- [ ] For [`statistics.class.property`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Statistics/statistics.class.property.schema.json#L8):

  - One **could** add constraints (like `!(min>max)`) here. But one could go pretty far, e.g. `!(mean>min)`, and even make mathematically based constraints about the possible values of the standard deviation and such. Maybe this is a step too far for now (there are more important constraints/checks to be added)

- [ ] For [`class.property`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Schema/class.property.schema.json#L94):

  - [x] The `enumType` should be disallowed for non-`"ENUM"` types
  - [x] I think the `componentType` should be required for the SCALAR/VECn/MATn types.
  - [ ] One could go full pedantic mode, and require `min`/`max` to be integer values when the component type is integral, and require the bounds of the values to be obeyed. For a `UINT8` (without `offset` or `scale`), a `max=0.8` does not make sense, and neither does a `min=-10000`.
  - [x] The `min/max/offset/scale` do not make sense for variable length arrays. This is mentioned at https://github.com/CesiumGS/3d-tiles/tree/draft-1.1/specification/Metadata#minimum-and-maximum-values but could also be mentioned in the schema
 
- [ ] For [`templateUri`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/templateUri.schema.json#L6)

  - One could consider to add the requirement that the variables `{level}`, `{x}`, `{y}`, `{z}` MUST be present for the respective tree types. For now, I assume that any `{unknown}` expression is an error, and "missing" ones (e.g. a missing `{z}` for an `OCTREE`) should cause a warning.

- [ ] For [`subtree`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Subtree/subtree.schema.json#L16)

  - I thought that one _could_ add the constraint that buffer views may not overlap, so that
    ```
    { byteOffset: 0, byteLength: 10 }
    { byteOffset: 5, byteLength: 5 }
    ```
    was disallowed. But I don't see a _strong_ reason to disallow that (and glTF apparently _does_ allow this as well...)

- [ ] For [`availability`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Subtree/availability.schema.json#L13)

  - [ ] One could add a constraint that `availableCount` may not be larger than `bufferView.byteLength/8` (?)
  - [x] The [Content Availability](https://github.com/CesiumGS/3d-tiles/blob/draft-1.1/specification/ImplicitTiling/README.adoc#content-availability) section could be a bit clearer about multiple contents (it only refers to `tile.content` now, and does not mention `tile.contents`)
  - [x] We could add a rule to say that trailing (unused) bits in an availability buffer should always be `0`

- [ ] For [`tile`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/tile.schema.json#L35) : 

  - One could consider to add constraints to the `transform`. From "weak" to strict: It should probably not be zero. It should probably be invertible. It should probably have a positive determinant. It should probably not contain skew/shear components. It should probably be decomposable into translation, rotation, scale components. See https://registry.khronos.org/glTF/specs/2.0/glTF-2.0.html#transformations for inspiration.
  - The conditions for when `content`/`contents` MUST be present are difficult. One could say that one of them MUST be present when `implicitTiling` is present, but ... strictly speaking, only when there is any `contentAvailability[i]=true`. 

- [x] For [`schema`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Schema/schema.schema.json#L8)

  - [x] The `schema.id` is marked as 'required' and must match the usual "ID regex". The ID may eventually be used (e.g. by CesiumJS) to uniquely identify properties in Custom Shaders, or to identify that the `schema` objects of two loaded tilesets are _the same_ schema. But the spec should (at least as an 'Implementation Note') describe the purpose, and give usage hints - for example, whether it should only be a semi-random "unique ID", or a human-readable string, or maybe something that includes a concept of "versioning".

- [x] For [`tileset`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/tileset.schema.json#L54): 
  - One could mention that each element of `extensionsRequired` must also appear in `extensionsUsed`

- [ ] Structural:

  - It should be clarified under which conditions certain constraints are condidered to be satisfied. For example, when a tile defines a `geometricError` that is _larger_ than the `geometricError` of the parent, **but** overrides this geometric error with metdata so that it is _smaller_ than that of the parent, should this cause a validation issue? 

<strike>
- [ ] For [`subtree`](https://github.com/CesiumGS/3d-tiles/blob/aafebe993986aa9c668d35a63a42cc72ccdcc327/specification/schema/Subtree/subtree.schema.json#L16)

  - The specification should mention that for _binary_ subtrees, only **one** `buffer` may omit the `uri` property (only one buffer may refer to the binary chunk)
    - This is mentioned in the text at https://github.com/CesiumGS/3d-tiles/blob/draft-1.1/specification/ImplicitTiling/README.adoc#buffers-and-buffer-views : 
      > In the Binary Format the first buffer may instead refer to the binary chunk of the subtree file, in which case the uri property shall be undefined. 
</strike>

- [ ] Regarding `extensionsUsed`: 

  - The spec says at https://github.com/CesiumGS/3d-tiles/tree/draft-1.1/specification#extensions-1 
    > All extensions used in a tileset or any descendant external tilesets shall be listed in the entry tileset JSON 

    This should be reviewed and possibly clarified. If one tileset refers to an external tileset that does *not* use any extensions, but is later modified to use a certain extension, then the containing tileset may become "invalid", because it does not declare this extension as `extensionsUsed`. While this makes sense from the perspective of _knowing which extensions will be used as a whole_, it somewhat contradicts the idea of external tilesets being "building blocks" that can be combined arbitrarily.

### DIFF
--- a/specification/ImplicitTiling/README.adoc
+++ b/specification/ImplicitTiling/README.adoc
@@ -261,8 +261,6 @@ Each subtree contains tile availability, content availability, and child subtree
 
 Each type of availability is represented as a separate bitstream. Each bitstream is a 1D array where each element represents a node in the quadtree or octree. A 1 bit indicates that the element is available, while a 0 bit indicates that the element is unavailable. Alternatively, if all the bits in a bitstream are the same, a single constant value can be used instead.
 
-The bitstream is encoded in bytes that are stored in a <<implicittiling-buffers-and-buffer-views,buffer>>. Trailing bits that result from the number of nodes not being divisible by 8 shall be set to 0. 
-
 To form the 1D bitstream, the tiles are ordered with the following rules:
 
 * Within each level of the subtree, the tiles are ordered using the https://en.wikipedia.org/wiki/Z-order_curve[Morton Z-order curve]
@@ -404,7 +402,8 @@ Content availability (`contentAvailability`) is an array of content availability
 
 Availability may be represented either as a bitstream or a constant value. `bitstream` is an integer index that identifies the buffer view containing the availability bitstream. `constant` is an integer indicating whether all of the elements are available (`1`) or all are unavailable (`0`). `availableCount` is an integer indicating how many `1` bits exist in the availability bitstream.
 
-Availability bitstreams are packed in binary using the format described in the xref:{url-specification-metadata}README.adoc#metadata-booleans[Booleans] section of the 3D Metadata Specification.
+Availability bitstreams are packed in binary using the format described in the xref:{url-specification-metadata}README.adoc#metadata-booleans[Booleans] section of the 3D Metadata Specification. Trailing bits of availability bitstreams that result from the number of nodes not being divisible by 8 shall be set to 0.
+
 
 [NOTE]
 .Example
--- a/specification/ImplicitTiling/README.adoc
+++ b/specification/ImplicitTiling/README.adoc
@@ -261,6 +261,8 @@ Each subtree contains tile availability, content availability, and child subtree
 
 Each type of availability is represented as a separate bitstream. Each bitstream is a 1D array where each element represents a node in the quadtree or octree. A 1 bit indicates that the element is available, while a 0 bit indicates that the element is unavailable. Alternatively, if all the bits in a bitstream are the same, a single constant value can be used instead.
 
+The bitstream is encoded in bytes that are stored in a <<implicittiling-buffers-and-buffer-views,buffer>>. Trailing bits that result from the number of nodes not being divisible by 8 shall be set to 0. 
+
 To form the 1D bitstream, the tiles are ordered with the following rules:
 
 * Within each level of the subtree, the tiles are ordered using the https://en.wikipedia.org/wiki/Z-order_curve[Morton Z-order curve]
--- a/specification/ImplicitTiling/README.adoc
+++ b/specification/ImplicitTiling/README.adoc
@@ -296,7 +296,7 @@ image::figures/tile-availability.png[Tile Availability]
 [#implicittiling-content-availability]
 ==== Content Availability
 
-Content availability determines which tiles have a content resource. The content resource is located using the `content.uri` template URI. If there are no tiles with a content resource, `tile.content` shall be omitted.
+Content availability determines which tiles have a content resource. The content resource is located using the template URI of the tile content. When the tile has xref:{url-specification}README.adoc#core-tile-content[multiple contents], then there is one content availability bitstream for each content. If there are no tiles with a content resource, then `tile.content` and `tile.contents` shall be omitted.
 
 Content availability has the following restrictions:
 
--- a/specification/Metadata/README.adoc
+++ b/specification/Metadata/README.adoc
@@ -70,6 +70,13 @@ Components of a schema are listed below, and implementations may define addition
 
 IDs (`id`) are unique <<metadata-identifiers,identifiers>> for a schema.
 
+[NOTE]
+.Informative
+====
+The schema ID is _required_ for each schema. The main purpose of this ID is to be able to resolve possible ambiguities. The exact mechanism for this disabiguation depends on the client application, and how the access to metadata properties is implemented. But the schema ID makes sure that a compound identifier of the form `<schema-id>.<class-name>.<property-name>` is globally unique, even when the same class name and property name appears in multiple schemas.
+====
+
+
 [#metadata-version]
 ==== Version
 
--- a/specification/README.adoc
+++ b/specification/README.adoc
@@ -145,7 +145,7 @@ The CRS of a tileset may be defined explicitly, as part of the <<core-metadata,t
 
 An additional <<core-tile-transforms,tile transform>> may be applied to transform a tile's local coordinate system to the parent tile's coordinate system.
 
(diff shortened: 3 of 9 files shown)


javagl:
A summary of the points that are **not** ticked:

----

For `statistics.class.property`... One could add constraints (like `!(min>max)`)
 - There are some constraints that could easily be added and checked. But it's hard to say which constraints could be justified mathematically.

---

For `class.property`: One could go full pedantic mode, and require `min`/`max` to be integer values when the component type is integral, and require the bounds of the values to be obeyed. For a `UINT8` (without `offset` or `scale`), a `max=0.8` does not make sense, and neither does a `min=-10000`.
- I think that this _could_ make sense. But I'd tend to tackle this as its own issue/PR, if desired. Taking the `offset` and `scale` into account makes this not entirely trivial. One would have to write it in a form like
  > The `max` may not be larger than `maxDataTypeValue*scale+offset` ...

   with proper specification of what `maxDataTypeValue` is...

---

For `templateUri` ... One could consider to add the requirement that the variables `{level}`, `{x}`, `{y}`, `{z}` MUST be present for the respective tree types.
  - I'm hesitating to constrain this further for now, in view of the possible alternatives for template URIs that may be required in the context of https://github.com/CesiumGS/3d-tiles/issues/598 . For example, **iff** we, say, define an extension that allows a variable like `{mortonIndex}`, then the other variables would be unused...

---

For subtree ... I thought that one could add the constraint that buffer views may not overlap

- There doesn't seem to be a strong reason to add that requirement

---

For `availability` ... One could add a constraint that `availableCount` may not be larger than `bufferView.byteLength/8` (?)

- That doesn't seem to be necessary, in hindsight.

---

For `tile` ... _(see checklist)_

- I think that adding constraints for the `transform` could make sense. If this should be addressed, it should be tracked in a dedicated issue. 
- About the presence of `content` or `contents`: Right now, the implicit tiling section says that `tile.content` and `tile.contents` shall be omitted when there is no `contentAvailability[x]===1`. It seems natural to require it to be present when there _is_ a `1`-bit in the bitstream. Should this be added in the spec text and the `tile.content` schema? More importantly: Are there other (less tangible or obvious) conditions under which tile content _must_ be present? 

---

Structural: ...

- Somehow related to the second point of the `tile` considerations (above): It's sometimes hard to pin down how _semantics_ affect the _validity_ of an asset. This might be a broader question, once on the spec level, and once on the level of how the validator treats _semantics_. 


---

Regarding extensionsUsed: ...

- This is, to a reasonable extent, tracked in https://github.com/CesiumGS/3d-tiles-validator/issues/231 (which may affect details of the wording in the specification, but probably only on the level of 'implementation notes')

lilleyse:
Looks good. Thanks @javagl.

PR #728, opened 2023-01-03
Add `TILESET_TILE_COUNT` semantic

lilleyse:
Adds a new tileset semantic, `TILESET_TILE_COUNT`, which is the total number of tiles in the tileset, including empty tiles and tiles from external tilesets.

Knowing the tile count in advance can be a useful optimization hint, e.g. for preallocating data textures for [voxels](https://github.com/CesiumGS/cesium/pull/10996) in CesiumJS.


javagl (approved):
Seems fine for me.

PR #730, opened 2023-01-20
Fix conflict in `BinaryBodyReference.componentType`

j9liu:
While looking through the specs I found a typo that says `BinaryBodyReference.componentType` should be required, even though it is not.


lilleyse:
Thanks @j9liu!

PR #731, opened 2023-01-30
Fix broken link to decodeDraco.js file

martin-schilling:


lilleyse:
Thanks @martin-schilling, looks like this was fixed in https://github.com/CesiumGS/3d-tiles/pull/740.

PR #733, opened 2023-02-03
Additions to Point Cloud spec

j9liu:
The Point Cloud guide doesn't mention anything about its colors being stored in sRGB color space. This leads to unexpected rendering differences when migrating from `pnts` to glTF, because a glTF stores colors in linear RGB.

To minimize future confusion, I added some lines about point clouds storing colors in sRGB, and how they must be converted to linear RGB for glTFs.

I also added a line requiring `POINTS_LENGTH` to be greater than zero, as suggested in https://github.com/CesiumGS/3d-tiles/issues/711#issuecomment-1387331864.

@javagl could you review? Feel free to rewrite my lines if they are not clear enough.

### DIFF
--- a/specification/TileFormats/PointCloud/README.adoc
+++ b/specification/TileFormats/PointCloud/README.adoc
@@ -175,7 +175,7 @@ These semantics define global properties for all points.
 
 | `POINTS_LENGTH`
 | `uint32`
-| The number of points to render. The length of each array value for a point semantic should be equal to this.
+| The number of points to render. This must be greater than zero. The length of each array value for a point semantic should be equal to this.
 | Yes.
 
 | `RTC_CENTER`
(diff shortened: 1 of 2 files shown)


javagl:
Looks fine for me. I'll read a little bit more about the context (and what e.g. CesiumJS is doing with the vertex colors), and probably merge soon. 

Interstingly, a quick search for `srgb` does not bring up _any_ mention in this repo. And websearching for `3d+tiles+"srgb"` brings up a [fairly recent presentation](https://www.khronos.org/assets/uploads/developers/presentations/Scaling_glTF_with_3D_Tiles.pdf) with the lines

> Point clouds
>
> Vertex color sRGB (?)

so I guess that "(?)" may become a "(!)" soon...)

javagl (approved):
I looked a little bit at the handling of the color spaces in CesiumJS, and it seems like there is a dedicated step for converting the sRGB colors from point clouds into the linear color space. One could zoom pretty deeply into that, and ask questions about the 'Styling' section of the specification, but that may be unrelated to this PR.

PR #745, opened 2023-06-08
Add matterport to resources.md

dbuck:
Matterport has been creating and serving 3d-tiles/tilesets for all scanned spaces since late 2022, thanks! At many millions of weekly views, maybe it is time to add a link to resources?


javagl:
Indeed, I was aware of matterport and some applications that use their data, and _thought_ that it already was contained in the resources list, but ... it wasn't. Thanks!

Issue #752, opened 2023-10-09
Inconsistency in tile schema and its description


javagl:
The description of the `tile.children` property at https://github.com/CesiumGS/3d-tiles/blob/8e5e67e078850cc8ce15bd1873fe54f11bbee02f/specification/schema/tile.schema.json#L84 says

> [...] For leaf tiles, the length of this array is zero, and children may not be defined.

But at the same time, the schema says `"minItems": 1`.

PR #753, opened 2023-10-16
Fix wording in description for tile children

javagl:
Fixes https://github.com/CesiumGS/3d-tiles/issues/752

OLD: "For leaf tiles, the length of this array is zero, and children may not be defined"
NEW: "For leaf tiles, there are no children, and this property may not be defined"

### DIFF
--- a/specification/PropertiesReference_3dtiles.adoc
+++ b/specification/PropertiesReference_3dtiles.adoc
@@ -2259,12 +2259,12 @@ A dictionary object of metadata about per-feature properties.
 
 |**maximum**
 |`number`
-|The maximum value of this property of all the features in the tileset. The maximum value shall not be larger than the minimum value.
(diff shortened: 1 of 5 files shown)


lilleyse:
There are a few other occurrences of "length of this array is zero" that need to be fixed

image

javagl:
These are the parts that are auto-generated from the schema. I wasn't sure whether this should be part of this PR, but I can update them as well.

javagl:
The auto-generated files apparently had not been updated for a while, e.g. after https://github.com/CesiumGS/3d-tiles/commit/c32e4cf25dc094c555e78429eec82ff6d387101e - so it's probably good to have them on the latest state here again.

javagl:
While looking at the diff of the generated output, I noticed that there was a wrong description for the `properties.minimum/maximum` properties. I hope it's OK to sneak that fix in here...

Issue #758, opened 2024-01-16
Can Map Data Explorer iOS and Android be added as Viewer - it supports 3DTILES


techmavengeospatial:
Map Data Explorer iOS and Android
https://mapexplorer.techmaven.net
supports 3DTILES online and offline

Issue #757, opened 2024-01-16
Can Earth Explorer 3D Map with Augmented reality be added as Viewer?


techmavengeospatial:
https://earthexplorer.techmaven.net
is based on fork of Cesium terriajs and works online/offline
Supports 3DTILES URL's and offline 3TZ archive format or .3DTILES zip archives
and CesiumION Assets
Supports offline map tiles from GPKG and MBTILES
Supports offline/online static GIS Files
Supports Mapping Services from ESRI and OGC

Issue #761, opened 2024-04-29
Basic, minimal examples; sources for README figures


andreasplesch:
We are considering implementing a viewer for OGC 3D Tiles 1.1 based on an existing glTF renderer and existing geospatial support functionality.

Looking for the most basic example data set, there is 

https://github.com/CesiumGS/3d-tiles/blob/main/RESOURCES.md#sample-data

which test certain features of 3D Tiles. However, it would be really helpful to have an even more basic set of 3D Tiles which only uses core functionality, eg. no extensions (for 3D Tiles or gtTF), no styles or other optional features.

I think the introductory figures in https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.adoc#tiles show such very basic tiles, essentially starting with a box located with coordinates. So if it is possible to share the .json and content which was used to generate these figures, that could be very helpful if these are not available here or elsewhere ?

Thanks for any feedback or pointers.

javagl:
Most of the samples from the "Sample Data" section try to demonstrate specific features of 3D Tilles. When it comes to even more basic test data, one could go very far in terms of the "exhaustiveness" that should be tested. One could start with a complete and valid 3D Tiles tileset....
```json
{
  "asset" : {
    "version" : "1.1"
  },
  "geometricError" : 2.0,
  "root" : {
    "refine": "REPLACE",
    "boundingVolume" : {
      "box" : [ 0.5, 0.5, 0.5, 0.5, 0.0, 0.0, 0.0, 0.5, 0.0, 0.0, 0.0, 0.5 ]
    },
    "geometricError" : 1.0
  }
}
```
and from that, explore the space that is opened up by the combinatorial explosion on the lowest technical level.

Something that should be far more **useful** for implementors and tests: The [`Specs/Data/Cesium3DTiles`](https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Cesium3DTiles) directory of CesiumJS contains a large variety of "real" (but usually simple) 3D Tiles data sets. These are used for the Specs (unit tests) of the CesiumJS 3D Tiles support. And it should include most of the data sets that have been used for the screenshots in the specification.

For example, the tileset that is shown in the section that you linked to is (the same or similar to) the one that is contained in https://github.com/CesiumGS/cesium/tree/main/Specs/Data/Cesium3DTiles/Tilesets/TilesetRefinementMix 

[...]

PR #762, opened 2024-04-30
Update RESOURCES.md, 2024-04

javagl:
Fixes https://github.com/CesiumGS/3d-tiles/issues/757 
Fixes https://github.com/CesiumGS/3d-tiles/issues/758

- Add Earth Explorer 3D
- Add Map Data Explorer
- Add Oracle Spatial Studio
- Add ArcGIS Maps SDK for Native Apps
- Add a link to the CesiumJS spec files as one source of 'sample data' (and a link to the samples generator - this is otherwise not listed prominently, because it is not really maintained, but could be interesting for those who want to know where the spec files once came from...)


andreasplesch:
Reference to #761

PR #778, opened 2024-09-02
Add open-data section just below sample-data

jo-chemla:
For sharing open-data 3d-tiles tilesets

### DIFF
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -28,6 +28,7 @@
 - [Earth Explorer 3D](https://earthexplorer.techmaven.net/) - Earth Explorer 3D is an iOS, Android and Windows app that supports 3D Tiles, including 3D Tiles packages in online- and offline mode, as well as Cesium ion assets and different mapping services
 - [Map Data Explorer](https://mapexplorer.techmaven.net/) - Map Data Explorer is an iOS and Android App with support for different tile data formats, including 3D Tiles
 - [Oracle Spatial Studio](https://www.oracle.com/database/spatial/) - A spatial database platform that can display 3D Tiles data sets by integrating CesiumJS as a plugin, as described in [the release blog post](https://blogs.oracle.com/oraclespatial/post/oracle-spatial-studio-now-supports-displaying-3d-tiles-using-cesiumjs-plugin)
+ - [Geofox Viewer](https://viewer.geofox.ai/?tileset=tilesetURL) which parses a `tileset` url-param search query and displays the provided `tileset.json` url dataset
 
 ### Implementations
 
--- a/RESOURCES.md
+++ b/RESOURCES.md
@@ -51,6 +51,13 @@
 - [cesium_3dtiles_samples](https://github.com/bertt/cesium_3dtiles_samples) - Sample data sets demonstrating different features of 3D Tiles
 - [mapbox_3dtiles_samples](https://github.com/bertt/mapbox_3dtiles_samples) - Sample data for MapBox 3D Tiles support, with a [blog post 'Exploring 3D Options of MapBox V3'](https://bertt.wordpress.com/2023/08/21/exploring-3d-options-of-mapbox-v3/)
 
+### Open-data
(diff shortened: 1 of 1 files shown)


javagl:
Cool, I think that such a section certainly makes sense

- I'd call it "Open Data" (instead of "Open-data")
- The first link seems to be broken (missing the `https://`)
- Details (I think it's "e.g." and not "eg") 
- Maybe the _direct_ links to tileset JSON files could be omitted (unless they are particularly hard to find from the respective landing pages), but no strong opinion on that...

pjcozzi:
@jo-chemla great idea, thanks for contributing!  Mentioning @dkbraig who works on data curation at Cesium and may be interested to see what you included.

jo-chemla:
Thanks for the feedback, all great points - just edited the PR. 
Indeed I was considering removing `tileset.json` urls, since most are easy to find-out - but having them accessible make it easier to create a web-app based on these tilesets - or even at some point, a STAC catalog based on the [3D Tiles web-map-links](https://github.com/stac-extensions/web-map-links#3d-tiles) extension. 
There are more datasets that were available at some point, like the below, but I could not find the endpoints so I did not include them: 

- [Strasbourg Od@CiT city](https://github.com/Ville-Eurometropole-Strasbourg/ODACiT/blob/master/Cesium_utilisateur_debutant.pdf) had an amazing collection of datasets available [here](https://odacit.3d.strasbourg.eu/) (3dtiles tilesets and others) but seem unavailable at the moment
- [Plateau for Japan data](https://webimaging.lillemetropole.fr/externe/maillage/2016_mel_10cm/3dtiles/tileset.json) is an interesting resource, but could not find direct link to tilesets

javagl:
The link to https://odacit.3d.strasbourg.eu/ seems to work for me (and from a quick glance, it seems to be pretty high quality/detail) - does "unavailable" mean that you receive some 404s for that...?

The PLATEAU data is on the radar. (And there are some related projects like https://github.com/MIERUNE/plateau-gis-converter that have been on my "maybe-have-a-look-at-that" list for a while now). If I remember correctly, they already offer _some_ data in 3DTiles, but think that this only had been untextured buildings, similar to OSM buildings. But the actual _source_ data that they offer already served as the basis for the Cesium Japan 3D Buildings data https://cesium.com/platform/cesium-ion/content/japan-3d-buildings/

I'll wait a moment for the case that someone has further comments or entries to add, but otherwise, I'd merge this soon. (Further entries can always be added later, and I do have a few entries to add to other sections of the RESOURCES.md anyhow)

jo-chemla:
Thanks for getting back again. 
 - Weird for [StrasbourgOd@CiT](https://odacit.3d.strasbourg.eu/) , the whole app just return 404 from our office, but does run great from another IP, so will update the PR to include it. Indeed this is one of the best multi-city-wide datasets I've seen coming from a public organization - let aside Google or Bing 3D Cities which should soon open its 3dtiles endpoint. 
 - Plateau seem like a really great project, indeed they show more polygonal OSM-like viz, but I thought they also shared 3D-scan data - viewer is [here](https://plateauview.mlit.go.jp/), could not find it. Thanks for the link to that cesium showcase for Plateau!
 - I'll quickly reach-out to see if other orgs/companies have open-data 3dtiles datasets they want to share on this Resources and include them if I get some!

dkbraig:
@jo-chemla Thanks for pulling this information together.  I noticed Aerometrex and Vexcel listed in your open data section.  I do not believe the data they provide can be considered open data.  There are some sample data sets available through Cesium ion.  

RE the Plateau data, the source data can be found through the G Spatial Information Center - https://front.geospatial.jp/.  In addition they have provided tutorials for accessing 3D Tiles data here: https://github.com/Project-PLATEAU/plateau-streaming-tutorial/blob/main/3d-tiles/plateau-3dtiles-streaming.md

jo-chemla:
Thanks for getting back. Indeed, Aerometrex and Vexcel are more like dataset samples rather than open-data, so I removed their viewers from the resources. And nice to see this plateau tutorial. When digging through I indeed found only vector 3D-tiles datasets, but I included one in the open-data section anyway.

javagl:
Thanks again for the contibution and for integrating the feedback. 

Sooner or later, there probably will be some (more) overlap between "Data Providers" (and their demos), "Sample Data", and "Open Data", but it's good to have all resources in one place, and think that the "Open Data" section in particular will grow quickly and become a useful collection here.

PR #780, opened 2024-12-03
Switch z and y semantics for implicitly tiled cylinders

j9liu:
In `3DTILES_bounding_cylinder`, the schema says that implicitly tiled cylinders should use `y` to indicate changes in height, and `z` for angle. This felt inconsistent with how the height of the cylinder is aligned with the `z`-axis.

This PR just switches the two. (This will probably break existing tilesets, but hopefully any tilers out there can make an easy switch.)


lilleyse:
Thanks @j9liu

PR #781, opened 2024-12-04
Fix schemas references

nickcomua:
I tried to use you json schemas but they didn't work. so i fixed references.

### DIFF
--- a/specification/schema/PropertyTable/propertyTable.property.schema.json
+++ b/specification/schema/PropertyTable/propertyTable.property.schema.json
@@ -2,7 +2,7 @@
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "$id": "propertyTable.property.schema.json",
     "title": "Property Table Property",
-    "$ref": "rootProperty.schema.json",
+    "$ref": "../common/rootProperty.schema.json",
     "description": "An array of binary property values. This represents one column of a property table, and contains one value of a certain property for each metadata entity.",
     "properties": {
         "values": {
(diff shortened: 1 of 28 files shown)


javagl:
Yes, these `$ref`s are not canonical and seemingly refer to "wrong directories". There is some vagueness in what is the "right" or "best" way to solve this, with some subtleties in https://json-schema.org/understanding-json-schema/structuring#dollarref ...

Right now, the idea is, _very_ roughly speaking: ~"All the schema files belong into _one_ directory. And the fact that they are in different directories is only intended for keeping _some_ structure in the repository". That's probably not ideal - particularly, from the perspective of _tools_ that may want to process these schemas in an automated fashion, and therefore will _rely_ on the `$ref`s to be valid. 

(What is the context or tool in which you wanted to use the JSON schemas?)

nickcomua:
I just want to add "$schema" with a link to the GitHub schema file  in my JSON file to have autocompletion.

javagl:
OK, this could either be ridiculously trivial, or ridiculously complicated.

I tried out this state with https://github.com/CesiumGS/cesium-native/tree/main/tools/generate-classes , and this worked smoothly: The generated output was identical to the one that was generated with the current `main` state of 3D Tiles.

Then I wanted to run this with https://github.com/CesiumGS/wetzel/tree/generate-3dtiles (which is used for generating https://github.com/CesiumGS/3d-tiles/blob/main/specification/PropertiesReference_3dtiles.adoc and PropertiesReference_3dtiles_schema.adoc). And I changed the call for wetzel to omit the "search paths" argument. This should plainly no longer be necessary. I then descended into a few hours of whack-a-mole (more than I'd like to admit...), with issues where some `$ref` (or associated type information) could no longer be resolved at _some_ point in the whole generation process, and considerations about possible refactorings for that wetzel state. But eventually, I thought: "Why bother, when there is a working solution?". That "working solution" is to simply _keep_ the "search paths". Yes, they should no longer be necessary. Yes, this hints at some aspect of the wetzel schema resolving mechanisms being not as generic as they _should_ be. But we can simply say that wetzel is an internal tool, without any requirement that goes beyond "It can generate the desired adoc files". 

And it does.

When keeping the "search paths", then it still happens to generate the required `adoc` files, _verbatim_ as in the state before, (except for the inlined schemas in the `PropertiesReference_3dtiles_schema.adoc`, of course).

I'll confirm a few things (roughly, that ~"everything works on a freshly cloned state"), but think that this could be OK to merge, unless we _really_ want to achieve a state where the search paths can be omitted in wetzel.

javagl:
I confirmed that with this change, the `cesium-native` toolchain still seems to work for generating the code, and the resulting generated classes are _equal_ to the previous ones.

The fact that `wetzel` still requires these "search paths" is... a bit unfortunate, but not a deal breaker. Trying to generalize wetzel to no longer require them may imply some refactoring, and this may not be worthwhile, considering that _without any changes_ (and when still passing in the search paths), it still works and still generates the exact same output, using the branch behind https://github.com/CesiumGS/wetzel/pull/89 

So I think that's OK to merge.

@lilleyse Unless you disagree, I'd also update the [`PropertiesReference_3dtiles_schema.adoc`](https://github.com/CesiumGS/3d-tiles/blob/dee44b1c10e033723de932411e7904cf439ddeae/specification/PropertiesReference_3dtiles_schema.adoc) (which only contains the `.schema.json` files, inlined) with the one that contains the corresponding updates:

Cesium 3D Tiles Path Updates

javagl:
See https://github.com/CesiumGS/3d-tiles/pull/787 for the follow up

PR #788, opened 2025-01-28
Update `3DTILES_content_voxels` for glTF voxels

j9liu:
This favors the glTF `EXT_primitive_voxels` extension and removes the old voxel format from the spec.


lilleyse:
This looks good. My only suggestion is to add a sentence that explicitly requires `dimensions`, `padding`, and `class` to match the properties in the glTF. And maybe also a requirement that the bounding volume matches the `shape`.

j9liu:
Good point @lilleyse, thanks. What do you think of my changes?

lilleyse:
:+1: looks good

PR #791, opened 2025-02-13
Changes to `3DTILES_bounding_volume_cylinder`

j9liu:
This PR lets `3DTILES_bounding_volume_cylinder` now represent partial cylinder regions, instead of just solid cylinders. This is for better interfacing with implicit tiling, where a full cylinder can be subdivided into smaller cylinder regions -- those smaller regions can now be more accurately represented.

I wasn't sure if I should rename this to `3DTILES_bounding_volume_cylinder_region`? It's a lengthy extension name 😅 

Also, the angle now opens counter-clockwise, to align with https://github.com/CesiumGS/glTF/pull/78.


lilleyse:
Thanks @j9liu. I think the current extension name is ok.

PR #793, opened 2025-02-25
Cylinder angle changes

j9liu:
Based on offline discussion with @jjhembd and @lilleyse, we found that for cylinder regions,

- The `angle = 0` alignment along the `+y`-axis was not intuitive; usually in trig, `angle = 0` aligns with `+x`. 
- CesiumJS was computing angle relative to the x-axis anyways.

This change rotates the cylinder such that `angle = 0` aligns with `+x`, and the `-pi/pi` discontinuity aligns with `-x`. This helps with both points -- making the spec more intuitive, and preventing breaking changes in CesiumJS.

CC https://github.com/CesiumGS/glTF/pull/79 for related glTF changes


extensions/3DTILES_content_voxels/README.md
```diff
@@ -201,4 +211,48 @@ Tileset JSON with implicit tiling
...
+
+As long as the region's angular range is less than `2pi`, the ordering of the voxel data remains the same. Instead, the difference can be treated as an offset in the `minAngle` and `maxAngle` properties of the region.
+
+Given an `EXT_implicit_cylinder_region` in glTF, its equivalent in `3DTILES_bounding_cylinder` can be deduced by adding `pi` to its angular bounds. If the resulting angle is greater than `2pi`, then it should be wrapped around to its equivalent in the `[-pi, pi]` range (see below).
+
+![](figures/cylinder-angle-rotation.png)
+
+No matter how the `minAngle` and `maxAngle` change, a voxel located at `[x, y, z]` in the glTF will still be located at [`x, z, y`] in 3D Tiles, and vice versa. (`y` and `z` are swapped due to the different up-axes.)
+
+However, this rule does not work if the cylinder region fills the entire `2pi` range (such that `minAngle = -pi` and `maxAngle = pi`). In this case, the indices used to access the voxel data will be affected. A **cylinder** voxel at `[x, y, z]` within the glTF will be located at `[x, z, a]` in 3D Tiles, where `a = [y + (angle / 2)] % angle`. The math is the same vice versa.
```
jjhembd, line 248: Are `y` and `z` switched here?

A cylinder voxel at `[x, y, z]` within the glTF has its angle defined by its `z` coordinate, and its height defined by `y`. So the 3D Tiles coordinate would be `[x, a, y]`, where `a = (y + angleRange / 2) % angleRange`.

I was initially confused by the word `angle`, until I realized it meant the tile dimension along the angle axis. Would `angleRange` be clearer?

j9liu: Ah you're right, thanks for the catch @jjhembd !

PR #794, opened 2025-02-28
Fix cylinder math in README.md

j9liu:
Addresses comment on #793 . Thanks for the catch @jjhembd !


jjhembd (approved):
Looks good, thanks @j9liu!

Issue #701, opened 2022-06-29
Top-level information about tile content


lilleyse:
Right now there's no place to put top-level information about tile content. Some use cases include:

* If the tileset has glTF content, list out the `extensionsUsed` and `extensionsRequired` so that an engine can bail out early if it sees a glTF extension it doesn't support
* As [glTF profiles](https://github.com/KhronosGroup/glTF/issues/2092) start to develop, a tileset could require content to conform to a certain profile. This can help engines use specialized code paths (e.g. point clouds vs. unlit textured meshes).
* Conventions for multiple contents (e.g. "first content uses this profile, second content uses this profile")
* Top-level information for content types besides glTF

An early version of this was introduced in [3DTILES_content_gltf](https://github.com/CesiumGS/3d-tiles/tree/main/extensions/3DTILES_content_gltf) but did not make it back into 3D Tiles 1.1 because of some uncertainty about the use cases above.

```json
  "extensions": {
    "3DTILES_content_gltf": {
      "extensionsUsed": ["EXT_mesh_gpu_instancing"],
      "extensionsRequired": ["EXT_mesh_gpu_instancing"]
    }
  }
```

bjornblissing:
An interim solution would probably be a new extension. For example named `3DTILES_gltf_extensions`, which specify the used and required extension in a similar way as the `3DTILES_content_gltf` did:
```json
  "extensions": {
    "3DTILES_gltf_extensions": {
      "extensionsUsed": ["EXT_mesh_gpu_instancing"],
      "extensionsRequired": ["EXT_mesh_gpu_instancing"]
    }
  }
```

PR #810, opened 2025-06-02
Un-deprecate `3DTILES_content_gltf`

javagl:
The [`3DTILES_content_gltf`](https://github.com/CesiumGS/3d-tiles/blob/e57f89c1ab779f38492a7fa1ef89be161474d118/extensions/3DTILES_content_gltf/README.md) extension originally had the purpose of allowing 3D Tiles 1.0 tilesets to directly contain glTF assets as the tile content.

When support for glTF tile content became part of the core specification in 3D Tiles 1.1, this extension was deprecated.

But in addition to just allowing glTF tile content, the extension also served another purpose - namely, to allow specifying the `extensionsUsed/Required` _of the glTF content(!)_ inside the tileset JSON. This is important for clients, to determine early whether they will be able to load/process a tileset, _without_ having to inspect each tile content, and possibly reject it as they load it.

There is no strong reason to not allow the extension being present in a 3D Tiles 1.1 tileset for this exact purpose.

So this PR un-deprecates the extension. Minor wording updates in the README.md aim at clarifying the different roles that the extension has in 3D Tiles 1.0 and 1.1.

This is one tiny part of https://github.com/CesiumGS/3d-tiles/issues/701 , I'll add another comment there.

---

<sup>(Searching for the "opposite of 'deprecate'" brought up 'praise' - hence the branch name)</sup>

### DIFF
--- a/extensions/3DTILES_content_gltf/README.md
+++ b/extensions/3DTILES_content_gltf/README.md
@@ -1,5 +1,3 @@
-### ⚠️ `3DTILES_content_gltf` has been promoted to core in 3D Tiles 1.1. See [glTF Tile Format](../../specification/TileFormats/glTF/). ⚠️
-
 # 3DTILES_content_gltf
 
 ## Contributors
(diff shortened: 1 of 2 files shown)


lilleyse:
Thanks @javagl

PR #830, opened 2025-11-12
Added ESRI_crs to the list of extensions

jeanphilippepons:
This extension was introduced to handle projected coordinates in 3D tilesets of gaussian splats, without client-side reprojection. It's supported by ArcGIS Pro 3.6 and the associated Esri products (ArcGIS Reality Studio, ArcGIS Drone2Map), released in November 2025. It will also be supported by ArcGIS Online 2026 R1.
In the future, we ambition to also use this extension for the 3D Meshes produced by Esri's Reality Mapping products.
Please refer to the specs of the vendor extension for more detail.

### DIFF
--- a/extensions/README.md
+++ b/extensions/README.md
@@ -30,7 +30,7 @@ Extension|Notes
 [MAXAR_content_geojson](https://github.com/Maxar-Public/3d-tiles/tree/wff1.7.0/extensions/MAXAR_content_geojson)|_Written against 3D Tiles 1.1_
 [MAXAR_extent](https://github.com/Maxar-Public/3d-tiles/tree/wff1.7.0/extensions/MAXAR_extent)|_Written against 3D Tiles 1.0 and 1.1_
 [VRICON_class](https://github.com/Maxar-Public/3d-tiles/tree/wff1.7.0/extensions/VRICON_class)|_Written against 3D Tiles 1.0_
-
+[ESRI_crs](https://github.com/Esri/3D-tiles-layer/blob/main/ESRI_CRS/Esri_crs_extension.md)|_Written against 3D Tiles 1.1_
 
 ## About
 


lilleyse:
@jeanphilippepons thanks for the PR! I like the approach here.

Some feedback on the extension:

* Should geographic coordinate systems like `EPSG:4326` be explicitly disallowed? If I'm understanding correctly this approach only works for projected cartesian coordinate systems.
* Should this be a required extension? If the extension is ignored the content won't be transformed to the correct location. On a related note, are both the tile transform and extension transform applied? Or does the extension transform replace the tile transform.
* "without client-side reprojection" - I think this should be explicitly called out in the extension. Maybe something like "The data remain compatible with 3D Tiles' Earth-centered, Earth-fixed (ECEF) reference frame without client-side reprojection."
* What is the intended usage of `boundingVolume`?
* `"extensionsRequired": null` should be removed or fixed
* There are a few cases where "3D tiles" should be "3D Tiles"
* Once these changes are nailed down, it would be good to provide a JSON schema

Overall looks great. We can get this merged soon.

jeanphilippepons:
Hi @lilleyse and thanks for taking the time to look at this. Before I answer your questions, I have a high level remark. This PR is on the 3D Tiles specs repo, and it is about adding a reference to our vendor extension, it's not about reviewing the specs of our extension. Or at least it's how I understand the process. We went for a vendor extension precisely to go to market early without validation from other parties.
This said, we hope that this extension can be generalized into a 3D Tiles extension at some point, incorporating feedback like the one you provided. But it's not in the same time frame and it's not covered by this PR.
Do you agree with this?

lilleyse:
@jeanphilippepons sorry for the confusion, I didn't intend to block this PR with my feedback, I just wanted to provide feedback in case you wanted to make changes to the extension before it was listed.

I'll merge this PR, and we can continue the discussion.

jeanphilippepons:
Thanks @lilleyse, it makes sense. I will take a look at your feedback and I will follow up soon.

PR #832, opened 2026-01-13
Amend glTF ->3D Tiles math for box voxels

j9liu:
Thanks to @jjhembd for pointing this out -- I unfortunately inverted the rotation matrices to translate voxel access between glTF/3D Tiles. (I referenced matrices in Cesium Native + CesiumJS without accounting for column-major order, which resulted in transposed matrices.) This PR corrects that math.


lilleyse:
Thanks @j9liu and @jjhembd!

PR #834, opened 2026-03-18
Add `3DTILES_content_conditional`

javagl:
(This was originally called `3DTILES_dynamic`, but for alignment with the `<PREFIX>_<scope>_<feature>` pattern, it is now called `3DTILES_content_conditional`)

This extension is supposed to be a generalization of what was originally referred to as "time-dynamic" tiles. 

### Context

The idea of "Time-dynamic 3D Tiles" was originally brought up in https://github.com/CesiumGS/3d-tiles/issues/102, about 10 years ago. The issue was largely dormant since then. Recently, it has gained some traction. The goal would be to define the scope, use-cases, and possible implementations for (time-) dynamic 3D tiles more clearly (and eventually consider to make this a core feature of 3D Tiles 2.0). 

Despite the topic only recently receiving more attention, there are some discussions that are a bit scattered at different places. I'll try to summarize a few links here:

The initial proposal for (time) dynamic 3D Tiles was made in [this comment](https://github.com/CesiumGS/3d-tiles/issues/102#issuecomment-3020925540). It consisted of the following pull requests:
- https://github.com/CesiumGS/3d-tiles/pull/817 : This defines metadata with 'semantics' for the `content`. This metadata was supposed to be assigned to contents in the `tile.contents` array. The metadata defines a time-range, given as ISO8601 strings, indicating that the respective content should be associated with (or 'exist in') this time range.
- https://github.com/CesiumGS/3d-tiles/pull/816 : This defines a generalization of the concept of "multiple contents" for implicit tiling, with the goal of having a more compact representation of content metadata in implicit tiling, for the case that the number of contents should vary between the implicit tiles
- https://github.com/CesiumGS/3d-tiles/pull/815 : This defines a generalization of the 'template URI' concept of implicit tiling, with the goal of being able to access contents not only based on the implicit tile coordinates, but also based on 'time stamps' (or other properties)

I wrote a longer [comment with a first feedback pass](https://github.com/CesiumGS/3d-tiles/issues/102#issuecomment-3084996001) on that. The summary of that is

- Representing "(time) dynamic contents" as "multiple contents" (i.e. the `tile.contents` array) raises a few questions. Multiple contents are not so widely supported. It is not clear how to combine the metadata that determines the "time-dynamicness", and _other_ metadata that is supposed to be stored for the content. There is no established infrastructure for maintaining "metadata semantics". Fixing the "dynamicness" to _time_, and to ISO8601 strings in particular, may be too restrictive.
- The generalization of "multiple contents for implicit tilings" was just a workaround for some of these limitations
- The generalization of the "template URI" is something that should be considered, but is not specific for "(time)-dynamic content"

I therefore proposed to generalize some of that to _dynamic_ content (not restricted to _time_). The initial drafts/suggestions have been added in that comment and the subsequent comments in that issue. Further discussion has happened there and at other places. The draft pull request for the _implementation_ in CesiumJS, at https://github.com/CesiumGS/cesium/pull/12977 , contains additional considerations. Some of them are focussed on the actual implementation, but obviously connected to the underlying data model (and what the implementation actually _needs_ from that data model).

### Current State

The `README.md` of this pull request reflects the current state of the discussion. This is still a _draft_. 

I'll try to summarize some of the open questions here. Given that these questions have mainly been brought up during the development of the implementation for CesiumJS, it might be better to discuss this in the [implementation PR in CesiumJS](https://github.com/CesiumGS/cesium/pull/12977): 

- Type information or schema definition: Right now, the top-level extension object defines the "dimensions" that serve as the basis for the selection of the active content. The types of these properties are currently limited to `string`. One could make a strong case to allow additional types here, and maybe include additional information about valid value ranges. At some point, this would resemble a "metadata schema". Actually _using_ a "3D Tiles Metadata Schema" was considered in [a comment about using metadata in the implementation pull request](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3576792432), but did imply too many possible inconsistencies and undesired side-effects. But still, there are several levels for the types that could be _allowed_ for these dimensions. Some of the options are summarized in [another comment about type information in the implementation pull request](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-4069090697). 

- Selection/activation criteria: Until recently, the implementation made the assumption that two content items could not be 'active' at the same time when one of their 'keys' properties had different values. This allowed building an efficient "lookup" in the implementation, but it did limit the flexibility of what could (easily) be modeled. As part of a recent generalization, that assumption has been lifted. The new approach allows defining arbitrary conditions for a content item to be 'active'. This may have performance implications and may warrant further investigations. Some details are in a [comment about activation criteria in the implementation pull request](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-4017076580).


lilleyse:
Should `conditionalContent.schema.json` inherit from `common/rootProperty.schema.json`? Then it would be possible to include `extras` and `extensions`.

javagl:
Given the current pattern of the extensions _not_ referring to a directory (which would be `../specification/schema/common...` in this case), I just added them as `rootProperty.schema.json`.

extensions/3DTILES_content_conditional/schema/conditionalContentItem.schema.json
```diff
@@ -0,0 +1,19 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "conditionalContentItem.schema.json",
+  "title": "One item of the dynamic content for 3DTILES_content_conditional",
+  "description": "Defines the structure of a single content item in 3DTILES_content_conditional",
+  "$ref": "content.schema.json",
+  "properties": {
+    "keys": {
+      "type": "object",
+      "additionalProperties": {
+        "type": "string"
```
lilleyse: Any concern with supporting "boolean" and "number" as well?

I think this came up a few weeks ago but don't remember what we decided.

javagl: Yes, it was discussed and decided at https://github.com/CesiumGS/cesium/pull/12977#issuecomment-4069511572 , but not yet updated here. I now updated it accordingly.

bjornblissing:
In `README.md` the conditional content file is referenced via `tile.content.uri`, but it doesn’t state how to resolve the inner `conditionalContents[i].uri` values (relative to the tileset JSON vs. relative to the conditional content JSON file itself). This is a real interoperability risk because base-URI rules are normally _“relative to the JSON that contains the URI”_, but that expectation isn’t stated here.

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Conditional Content Type
+
+The new content type for the conditional content is represented as a JSON file. Such a file can be referred to via the `tile.content.uri` in a 3D Tiles data set. The file contains an array `conditionalContents`, where each item is a [3D Tiles 1.1 `content`](https://github.com/CesiumGS/3d-tiles/blob/1.1/specification/schema/content.schema.json) with additional properties.
+
+The additional properties that are defined for each item are the `keys`. These keys serve as the basis for deciding whether the respective content item should be active. The keys are objects with arbitrary string-typed properties. The set of properties is defined by the top-level extension object (as described below).
```
bjornblissing: > “keys are objects with arbitrary string-typed properties”

However, the `conditionalContentItem.schema.json` allows for values of `"string" | "number" | "boolean"`

extensions/3DTILES_content_conditional/examples/exampleTileset.json
bjornblissing, line 1: README says the extension is required and _“must be placed in both `extensionsUsed` and `extensionsRequired`”_ 
But this example does not contain either list.

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+## Optional vs. Required
+
+This extension is required, meaning it must be placed in both the `extensionsUsed` and `extensionsRequired` lists in the tileset JSON.
+
+## Overview
+
+This extension adds support for conditional content in 3D Tiles, by defining the following elements:
+
+- A new content _type_ that can be referred to via the `tile.content.uri` in a 3D Tiles data set.
+- An _extension object_ in the top-level tileset JSON that descries the structure of the conditional content
```
bjornblissing: Minor spelling error:
```suggestion
- An _extension object_ in the top-level tileset JSON that describes the structure of the conditional content
```

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+      "exampleTimeStamp" : "2025-09-26",
+      "exampleRevision" : "revision1"
+    }
+  } ]
+}
+```
+
+### Top-level Extension Object
+
+The structure of the condition contents is defined with a top-level extension object in the tileset JSON. This object is stored as the `3DTILES_content_conditional` object in the `extensions` dictionary of the tileset JSON. An example of such an object - corresponding to the example content from the previous section - is shown here:
```
bjornblissing:
```suggestion
The structure of the conditional contents is defined with a top-level extension object in the tileset JSON. This object is stored as the `3DTILES_content_conditional` object in the `extensions` dictionary of the tileset JSON. An example of such an object - corresponding to the example content from the previous section - is shown here:
```

extensions/3DTILES_content_conditional/schema/conditionalContentItem.schema.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "conditionalContentItem.schema.json",
+  "title": "One item of the dynamic content for 3DTILES_content_conditional",
```
bjornblissing:
```suggestion
  "title": "One item of the conditional content for 3DTILES_content_conditional",
```

extensions/3DTILES_content_conditional/schema/tileset.3DTILES_content_conditional.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "3DTILES_content_conditional.schema.json",
```
bjornblissing: To align with common pattern of other extensions:
```suggestion
  "$id": "tileset.3DTILES_content_conditional.schema.json",
```

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+- A new content _type_ that can be referred to via the `tile.content.uri` in a 3D Tiles data set.
+- An _extension object_ in the top-level tileset JSON that descries the structure of the conditional content
+
+In the context of this extension, 'conditional content' is tile content where the actual content data that is loaded and rendered depends on user-selectable criteria. In the context of this specification, the content data item that should be loaded and rendered is referred to as the _'active'_ content data. 
+
+
+> **Implementation Note**
+>
+> This extension specification itself does not define the behavior of clients regarding this 'active' content data. Clients could choose to pre-load all content data items, to allow them to quickly switch between them when a certain item becomes active. Alternatively, they can choose to download a content item lazily when it becomes active. When they download the content lazily, they can choose to display the previously active content until the newly active content is available and ready to be rendered, or they can choose to display a placeholder or loading indicator instead.
```
bjornblissing, line 32: The implementation note says client behavior is out of scope (fine), but the spec doesn’t state basic determinism constraints, e.g.:
- What if multiple `conditionalContents` entries match the currently selected criteria?
- What if **none** match?
- Is there a required **default** active content when criteria are unset?

extensions/3DTILES_content_conditional/README.md
javagl: This was updated a few hours ago (see comments above), but the text is not updated accordingly. I also have to create test data and run tests with all that, so will update the text ASAP after that

extensions/3DTILES_content_conditional/README.md
javagl: Today I learned that (the spell checker did not catch this because) "descries" is an actual word

extensions/3DTILES_content_conditional/examples/exampleTileset.json
javagl: Updated (locally, will push soon)

javagl:
Some aspects of the state here - as indicated by the 'Draft' state - have not been 'final yet, and obviously, a proofreading pass was still missing. But it's good to capture some points as early as possible. 
(I'd likely have overlooked that 'descries', FWIW...)

The technical questions:

---

> ... it doesn’t state how to resolve the inner conditionalContents[i].uri values (relative to the tileset JSON vs. relative to the conditional content JSON file itself). This is a real interoperability risk because base-URI rules are normally “relative to the JSON that contains the URI”, but that expectation isn’t stated here.

I think that it makes sense to state this _explicitly_. 
(I'll have to create test data to see how CesiumJS is behaving in this case. I assume that it will already do that relative-to-container resolution automatically, but will verify)

---

> However, the conditionalContentItem.schema.json allows for values of "string" | "number" | "boolean"

The text was fixed, but similar to the above: I'll have to create test data for that.

---

> The implementation note says client behavior is out of scope (fine), but the spec doesn’t state basic determinism constraints, e.g.:
> 
> * What if multiple conditionalContents entries match the currently selected criteria?
> * What if none match?
> * Is there a required default active content when criteria are unset?

There currently are no constraints at all:

1. When multiple entries match, then all the matching contents are displayed. 
2. When none match, then no content is displayed
3. There is no default active content

This is intentional insofar that it is the most direct mapping of the definition in the JSON to a rendered state. 

If 1. or 2. wasnt the case: Imagine some terrain, with some building on top, and you'd like to show both, one of them, or none. Similarly, if 3. wasnt't the case, then it wouldn't be possible to distinguish a state where something matches from a state where nothing matches

**But**... the questions that you brought up here are strongly related to the questions about "wildcards" that I bought up earlier (e.g. in a [recent comment in the implementation PR](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3938779884), which links to one of the [first comments in the main issue](https://github.com/CesiumGS/3d-tiles/issues/102#issuecomment-3307045781)). 

There is a distinction for establishing such constraints or defaults. It refers to whether a certain decision is made "on the data side" or "on the implementation side". 

On the data side, one could define something like this:
```
    "uri" : "content-0-0-2025-09-26-any.glb",
    "keys" : {
      "exampleTimeStamp" : "2025-09-26",
      "exampleRevision" : null/undefined       // Leave this null, undefined, or omit it
    }
```
This could mean: This content should be displayed _always_, regardless of which `exampleRevision` the client selects.

On the implementation side, the client could select content with something like this:
```
const exampleTimeStamp = "2025-09-26";
const exampleRevision = undefined;  // Leave this undefined
selectContent(exampleTimeStamp, exampleRevision);
```
This could mean: Display _all_ contents, regardless of which `exampleRevision` they have.

This may appear to be subtle, but I think very important: It affects _what exactly can be modeled_. For example, with the first solution, it would be impossible for the client to _hide_ that particular content. I have to think through how to describe that more formally (and some of that has changed due to the generalization of the selection mechanism that was [mentioned in this implementation PR comment](https://github.com/CesiumGS/cesium/pull/12977#issuecomment-3938779884)).

This is one of the aspects that have to be described more rigorously in the spec: What happens with 'undefined' keys? Or properties that are not in the 'dimensions.keySet' arrays? (Corollary: Can the 'keySet' contain 'undefined'? What does this mean?)

If you have specific thoughts about how some of that _should_ or _shouldn't_ be specified (maybe related to actual usage scenarios), that could help to find potential caveats.

extensions/3DTILES_content_conditional/README.md
javagl: Added a comment below

bjornblissing: I am missing some form of normative language on how active content should be matched. Something like this:

```suggestion
### Active content selection

The `keys` properties of each element of `conditionalContents` define a discrete set of content variants. A client determines an **active criteria** object (outside the scope of this extension) that maps dimension names to selected criteria values.

If no active criteria object is provided (or the criteria object is empty), there is **no active content** and the tile is treated as having no content.

The active criteria object **MAY** specify values for a subset of the dimensions defined by the tileset’s `3DTILES_content_conditional.dimensions`. When the active criteria object is non-empty, dimensions that are not present in the active criteria object are treated as **wildcards** (i.e., they do not affect matching).

A conditional content item **matches** a non-empty active criteria object if, for every property `p` in the active criteria object where `p` is equal to the `name` of a dimension in `3DTILES_content_conditional.dimensions`, the item satisfies the criterion for that dimension.

The process for evaluating whether an item satisfies a criterion for a dimension is **implementation-defined**. Implementations **MAY** interpret `keys` values as opaque values (e.g., for exact equality comparison), or they **MAY** interpret them using application-defined semantics (for example, interpreting a string as a date range and considering a criterion date to match when it falls within that range). This extension does not define how values are interpreted.

If an item does not contain `keys[p]` for a dimension name `p` that appears in the active criteria object, then the item does **not** match.

Properties in the active criteria object that do not correspond to a dimension name in `3DTILES_content_conditional.dimensions` have **no effect** on matching.

Given a non-empty active criteria object, the content items are selected as follows:

1. Let `matches` be the list of items in `conditionalContents` that match the active criteria object.
2. If `matches` is empty, there is **no active content** and the tile is treated as having no content.
3. If `matches` contains exactly one item, that item is the active content.
4. If `matches` contains more than one item, then **all** matching items are active and **MUST** be enabled. The order of rendering is implementation-defined.

> **Implementation Note**
>
> This extension specification itself does not define the behavior of clients regarding this 'active' content data. Clients could choose to pre-load all content data items, to allow them to quickly switch between them when a certain item becomes active. Alternatively, they can choose to download a content item lazily when it becomes active. When they download the content lazily, they can choose to display the previously active content until the newly active content is available and ready to be rendered, or they can choose to display a placeholder or loading indicator instead.
```

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Optional vs. Required
+
+This extension is required, meaning it must be placed in both the `extensionsUsed` and `extensionsRequired` lists in the tileset JSON.
+
+## Overview
+
+This extension adds support for conditional content in 3D Tiles, by defining the following elements:
+
+- A new content _type_ that can be referred to via the `tile.content.uri` in a 3D Tiles data set.
```
bjornblissing: Only mentions using the new content type via `tile.content.uri` and doesn’t address whether the conditional content JSON may also appear in `tile.contents[i].uri`.

extensions/3DTILES_content_conditional/schema/tileset.3DTILES_content_conditional.schema.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "tileset.3DTILES_content_conditional.schema.json",
+  "$ref": "rootProperty.schema.json",
+  "title": "3DTILES_content_conditional extension",
+  "description": "Defines the structure of the extension object that is stored as the 3DTILES_content_conditional extension in a tileset JSON",
+  "type": "object",
+  "properties": {
+    "dimensions": {
+      "type": "array",
```
bjornblissing, line 11: `dimensions` has no `minItems`.

extensions/3DTILES_content_conditional/schema/tileset.3DTILES_content_conditional.schema.json
```diff
@@ -0,0 +1,35 @@
...
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          },
+          "keySet": {
+            "type": "array",
```
bjornblissing, line 21: `keySet` has no `minItems`.

extensions/3DTILES_content_conditional/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+> **Implementation Note**
+>
+> This extension specification itself does not define the behavior of clients regarding this 'active' content data. Clients could choose to pre-load all content data items, to allow them to quickly switch between them when a certain item becomes active. Alternatively, they can choose to download a content item lazily when it becomes active. When they download the content lazily, they can choose to display the previously active content until the newly active content is available and ready to be rendered, or they can choose to display a placeholder or loading indicator instead.
+
+### Conditional Content Type
+
+The new content type for the conditional content is represented as a JSON file. Such a file can be referred to via the `tile.content.uri` in a 3D Tiles data set. The file contains an array `conditionalContents`, where each item is a [3D Tiles 1.1 `content`](https://github.com/CesiumGS/3d-tiles/blob/1.1/specification/schema/content.schema.json) with additional properties.
```
bjornblissing: The core spec has a _“File Extensions and Media Types”_ section that says tile content files should use the file extensions and Media Type specific to their tile format specification.

This extension defines a new tile content type that is “represented as a JSON file” (i.e., referenced from `tile.content.uri`), but it doesn’t explicitly state a file extension / media type for that content.

extensions/3DTILES_content_conditional/schema/conditionalContentItem.schema.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "conditionalContentItem.schema.json",
+  "title": "One item of the conditional content for 3DTILES_content_conditional",
+  "description": "Defines the structure of a single content item in 3DTILES_content_conditional",
+  "$ref": "content.schema.json",
+  "properties": {
+    "keys": {
```
lilleyse, line 8: Should all properties have a `description`? See comments [here](https://github.com/CesiumGS/cesium-native/pull/1351#discussion_r3175309484).

extensions/3DTILES_content_conditional/README.md
javagl: Added a short statement about that (may be moved around in a later cleanup pass, but it's good to have it there for now)

extensions/3DTILES_content_conditional/schema/tileset.3DTILES_content_conditional.schema.json
javagl: Both `dimensions` and `keySet` do _currently_ not have `minItems` because we _might_ actually consider allowing them to be empty. (Roughly, to "emulate" the current "multiple contents" concept). But it might very well be that this doesn't make sense, and `minItems: 1` will be added later. (I'll leave this open for now)

extensions/3DTILES_content_conditional/schema/conditionalContentItem.schema.json
javagl: I added a short description to each. Some wordsmithing might happen in a final review pass.

extensions/3DTILES_content_conditional/README.md
javagl: I changed the current wording accordingly: Instead of `tile.content.uri` (where the `tile` part could be seen as questionable as well), I just refer to it as "content URI". The [main spec](https://github.com/CesiumGS/3d-tiles/blob/main/specification/README.adoc#tile-content) refers to `content.uri`, in the narrow context of explaining what the content is. Given that the spec here explicitly links to the `content` schema, this should probably be OK.

extensions/3DTILES_content_conditional/README.md
javagl: That last comment was apparently added _after_ I resolved the conversation, and I don't think I have been notified about this (but just stumbled over it now rather accidentally). 

I had addressed _some_ of the points from the first comment in my comment below. But the section that you added does contain some of the detailed description that would be required for the specification, eventually. Some of this had not yet been added to the current ('Draft') state because it was not yet entirely clear what the final decision would be.

The open questions are mainly related to the thoughts about 'wildcards', and I'll try to think this through more thoroughly, and how your proposed description aligns with the options (and maybe degrees of freedom) that we want to offer. 

(In any case, I now 'unresolved' this conversation, to prevent this from being overlooked so easily)

Issue #825, opened 2025-09-09
Vector data... in glTF?


lilleyse:
This is an idea I've been playing around with for a while. Can we use glTF for vector data? Specifically, large amounts of 3D cartesian vector data. Here are some initial ideas.

### Points

* Use primitive `mode: 0` (`POINTS`)
* Use [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/blob/proposal-EXT_mesh_features/extensions/2.0/Vendor/EXT_mesh_features/README.md) and [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/blob/4bc06a1b608a44d15ccc9826f4a7bd4a7e696692/extensions/2.0/Vendor/EXT_structural_metadata/README.md) to associate points with metadata
  * `Feature ID by Index` + property tables
* Styling is applied separately, e.g. with [3D Tiles Declarative Styling](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Styling)
  * This would include label rendering, anchor lines, etc.

> [!NOTE]
> * How does this relate to [KHR_billboard](https://github.com/KhronosGroup/glTF/pull/2522)?

### Lines

* Use primitive `mode: 1` (`LINES`), `mode: 2` (`LINE_LOOP`) or `mode: 3` (`LINE_STRIP`)
  * For the latter two, use [`EXT_mesh_primitive_restart`](https://github.com/KhronosGroup/glTF/pull/2478) to batch multiple lines into a single primitive
* Use [`EXT_mesh_features`](https://github.com/CesiumGS/glTF/blob/proposal-EXT_mesh_features/extensions/2.0/Vendor/EXT_mesh_features/README.md) and [`EXT_structural_metadata`](https://github.com/CesiumGS/glTF/blob/4bc06a1b608a44d15ccc9826f4a7bd4a7e696692/extensions/2.0/Vendor/EXT_structural_metadata/README.md) to associate lines with metadata.
  * `Feature ID by Vertex` + property tables.
* Styling is applied separately, e.g. with [3D Tiles Declarative Styling](https://github.com/CesiumGS/3d-tiles/tree/main/specification/Styling)
  * This would include mitering, line width, dashed patterns, etc.

### Polygons

* Use primitive `mode: 4` (`TRIANGLES`) - polygons are pre-triangulated for performance reasons
[...]

PR #838, opened 2026-04-20
Add `3DTILES_content_gltf_vector`

donmccurdy:
For #825. Defines "vector data" as applied to 3D Tiles, and methods of encoding and decoding vector data to/from glTF content.

The extension provides some fundamental structures (buffer regions, clipping) without which common visual styles would not be possible, but the extension otherwise leaves styling undefined, to be provided by current and future iterations of the 3D Tiles styling specification, and/or explored further by client implementations.


extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Working from the definition above, this extension proposes a mechanism for encoding vector point, line, and polygon data in 3D Tiles using glTF content, and for distinguishing such vector data from general glTF content. The extension further refines tile and content bounding volumes as needed to support seamless tiled vector rendering in a variety of visual styles.
+
+## Extending 3D Tiles content
+
+A `content` definition, containing a reference URL or template URL to glTF content, may be extended with the `3DTILES_content_gltf_vector`. The extension's boolean property, `vector: true`, indicates that the glTF content of the tile SHOULD be interpreted as vector data. The value of the `vector` property MUST be `true`; for non-vector content the extension is omitted.
```
donmccurdy, line 62: Question — which of the following do we want to allow?

1. Multiple vector "layers" in a tileset, represented e.g. as distinct glTF Nodes
2. The same 3D Tiles _tileset_ containing multiple contents, some vector layers, others non-vector glTF content, e.g. gaussian splats
3. The same 3D Tiles _content_ (i.e. glTF asset) containing both vector layers and non-vector glTF content, e.g. gaussian splats

I suspect yes on (1), but I'm not sure about 2 and/or 3.

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+Extending the concept of vector data into the domain of graphics APIs, and of 3D scenes already composed of graphics primitives — such as points, lines, triangles — this extension proposes and defines an additional distinction: \*\*“vector data” comprises point, polyline, and polygon geometries with intrinsic semantic meaning, but without intrinsic rendering intent.
+
+> **Implementation note:** Unlike typical glTF geometry, vector data geometry does not directly convey rendering intent. Vector data is a vehicle for topology or for associated properties. Points may be aggregated, clustered, or used as anchors for labels. Lines may be widened or dashed, or used as an invisible track for animation. Polygons may be outlined, extruded, subtracted from existing scene geometry, or used to define an abstract area of analysis.,
```
weegeekps:
```suggestion
> [!NOTE]
> Unlike typical glTF geometry, vector data geometry does not directly convey rendering intent. Vector data is a vehicle for topology or for associated properties. Points may be aggregated, clustered, or used as anchors for labels. Lines may be widened, dashed, or used as an invisible track for animation. Polygons may be outlined, extruded, subtracted from existing scene geometry, or used to define an abstract area of analysis.
```

GitHub markdown has [some helpers for notes, warnings, and more](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) now. I changed this note (and others here) to use that functionality and did a little grammar editing as well.

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+The buffer region is provided to mitigate seams and discontinuities at tile boundaries. For polylines and polygons crossing tile boundaries and rendered with certain visual styles — particularly "wide" lines or outlines — display in tile A may be affected by the continuation of the same geometry a short distance ("buffer") into tile B. By clipping precisely at the tile boundary, sections of the geometry in tile B may still influence display within tile A, without duplicate rendering and/or z-fighting in tile B.
+
+<img alt="Illustration of tiles, content, and buffers" src="./figures/tiles-and-buffers.png" width="400">
+
+> TODO: Improve and describe illustration.
+
+> **Implementation note:** Clipping may be implemented by pre-processing geometry, by discarding fragments in a pixel shader, or by any other means. For typical vector visual styles (involving, for example, wide lines), it is expected that most implementations will implement clipping in the fragment shader (or equivalent), in order to preserve the influence of geometry just outside the tile boundary on lines or outlines crossing the tile boundary.
```
weegeekps:
```suggestion
> [!NOTE]
> Clipping may be implemented by pre-processing geometry, by discarding fragments in a pixel shader, or by any other means. For typical vector visual styles (involving, for example, wide lines), it is expected that most implementations will implement clipping in the fragment shader (or equivalent), in order to preserve the influence of geometry just outside the tile boundary on lines or outlines crossing the tile boundary.
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+> **Implementation note:** To avoid visual artifacts, client implementations would (in the absence of "buffer" region data) need to connect geometries in each tile to their corresponding geometries in adjacent tiles. Such mapping and reconstruction at runtime, while not prohibited, is often impractical for realtime implementations. Overlapping buffers at tile boundaries are included in 3D Tiles as an alternative.
```
weegeekps:
```suggestion
> [!NOTE]
> To avoid visual artifacts, client implementations would (in the absence of "buffer" region data) need to connect geometries in each tile to their corresponding geometries in adjacent tiles. Such mapping and reconstruction at runtime, while not prohibited, is often impractical for realtime implementations. Overlapping buffers at tile boundaries are included in 3D Tiles as an alternative.
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+As a result, `content.boundingVolume` may extend arbitrarily outside of `tile.boundingVolume`. Client implementations SHOULD implement LOD selection and tile visibility based on the tile bounding volume, not the (potentially larger) content bounding volume; tile visibility is unchanged as compared to non-vector data. Authoring implementations SHOULD include content extending _outside_ the tile bounding volume only to the extent that such content is likely to influence visualization _inside_ the tile bounding volume.
+
+> **Implementation note:** A small buffer region, representing a single-digit percentage of a tile's width, is expected to be sufficient for most rendering styles. Client implementations may choose to limit line widths to the widths of tile buffers, in screen space or world space, to avoid visual artifacts.
```
weegeekps:
```suggestion
> [!NOTE]
> A small buffer region, representing a single-digit percentage of a tile's width, is expected to be sufficient for most rendering styles. Client implementations may choose to limit line widths to the widths of tile buffers, in screen space or world space, to avoid visual artifacts.
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+> **Implementation note:** Point geometries do not typically require buffer regions or clipping at tile boundaries for proper rendering, but authoring and client implementations are not prohibited from doing either.
```
weegeekps:
```suggestion
> [!NOTE]
> Point geometries do not typically require buffer regions or clipping at tile boundaries for proper rendering, but authoring and client implementations are not prohibited from doing either.
```

extensions/3DTILES_content_gltf_vector/README.md
weegeekps: 3 seems messy. From a spec standpoint either 1 or 2 should be okay.

donmccurdy: Agreed — if we needed (3) I suppose we'd need to either navigate the messiness of specifying which part of the content is which, or push some or all of this specification down into a glTF extension instead of a 3D Tiles extension. But defining "what is a vector, and how does it interact with tile boundaries" feels messy to define at the glTF level, without reference to geospatial concepts and 3D Tiles styling.

So I think I'd lean toward 1+2 but not 3, perhaps.

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,152 @@
...
+Working from the definition above, this extension proposes a mechanism for encoding vector point, line, and polygon data in 3D Tiles using glTF content, and for distinguishing such vector data from general glTF content. The extension further refines tile and content bounding volumes as needed to support seamless tiled vector rendering in a variety of visual styles.
+
+## Extending 3D Tiles content
+
+A `content` definition, containing a reference URL or template URL to glTF content, may be extended with the `3DTILES_content_gltf_vector`. The extension's boolean property, `vector: true`, indicates that the glTF content of the tile SHOULD be interpreted as vector data. The value of the `vector` property MUST be `true`; for non-vector content the extension is omitted.
+
+```jsonc
+{
+  "root": {
+    "extensionsUsed": [ "3DTILES_content_gltf_vector" ],
```
lilleyse: `"extensionsUsed"` should be a top-level property, not below `"root"`

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,152 @@
...
+- [Feature IDs and Properties](#feature-ids-and-properties)
+- [Visualization](#visualization)
+- [Schema](#schema)
+- [Implementation Examples](#implementation-examples)
+
+## Overview
+
+Applied to geospatial domains, the term “vector data” refers to geometric topologies: points in 2D or 3D coordinate systems; polylines connecting a series of points; or polygons having a closed, exterior loop of points, optionally with additional interior loops defining holes in the polygon. The term "vector data" exists in contrast to “raster data”, which stores pixel grids in image-like formats, rather than discrete geometric types.
+
+Extending the concept of vector data into the domain of graphics APIs, and of 3D scenes already composed of graphics primitives — such as points, lines, triangles — this extension proposes and defines an additional distinction: \*\*“vector data” comprises point, polyline, and polygon geometries with intrinsic semantic meaning, but without intrinsic rendering intent.
```
lilleyse: The formatting might be slightly off here

Image

lilleyse (commented):
Just a few minor comments. Everything looks good at a high level.

extensions/3DTILES_content_gltf_vector/README.md
danielzhong: I understand these cases, but wouldn't it be simpler to remove the `vector：true` flag and only detect whether the extension is present? For example:
```
// Vector content — extension present
"contents": [
  {
    "uri": "splats.glb"
    // No extension → not vector
  },
  {
    "uri": "roads.glb",
    "extensions": {
      "3DTILES_content_gltf_vector": {}   // Extension present → vector
    }
  }
]
```

weegeekps: That's a great point. Does a valid case exist where `vector` may be set to `false`, @donmccurdy?

donmccurdy: I don't know if there's a similar precedent in 3D Tiles, but in glTF the convention is that an extension attached to a core property should:

1. Have no effect on the core, parent property unless non-default extension properties are assigned
2. Use behaviors and property naming conventions so that the extension could (hypothetically) be promoted into core without changes

I can't find a link to the discussion off-hand, but I believe Ed Mackey originally articulated this goal. The extension KHR_materials_unlit was designed before that, and is the main exception I'm aware of.

Do we want 3D Tiles to follow a similar convention? In that case we'd want to design as if the extension could be collapsed onto the content object in the future:

```jsonc
  {
    "uri": "roads.glb",
    "vector": true
  }
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,153 @@
...
+}
+```
+
+Interpretation glTF content as particular vector types is explained further in the following sections. See [points](#points), [polylines](#polylines), and [polygons](#polygons).
+
+Requirements for bounding volumes on glTF vector content may differ from other 3D Tiles content types. See [bounding volumes](#bounding-volumes).
+
+## Points
+
+glTF mesh primitives having `primitive.mode = 0` ("POINTS") SHOULD be interpreted as vector point topologies. Authoring tools SHOULD encode multiple point features within the same glTF mesh primitive, to improve file size and rendering efficiency.
```
bjornblissing, line 87: What constitutes a “feature” for points?
- Is each vertex a feature?
- Can one feature be represented by multiple points?
- If a feature can be multi-vertex, how is feature membership encoded?

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,153 @@
...
+      }
+    }
+    ...
+  }
+}
+```
+
+Interpretation glTF content as particular vector types is explained further in the following sections. See [points](#points), [polylines](#polylines), and [polygons](#polygons).
+
+Requirements for bounding volumes on glTF vector content may differ from other 3D Tiles content types. See [bounding volumes](#bounding-volumes).
```
bjornblissing: Fix for broken anchor:
```suggestion
Requirements for bounding volumes on glTF vector content may differ from other 3D Tiles content types. See [Bounding volumes and clipping](#bounding-volumes-and-clipping).
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,153 @@
...
+      "extensions": {
+	    "3DTILES_content_gltf_vector": { "vector": true }
+      }
+    }
+    ...
+  }
+}
+```
+
+Interpretation glTF content as particular vector types is explained further in the following sections. See [points](#points), [polylines](#polylines), and [polygons](#polygons).
```
bjornblissing: Grammar:
```suggestion
Interpretation of glTF content as particular vector types is explained further in the following sections. See [points](#points), [polylines](#polylines), and [polygons](#polygons).
```

extensions/3DTILES_content_gltf_vector/README.md
```diff
@@ -0,0 +1,153 @@
...
+> A small buffer region, representing a single-digit percentage of a tile's width, is expected to be sufficient for most rendering styles. Client implementations may choose to limit line widths to the widths of tile buffers, in screen space or world space, to avoid visual artifacts.
+
+> [!NOTE]
+> Point geometries do not typically require buffer regions or clipping at tile boundaries for proper rendering, but authoring and client implementations are not prohibited from doing either.
+
+## Feature IDs and Properties
+
+_This section is non-normative._
+
+Point, polyline, and polygon geometries are often associated with other geometries comprising a single conceptual feature. Polyline and polygon geometries may also be cut across neighboring tiles during the tiling process. In such cases, `EXT_mesh_features` and `EXT_structural_metadata` may be used to reference and store properties associated with vector features. Global unique IDs, stored as columns in `EXT_structural_metadata`, may be used to allow features split across multiple tiles to participate in interaction (e.g. highlighting) as a single enttiy.
```
bjornblissing: Fix spelling:
```suggestion
Point, polyline, and polygon geometries are often associated with other geometries comprising a single conceptual feature. Polyline and polygon geometries may also be cut across neighboring tiles during the tiling process. In such cases, `EXT_mesh_features` and `EXT_structural_metadata` may be used to reference and store properties associated with vector features. Global unique IDs, stored as columns in `EXT_structural_metadata`, may be used to allow features split across multiple tiles to participate in interaction (e.g. highlighting) as a single entity.
```

extensions/3DTILES_content_gltf_vector/README.md
donmccurdy: All good questions, thanks! After re-reading this, I believe perhaps the word "feature" should be avoided here and elsewhere in this extension, except under the "Feature IDs and Properties" non-normative section below. Instead we should use the terms like "point geometries" or "point topological primitives", as distinct from glTF mesh primitives.

To your questions, though, the intention is that feature membership is defined separately by [KHR_mesh_features](https://github.com/KhronosGroup/glTF/pull/2082). Vertex/feature mappings may be 1:1, 1:N, N:1, or N:N. But certainly the most common would be 1:1, or N:1 (e.g. many vertices share the same `_FEATURE_ID_n` attribute).

kaioum01711-art:
- [ ] - - - - - - - Google ****

PR #841, opened 2026-05-08
Fix broken link

bjornblissing:
Fix broken link to Google s2-geometry-library-java repo.


javagl:
Thanks (for the fix, and for making me aware of that library to begin with - I'm sure that will come in handy when I want to add S2 extension support to my libraries)

PR #877, opened 2026-08-26
Better management of extension URLs

weegeekps:
Managing the extension URLs and keeping them all up to date is already becoming difficult. This PR changes things a little bit, so we have a single source of truth for all of the URLs to the extensions.

This is typically not hard to do in asciidoc. You can put all of the URLs in attributes within a single file and then `include` the file with those attributes in every doc. Then you update a single URL, and everything is updated. I confirmed this works well in Metanorma. I've named this file `extension-urls.adoc`.

GitHub's asciidoc rendering makes this a bit more difficult, as GitHub purposely doesn't allow `include` directives. I've worked around this by creating a small script named `update-urls.sh`. This script takes all of the content from `extension-urls.adoc` below a `"// COPY STARTS HERE"` line and copies it in between two other markers (`"// GITHUB_URL_START"` and `"// GITHUB_URL_END"`). These copied chunks are ignored during the Metanorma builds and instead those builds actually pull from the `include` directive themselves to allow proper caching and referencing.

Right now, it's an explicit list of the files that we want to have this chunk copied into. I thought about changing it to a glob but that seems unwieldy since we don't need the chunk in every file.


weegeekps:
I haven't swapped over all of the extension URLs yet. I will do that tomorrow AM. I did test it out and it works alright. Once I have swapped relevant files over to using this method, I will open this draft fully.

@javagl: I wouldn't mind a gut check on this idea, given your experience with Metanorma and asciidoc in GitHub.

javagl:
I see the intention behind this. I don't really "understand" the shell script (as there is no `.bat` file). And I haven't tried it out with the finicky metanorma toolchains (and I think that there are some open questions about that anyhow - I remember some hassle that I tried to summarize in the ["Cross Linking Between Files" section](https://github.com/CesiumGS/3d-tiles/blob/141528ec966794090d22eb9d38494e71e76fcc7d/specification/2.0/BUILDING.md#cross-linking-between-files)). I have not tried out all links.

---

Asides:

I randomly stumbled over the

> Requirement 1: Support [3DTILES_implicit_tiling](https://github.com/CesiumGS/3d-tiles/blob/141528ec966794090d22eb9d38494e71e76fcc7d/specification/2.0/%7Bgltf-3dtiles_implicit_tiling-url%7D)

link in the CONFORMANCE.adoc, and that link is broken (probably only a typo - a _systematic_ check of _all_ links should be done at some point). 

Also, https://github.com/CesiumGS/glTF/blob/3d-tiles-2.0/extensions/2.1/Vendor/EXT_geospatial_crs/README.md currently has `EXT_crs` as its title... 

More broadly: There's probably a lot to straighten out, but none of that relevant for this PR right now...

---

... so back to the PR:

Having the `extension-urls.adoc` and the shell script, when the shell script **only** inserts this exact ADOC file into another one raises the question of why there is that dedicated ADOC file to begin with. The consideration that ~"similar things may be done for other files", and "there is a file that _only_ contains the link definitions (and is the single source of truth)" may justify that. 

---

Much higher level: 

I don't know if there has been any decisions that give an agreed-upon(!) meaning to the term "Conformance Class". It looks like "We need profiles, but we don't have them, so let's define ✨ Conformance Classes ✨ ". In any case, I think that there should be one MD/ADOC file that simply lists these extensions anyhow. (The fact that the "Conformance" one currently is the only one that lists all of them is a mere coincidence...). Such a document - maybe with a fancy name like `EXTENSION_REGISTRY.adoc` or whatnot - could contain something like this:

> # All extensions
> 
> ## 3D Tiles extensions
> 
> ### [3DTILES_shape_s2](https://github.com/CesiumGS/glTF/tree/3d-tiles-2.0/extensions/2.1/Vendor/3DTILES_shape_s2/README.md)
>
> This extension defines S2 cells as an additional shape type for glTF 2.1 shapes.
>
> ### OTHER_one
>
> ...
> 
> ## Geospatial extensions
> 
> ### [EXT_geospatial_crs](https://github.com/CesiumGS/glTF/tree/3d-tiles-2.0/extensions/2.1/Vendor/EXT_geospatial_crs/README.md) 
> 
> Assets with this extension are declared to have been authored for geospecific usage, with a particular coordinate reference system (CRS) that overrides the coordinate systemof the base glTF specification.
>  
> ### ANOTHER_one
> ...

The intentions:

- some grouping by category (i.e. roughly the prefixes)
- **anchor-links** for each extension, so that other places can refer to `EXTENSION_REGISTRY#ExtensionName` **without** having to deal with the question of which directory/branch the spec is currently in
- Each section contains _a single paragraph_ that summarizes what the extension is (!)

Preferably, this single paragraph should be _exactly_ the first paragraph/sentence from the `Overview` section of the respective README. In this case, this whole file could be auto-generated - **iff** we agree on starting the respective `Overview` with a sentence that is suitable as the first paragraph of an 'Overview' (which currently is not always the case)

---

(In terms of whether this PR should be merged or not, I can only say: Whatever works for you (or someone else who uses Linux)... 🤷‍♂️ )

weegeekps:
> (as there is no .bat file)

I went with `awk` because outside of `sed` I don't know of another tool that can do this job without installing extra stuff. This is unfortunately only true on Linux. Once we ship, I will come up with a better script that is cross platform.

> Having the extension-urls.adoc and the shell script, when the shell script only inserts this exact ADOC file into another one raises the question of why there is that dedicated ADOC file to begin with.

I haven't swapped over all of the extension URLs to use this new method yet. My plan is all files that are using extension URLs should have this.

> I don't know if there has been any decisions that give an agreed-upon(!) meaning to the term "Conformance Class".

Conformance classes and how these are structured are an OGC thing. There's not much we can or should change to these at this point. @akmorgan99 has been very influential in how we shape these.

> Extension Registry

We sort of have this already: [glTF.adoc](https://github.com/CesiumGS/3d-tiles/blob/draft-2.0/specification/2.0/glTF.adoc). Yesterday we decided that some of the extension links will just live in `3DTILES_tileset`. That extension will be embedded into the document as Appendix A. I have a setup locally that can perform the markdown to asciidoc conversion for Metanorma but it's unlikely to be committed anywhere before the end of the week.

EDIT: I really like your thoughts about autogeneration of that registry. That would simplify some management of this longer term.

lilleyse:
> Also, https://github.com/CesiumGS/glTF/blob/3d-tiles-2.0/extensions/2.1/Vendor/EXT_geospatial_crs/README.md currently has `EXT_crs` as its title... 

Fixed

weegeekps:
Every extension should now be using this. I've hand tested all the links and they work. I am going to merge this to the draft.

